PR #33, opened 2020-11-04
Collecting commits without Docker

ninjinkun:
- Add `--executable` option to the `commit` command
  - It has `jar` and `docker` options. `jar` is the default.
- Download `exec_deploy.jar` if it doesn't exist

## What point do I want you to review?
- I added versioning to `exec_deploy.jar` on S3 and local. Do you agree it works well?
- I added `~/.launchable` as a jar cache directory. Do you think is this directory right place to store jar files?

### DIFF
--- a/launchable/commands/record/commit.py
+++ b/launchable/commands/record/commit.py
@@ -2,6 +2,7 @@ from ...utils.token import parse_token
 import os
 import click
 from ...utils.ingester_image import ingester_image
+import subprocess
 
 
 @click.command()
@@ -9,12 +10,40 @@ from ...utils.ingester_image import ingester_image
               help="repository path",
               default=os.getcwd(),
               type=str
-              )
-def commit(source):
-    token, org, workspace = parse_token()
+)
+@click.option('--executable',
+              help="",
+              type=click.Choice(['jar', 'docker']),
+              default='jar'
+)
+def commit(source, executable):
     source = os.path.abspath(source)
+
+    if executable == 'jar':
+        exec_jar(source)
+    elif executable == 'docker':
+        exec_docker(source)
+
+
+def exec_jar(source):
+    res = subprocess.run(["which", "java"], stdout=subprocess.DEVNULL)
+    java = None
+    if res.returncode == 0:
+        java = "java"
+    elif os.access(os.path.expandvars("$JAVA_HOME/bin/java"), os.X_OK):
+        java = "$JAVA_HOME/bin/java"
+
+    if not java:
+        exit("You need to install Java or try --executable docker")
+
     os.system(
-        "docker run -u $(id -u) -i --rm \
-        -v {}:{} --env LAUNCHABLE_TOKEN {} ingest:commit {}"
+        "{} -jar bin/exe_deploy.jar ingest:commit {}"
+        .format(java, source))
+
+
+def exec_docker(source):
+    os.system(
+        "docker run -u $(id -u) -i --rm " \
+        "-v {}:{} --env LAUNCHABLE_TOKEN {} ingest:commit {}"
         .format(source, source, ingester_image, source)
     )
--- a/launchable/commands/record/commit.py
+++ b/launchable/commands/record/commit.py
@@ -58,7 +58,8 @@ def download_jar():
                 sys.stdout.flush()
 
         f.close()
-        print("\nexe_deploy.jar is downloaded in {}", jar_dir_path)
+        print("")
+        print("exe_deploy.jar is downloaded in {}".format(jar_dir_path))
 
 
 def exec_jar(source):
--- a/launchable/commands/record/commit.py
+++ b/launchable/commands/record/commit.py
@@ -58,7 +58,6 @@ def download_jar():
                 sys.stdout.flush()
 
         f.close()
-        print("")
         print("exe_deploy.jar is downloaded in {}".format(jar_dir_path))
 
 
--- a/launchable/commands/record/commit.py
+++ b/launchable/commands/record/commit.py
@@ -41,7 +41,7 @@ def download_jar():
     total_length = int(res.headers.get('content-length'))
 
     if os.path.exists(jar_file_path) and os.path.getsize(jar_file_path) == total_length:
-        print("exe_deploy.jar exits at {}".format(jar_dir_path))
+        print("exe_deploy.jar exists at {}".format(jar_dir_path))
     else:
         total_downloaded = 0
         f = open(jar_file_path, 'wb')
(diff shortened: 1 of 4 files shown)


launchable/utils/ingester_jar.py
```diff
@@ -0,0 +1,3 @@
+ingester_jar_version = "0.1.0"
```
ninjinkun, line 1: If `exe_deploy.jar` is updated, you should bump up this version number. I assume the number will be the same as the CLI version number.

launchable/commands/record/commit.py
```diff
@@ -1,7 +1,16 @@
...
 from ...utils.ingester_image import ingester_image
+from ...utils.ingester_jar import ingester_jar_url, ingester_jar_version
+import subprocess
+import requests
+import pathlib
+import sys
+
+
+jar_dir_path = os.path.expanduser(
+    "~/.launchable/jar/ingester/{}".format(ingester_jar_version))
```
ninjinkun, line 12: I specified `~/.launchable/jar/ingester/0.1.0/exe_deploy.jar` as a local jar file path. Is this okay?

ninjinkun:
@kohsuke Please review this PR. Its run result is here https://github.com/launchableinc/rocket-car/runs/1353609834

launchable/commands/record/commit.py
```diff
@@ -10,11 +19,67 @@
...
+        download_jar()
+        exec_jar(source)
+    elif executable == 'docker':
+        exec_docker(source)
+
+
+def download_jar():
+    pathlib.Path(jar_dir_path).mkdir(parents=True, exist_ok=True)
+
+    res = requests.get(ingester_jar_url, stream=True)
```
kohsuke, line 40: I could be wrong, but [according to this](https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions) it seems to me like `res.raise_for_status()` would be needed to detect 404 and other errors.

launchable/commands/record/commit.py
```diff
@@ -10,11 +19,67 @@
...
+    total_length = int(res.headers.get('content-length'))
```
kohsuke, line 41: I'm also mildly concerned that every invocation of build recording now creates a request to S3 and any problems there will render the entire command inoperable. We already know for example that BMW's build environment needs IP-based whitelisting, so I think this would create an extra step for them.

If we download a file into a temporary file, then rename it to the expected name, then `os.path.exists(...)` would be sufficient to verify that completeness of a file, so that would work avoid this problem. In this way, BMW can also just manually download a file and place it in the location they expect. Also handy for people who create a builder image.

launchable/commands/record/commit.py
```diff
@@ -10,11 +19,67 @@
...
+        print("exe_deploy.jar exists in {}".format(jar_dir_path))
+    else:
+        total_downloaded = 0
+        f = open(jar_file_path, 'wb')
+        for chunk in res.iter_content(chunk_size=512 * 1024):
+            if chunk:
+                f.write(chunk)
+
+                total_downloaded += len(chunk)
+                sys.stdout.write("\rDownloading exe_deploy.jar: {:.1f}% ({:,} / {:,} byte)".format(
```
kohsuke, line 53: For interactive use this status report is nice but for CI systems where this command is expected to be used, this is effectively no different from NL, so it'd be quite noisy.

Not a major point, but I'd suggest toning this down. For example, one line to report the start of download (some CI systems can only show a line after it sees a whole line), then one dot for every 10MB, and NL at the end of download, would be nice. Alternatively, you can check if the stdout is a terminal, and use that as a cue to produce output geared for humans like this.

(But really, the whole on demand download is an extra nice-to-have, so it's probably not wise to get too smart in that regard.)

launchable/commands/record/commit.py
```diff
@@ -10,11 +19,67 @@
...
+    java = None
+    if res.returncode == 0:
+        java = "java"
+    elif os.access(os.path.expandvars("$JAVA_HOME/bin/java"), os.X_OK):
+        java = "$JAVA_HOME/bin/java"
+
+    if not java:
+        exit("You need to install Java or try --executable docker")
+
+    os.system(
```
kohsuke, line 75: When the commit ingester fails, how should we handle it? It seems to me that the code today will silently ignore that and moves on. Perhaps that's intended (so as not to disrupt the customer's build/test), or maybe it's an oversight.

More broadly, this does seem like something we should think carefully. There are competing factors we need to balance -- we don't want to disrupt customers' build & test and get kicked out of their build process, but we also want them to notice if things aren't working as it's supposed to be.

Whatever our policy is that we define there, we should apply that consistently in other commands, too.

launchable/commands/record/commit.py
```diff
@@ -10,11 +19,67 @@
...
+        sys.stdout.write("\nexe_deploy.jar is downloaded in {}\n".format(jar_dir_path))
+
+
+def exec_jar(source):
+    res = subprocess.run(["which", "java"], stdout=subprocess.DEVNULL)
+    java = None
+    if res.returncode == 0:
+        java = "java"
+    elif os.access(os.path.expandvars("$JAVA_HOME/bin/java"), os.X_OK):
+        java = "$JAVA_HOME/bin/java"
```
kohsuke, line 70: Nice touch!

kohsuke (approved):
I'm marking this as approved so as not to block the delivery to the rocket-car company, which is on the critical path for the company.

I left some comments around the design of the capabilities.

launchable/commands/record/commit.py
ninjinkun: Got it. I'm going to implement to handle HTTP errors.

launchable/commands/record/commit.py
ninjinkun: Thank you for your suggestion. I implemented moving `jar` from the temporary directory to the cache directory. As a result, the CLI doesn't need to access S3 if the jar file was downloaded.

launchable/commands/record/commit.py
ninjinkun: Okay. I fixed to tone down this prompt like this
```
Downloading exe_deploy.jar (262.03 MB)
....
```

launchable/commands/record/commit.py
ninjinkun: It's simply an oversight. I think `LAUNCHABLE_REPORT_ERROR` environment variable is a good option for this fail.  I'll fix it at another PR after merge this PR.

ninjinkun:
@kohsuke Thank you for your review! I merge it before your second review because I want to roll out it to the rocket car company. If you have another review, please leave comments on this PR. I'll fix them on another PR.

PR #36, opened 2020-11-10
Replace the jar file on S3 with the jar file in repository internal 

ninjinkun:
@kohsuke reduced `exe_deploy` to 20MB. Let's move it within the repository internal.


kohsuke (approved):
Great!!

PR #38, opened 2020-11-17
Update the latest exe_deploy.jar

ninjinkun:


kohsuke (approved):
I'm doing this without fully understanding the implication. I guess `0.1.3` jar was manually staged?

ninjinkun:
Yes. I built the jar on my local, and staged it manually.

PR #43, opened 2020-11-26
Add LaunchableClient and LAUNCHABLE_BASE_URL env

ninjinkun:
To became API base URL replaceable, I introduced LaunchableClient.
- Add LaunchableClient
- Add LAUNCHABLE_BASE_URL environment variable

### DIFF
--- a/launchable/utils/http_client.py
+++ b/launchable/utils/http_client.py
@@ -1,0 +1,25 @@
+import requests
+import os
+import platform
+from launchable.version import __version__
+
+
+BASE_URL_KEY = "LAUNCHABLE_BASE_URL"
+DEFAULT_BASE_URL = "https://api.mercury.launchableinc.com"
+
+
+class LaunchableClient:
+    def __init__(self, token, base_url="", http=requests):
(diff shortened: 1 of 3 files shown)


launchable/utils/http_client.py
```diff
@@ -0,0 +1,25 @@
+import requests
+import os
+import platform
+from launchable.version import __version__
+
+
+BASE_URL_KEY = "LAUNCHABLE_BASE_URL"
+DEFAULT_BASE_URL = "https://api-static.mercury.launchableinc.com"
```
shuheiktgw, line 8: @ninjinkun The default base URL should be `https://api.mercury.launchableinc.com`. `api-static` subdomain is used only when you want static IPs.

shuheiktgw:
@ninjinkun I left one comment but except that it looks good to me!

launchable/utils/http_client.py
ninjinkun: Oh, I see

ninjinkun:
@shuheiktgw Thank you for your review. I fixed the API host.

PR #44, opened 2020-11-30
Record test from JUnit xml for the generic CLI

ninjinkun:
- Add `launchable record test` command
  - Read JUnit XML report
  - Send test cases to the API
- Test structure parameters such as file/class/case will implement when the corresponding API is implemented


shuheiktgw (approved):
LGTM 👍

PR #45, opened 2020-12-01
Pass LAUNCHABLE_BASE_URL to Ingester jar file

ninjinkun:
Ingester has `-endpoint` flag which overrides its API host. To override the jar internal API host variable from the CLI, pass `LAUNCHABLE_BASE_URL` as a `-endpoint` flag.


shuheiktgw (approved):
👍

PR #47, opened 2020-12-04
Implement a verify command

shuheiktgw:
I implemented a verify command to test if the Launchable CLI is configured properly or not.

Success

```
Platform: macOS-10.15.7-x86_64-i386-64bit
Python version: 3.8.3
Java command: java
Your CLI configuration is successfully verified 🎉
```

Failure

```
Authentication failed. Most likely the value for LAUNCHABLE_TOKEN environment variable is invalid.

or 

Authentication failed. Most likely the value for the LAUNCHABLE_TOKEN environment variable is invalid.

or 

Java is not installed. You need Java to use the Launchable CLI.
```

### DIFF
--- a/launchable/commands/verify.py
+++ b/launchable/commands/verify.py
@@ -8,13 +8,12 @@ from ..utils.http_client import LaunchableClient
 from ..utils.token import parse_token
 from ..utils.java import get_java_command
 
-
 @click.command(name="verify")
 def verify():
     try:
         if os.getenv("LAUNCHABLE_TOKEN") is None:
-            click.echo(click.style(f"Could not find the LAUNCHABLE_TOKEN environment variable. Please check if you "
-                                   f"set it properly.", fg="red"))
+            click.echo(click.style("Could not find the LAUNCHABLE_TOKEN environment variable. Please check if you "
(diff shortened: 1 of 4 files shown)


launchable/commands/verify.py
```diff
@@ -0,0 +1,55 @@
...
+
+        res.raise_for_status()
+
+        click.echo("Platform: " + platform.platform())
+        click.echo("Python version: " + platform.python_version())
+
+        java = get_java_command()
+
+        if java is None:
+            click.echo(click.style(f"Java is not installed. You need Java to use the Launchable CLI", fg="red"))
```
shuheiktgw, line 44: Do we need to check Java installation strictly or is it ok a user does not install Java since we have a docker option?

ninjinkun: I think we need to check Java because the docker option is no longer maintained. The docker option should be deprecated.

ninjinkun (approved):

ninjinkun:
I recommend `autopep8 --in-place -r .` to fix the lint error.

shuheiktgw:
Yes, I finally nailed it. Thank you for your review!

PR #50, opened 2020-12-07
Recording test paths

ninjinkun:
- Recording `file`, `classname`, `timestamp`, `lineno` when the CLI sends test cases


shuheiktgw (approved):
LGTM 👍

PR #51, opened 2020-12-07
Change to exit code 1 when the verify fails

yoshiori:
Change to exit code 1 when the verify fails.
Makes it easier for the computer to determine success or failure.
(e.g. unix pipe processing)

for example:

```
$ laucnable verify && launchable record build --name BUILD_ID --source .
```

### DIFF
--- a/launchable/commands/verify.py
+++ b/launchable/commands/verify.py
@@ -12,9 +12,8 @@ from ..utils.java import get_java_command
 def verify():
     try:
         if os.getenv("LAUNCHABLE_TOKEN") is None:
-            click.echo(click.style("Could not find the LAUNCHABLE_TOKEN environment variable. Please check if you "
+            raise click.UsageError(click.style("Could not find the LAUNCHABLE_TOKEN environment variable. Please check if you "
                                    "set it properly.", fg="red"))
-            return
 
         token, org, workspace = parse_token()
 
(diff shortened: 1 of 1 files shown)


shuheiktgw:
Thank you for your PR! 👍 

Actually, the design that our CLI exit with a non-zero exit code is rather intentional, which follows one of our design manifests, `Customers' builds cannot stop`. However, I agree that sometimes users want the CLI just fails when it fails. In that case, we (I and @ninjinkun) prepared the `LAUNCHABLE_REPORT_ERROR` environment variable. I think we should use the convention here but what do you think?

yoshiori:
[IMO]
It's a different story if the end state is easy for the computer to understand and whether to stop the build. I think the primitive conventions of unix command should be followed.
Of course, we should not stop the build, but that should be done in a different way.
For example, connect pipes with `|` instead of `&&`.

```
$ laucnable verify | launchable record build --name BUILD_ID --source .
```

kohsuke:
Seems to me that the disconnect is around how this command is supposed to be used. I'm not finding any discussion that led to this command in Confluence, but is this meant to be used interactively and/or only during the onboarding, or is this meant to be a part of the build script perpetually?

If the former, it makes sense for this to return non-zero exit code given that the intent is to catch errors. If the latter, it makes sense to give consideration to "customers build cannot stop"

shuheiktgw:
Hmm, I think you are right. Considering the current use case of the verify command, our customers mostly use this command when onboarding so exiting with non-zero code is the expected behavior. Thank you for your fix!

shuheiktgw (approved):

PR #52, opened 2020-12-09
Add record session command

ninjinkun:
`optimize test` command need `session_id` before recording test results. Therefore split session creation request from `record test` command.

`record session` command is used like this;
```shell
$ LAUNCHABLE_TEST_SESSION=$(launchable record session --name 123)             
$ rails test $(launchable optimize test $(find test -regex .*_test.rb | tr "\n" " ") --name 123 --session $LAUNCHABLE_TEST_SESSION)
$ launchable record test test.xml --name 123 --session $LAUNCHABLE_TEST_SESSION
```

### DIFF
--- a/launchable/commands/record/__init__.py
+++ b/launchable/commands/record/__init__.py
@@ -1,6 +1,7 @@
 from .build import build
 from .commit import commit
 from .test import test
+from .session import session
 import click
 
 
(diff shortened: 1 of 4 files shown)


shuheiktgw (approved):
👍

PR #54, opened 2020-12-09
We require Python3

kohsuke:
As per https://www.python.org/dev/peps/pep-0394/, as we require Python3, our instruction should be `pip3` to be more portable regardless of the Python version that `pip` resolves to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Installation
 
 ```shell
-pip install launchable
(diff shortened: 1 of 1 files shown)


shuheiktgw (approved):
👍 

I hope this PR catches @ninjinkun's eye too

ninjinkun (approved):

PR #56, opened 2020-12-16
Show self version in verify

yoshiori:
It also shows the version of itself.

cli — -fish — 302×88 2020-12-16 22-00-14


kohsuke (approved):
Nice!

PR #57, opened 2020-12-17
Ignore API errors in optimize command

yoshiori:
When Error occurs, return the test name as it is passed. because We shud not stop build process.

SEE https://launchableinc.atlassian.net/browse/ST-634


ninjinkun (approved):
LGTM

PR #58, opened 2020-12-17
Grab module version from Git tag

ninjinkun:
We maintain two version representatives such as `/version.py` and [Git tags](https://github.com/launchableinc/cli/tags). This PR integrates them into Git tag via [setuptools_scm](https://pypi.org/project/setuptools-scm/).


yoshiori:
please change this file. because using version.py

launchable/commands/verify.py

yoshiori (approved):
Others look good

PR #60, opened 2020-12-18
Ignore session creation failure

ninjinkun:
Current `record test` and `optimize test` require `session_id`. However, if Launchable API caused an error in the session creation phase, those commands would be failed. To keep working customer's CI, I modified the CLI to ignore missing session_id.

### DIFF
--- a/launchable/commands/optimize/test.py
+++ b/launchable/commands/optimize/test.py
@@ -20,8 +20,8 @@ from ...utils.token import parse_token
     '--session',
     'session_id',
     help='Test session ID',
-    required=True,
     type=int,
+    required=os.getenv(REPORT_ERROR_KEY),
 )
 @click.option(
     '--source',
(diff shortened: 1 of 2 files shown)


kohsuke:
I agree with the motivation of the change, but I'm not too sure about how the proposed change actually behaves.

Citing [our error handling design principle](https://launchableinc.atlassian.net/wiki/spaces/PRODUCT/pages/761331921/CLI+Error+Handling+Design+Principles),

* **Can they solve the problem?** When I see error messages that say "it may be X", I ask myself, what are the other possibilities? Can we tell them apart for users? In this case, I assume that other possibilities are misconfiguration, as the `--session` option is now made optional. If so, I wonder if a better thing to do is for a session creation to leave a known "error" value in case of a failure. That way, we can be certain that the installation is proper, but for a reason we just failed to create a session.
* From the same question, I think the error message needs to be improved. Put yourself in the shoes of the person who sees this messag. What are you gonna do? Put yourself in the shoes of the CSM person who is going to help this user. What do you tell them? What can we do to help their customer support work easier? If we do the suggested change above, I think there'll be two things we want to tell them in subsequent cascading failures. (1) what this failure means, in this case that we are running full tests in the default order because of an earlier error, and (2) re-print what that earlier error was (this is information we can carry in the error value), and what their action should be, which can range from "our apologiese for service disruption, contact the support if this doesn't resolve itself in one hour" to "create a new API key".
* **Issue a visible warning message but keep on** I'd think `optimize test` command should also issue a warning in case of a botched session creation.

kohsuke (approved):
Leave it up to the discretion if this is OK enough or not

kohsuke:
Branches are going stale quickly given the rapid pace of the change. I'm merging this in for now as is.

PR #64, opened 2020-12-21
Lint with autopep8

ninjinkun:


kohsuke:
Merging this in to avoid merge problems down the road. This looks all just syntactic change.

PR #65, opened 2020-12-21
Send test paths in the optimize tests command

ninjinkun:
Send test paths in the `launchable optimize test`. 

For example,
```json
[{ 
  "type": "file", 
  "name": "test/models/admin/user_test.rb"
}]


kohsuke (approved):
Thanks. A step in the right direction but it does open up more questions now that I see this!

PR #66, opened 2020-12-21
Pushed test name handling forward.

kohsuke:
In both `optimize tests` and `record tests`, tests are identified by their test paths.

Introduced an abstraction in the client to build test paths. For `optimize tests`, Optimize still access a string that looks like a file
name and builds a default test path. But this is currently not aligned with what `record tests` do, so something needs to change.

Bazel implementation & Gradle is now complete, and it demonstrates why this change was needed in the first place --- test paths reported by `record tests` should be a sub-path of those used by `optimize tests`

# Other improvements
* Robustness: `optimize tests generic` to read tests from stdin

# Next Steps

- [x] demonstrates Mini Test binding
- [ ] Writing some unit tests. How?
- [x] Update the server side

### DIFF
--- a/launchable/commands/optimize/tests.py
+++ b/launchable/commands/optimize/tests.py
@@ -46,7 +46,7 @@ def tests(context, target, session_id, source, build_name):
     # TODO: placed here to minimize invasion in this PR to reduce the likelihood of
     # PR merge hell. This should be moved to a top-level class
     class Optimize:
-        test_paths: List[TestPath]
+        # test_paths: List[TestPath]  # doesn't work with Python 3.5
 
         # Where we take TestPath, we also accept a path name as a string.
         TestPathLike = Union[str,TestPath]
--- a/launchable/commands/optimize/tests.py
+++ b/launchable/commands/optimize/tests.py
@@ -80,7 +80,7 @@ def tests(context, target, session_id, source, build_name):
             else:
                 return x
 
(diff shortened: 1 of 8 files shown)


launchable/commands/optimize/tests.py
```diff
@@ -45,48 +46,65 @@ def tests(context, target, session_id, source, build_name):
...
         def __init__(self):
             self.test_paths = []
-            self._formatter = lambda x: x
+            # default formatter t hat's in line with to_test_path(str)
+            # TODO: robustness improvement.
+            self._formatter = lambda x: x[0]['name']
 
         @property
-        def formatter(self) -> Callable[[str], str]:
+        def formatter(self) -> Callable[[TestPath], str]:
```
ninjinkun, line 61: After merged this PR, we have to run mypy on the GithubActions.

launchable/test_runners/minitest.py
```diff
@@ -0,0 +1,39 @@
...
+@click.argument('files', required=True, nargs=-1)
+@launchable.optimize.tests
+def optimize_tests(client, files):
+    def parse(fname:str):
+        '''
+        Scan a file, directory full of *.rb, or @FILE
+        '''
+        if os.path.isdir(fname):
+            client.scan(fname, '**/*.rb')
+        elif fname=='@-':
```
ninjinkun, line 15: We need to note the behavior to `help` prop in the `@click.argument()` annotation.

ninjinkun (approved):
Go ahead!

PR #67, opened 2020-12-22
Revive base path

ninjinkun:
Current implementation drops the base file path. For example, if you specify `launchable optimize minitest tests`, the command outputs `models/open_class_user_test.rb` without `test` directory.  However, minitest needs full path such as `test/models/open_class_user_test.rb` to test run.  That is why I deleted the line removing a base path component. Instead of the line, I added file path normalization with`sourc option.

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -97,6 +97,7 @@ def subset(context, target, session_id, source, build_name):
                 - if a TestPath is returned, that's added as is
             """
             for t in glob.iglob(os.path.join(base, pattern), recursive=True):
+                t = os.path.normpath(os.path.relpath(t, start=source)) if source else t
                 t = path_builder(t)
                 if t:
                     self.test_paths.append(self.to_test_path(t))
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -97,7 +97,6 @@ def subset(context, target, session_id, source, build_name):
                 - if a TestPath is returned, that's added as is
             """
             for t in glob.iglob(os.path.join(base, pattern), recursive=True):
-                t = t[len(base) + 1:]  # drop the base portion
                 t = path_builder(t)
                 if t:
                     self.test_paths.append(self.to_test_path(t))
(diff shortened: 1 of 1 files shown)


kohsuke:
Thanks for discovering this subtle and important problem, this is indeed broken. We need to be careful what these strings represent, indeed. But this change now breaks Gradle side, so we need to come up with something else.

Does Python have a type that represents a path name, ala `java.io.File` or `java.nio.file.Path`? If so, using that would tighten up the abstraction.

Several competing requirements for this `scan` method:

* The `base` parameter is assumed to be resolvable path (or else `glob.iglob` won't work). We will state this.
* Test runners like Mini Test, needing to receive resolvable paths. For those, the existing behaviour of silently dropping the base would be rather surprising.
* Test runners like Gradle & Bazel, which manipulates the obtained paths into something else such as class name and labels. They need to first drop the base portion from the enumerated paths, then do the string manipulation to the rest. Ideally they need enumerated paths relative to `base`, and if that's difficult, at least they need a guarantee that the enumerated path contains `base` as the prefix. Here, `os.path.normpath` would break that.

I think one way to do this is to modify `path_builder` to take two arguments, `base` and an enumerated path resolvable with `base` (IOW, `t[len(base)+1:]`) and expect that method to combine them back into one resolvable path, which shall be the default behavior, or build a custom `TestPath`. The only down side I can think of is that the caller who just wants to do filtering now at least needs to do the path combination, though so far we haven't come across such a caller yet.

Perhaps another way is to split those two into two separate methods. Maybe that is less confusing.


Oh, and if the intent of `relpath` is to capture test paths that are portable, then we need the reverse operation in the formatter side or else the path names the command eventually produce will not be resolvable.

And finally, in my mind this subtle problem and nuances in fixes really underscore the need to write tests to ensure these key invariants are maintained going forward.

kohsuke (changes requested):
We should probably discuss this to decide how best to fix this, but I'm also quite happy to leave it up to you to decide how to move forward on this one for the interest of progress. I captured all the problems I'm aware of in the above lengthy comment.

kohsuke:
We decided to go clam everything into `path_builder`

PR #68, opened 2020-12-22
Specify **/*_test.rb for ignoring test helpers

ninjinkun:


kohsuke (commented):
This is a conditional :+1:. Namely, is this as established naming convention in the community where this is used?

I ask because I did spend like 15mins looking around Mini test documentation to see if I can find the evidence of an established naming convention, but I couldn't. But if it is, then I think it's the right change.

ninjinkun:
According to my 5min experiment on the Rails repo, I found at least **Rails with Minitest** expects "_test.rb" naming. When I renamed one test from "user_test.rb" to "test_user.rb", `rails test` don't run the test.

ninjinkun:
I'm going to go with `**/*_test.rb` in this time because I think most of Minitst users uses Rails.

ninjinkun:
This PR moves to https://github.com/launchableinc/cli/pull/76

PR #69, opened 2020-12-22
Add bazel subset formatter

ninjinkun:
On Bazel request, subsetting API returns a package and a target such as  ` [{"type":"package","name":"//src/test/java/com/ninjinkun"},{"type":"target","name":"mylib_test2\n"}]`. We need to format them to `/src/test/java/com/ninjinkun:mylib_test2`.


kohsuke (approved):
Thanks!

kohsuke:
Merging in what I can

PR #77, opened 2020-12-23
One more round of touch-ups

kohsuke:
... as I was writing the documentation: https://github.com/launchableinc/mothership/pull/1007

### DIFF
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -1,7 +1,6 @@
 import re
 import click
 import subprocess
-import urllib.request
 import json
 import os
 from ...utils.token import parse_token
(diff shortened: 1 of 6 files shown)


launchable/commands/record/build.py
```diff
@@ -13,33 +12,35 @@
...
-    'REPO_DIST like --source . '
-    'or REPO_NAME=REPO_DIST like --source main=./main --source lib=./main/lib',
+    help='path to local Git workspace, optionally prefixed by a label.  '
+    ' like --source path/to/ws or --source main=path/to/ws',
     default=["."],
     metavar="REPO_NAME",
     multiple=True
 )
-@click.option('--with-commit/--without-commit', help='', default=True)
+@click.option('--with-commits', help='We ask for your explicit consent that we do process your commits', is_flag=True)
```
ninjinkun, line 29: In my idea, we should remove `--with-commit` and `--without-commit` options. I wrote the reason on the confluence https://launchableinc.atlassian.net/wiki/spaces/PRODUCT/pages/887226612/CLI+UX+improvement+proposals?focusedCommentId=908066921#comment-908066921

kohsuke: Thanks. I'll see how they respond, but as I understand it, the `--with-commits` is adovacated by John as an "opt-in" mechanism to ensure people understand that we look at their commits. I don't have any strong opinion on this one way or the other, but I do want to cut v1.0 within this side of the year and write to PAs.

Is there anything I can do to get the rest of the changes in?

ninjinkun:
Except `--with-commits` option things look good.

launchable/commands/record/build.py
ninjinkun: I strongly against with force `--with-commit` option. Customers don't have any choice except with `--with-commits`. That is why I insisted we should remove those options. It's to be a default. An alternative idea, I propose we keep `--with-commit` with `default=True`.

However, I don't want to block the CLI v1.0 release. Let's discuss tomorrow what the CLI will be in 1.0. We can consider the option to change it in 1.1 after the release.

kohsuke: For others, when we say keep '--with-commits` with `default=True` what that means is that  the option is tolerated but meaningless. The behavior will be exactly the same regardless of whether or not the option is present.

awilkes: Yeah, OK. I'm on board with removing the option altogether. @jlong indicated he was OK with it in Slack, too.

ninjinkun (approved):

PR #78, opened 2020-12-25
Fix "record tests" is sending test names in an array

ninjinkun:
Python returns a tuple from a tailing comma term such as `result = term_a,`!


kohsuke (approved):
Wow, today I learned.

PR #79, opened 2020-12-28
Add GoogleTest plugin

ninjinkun:
This GoogleTest plugin support GoogleTest + CTest output and GoogleTest standalone + `--gtest_list_tests` output as shown below.

## Google Test + CTest output sample
```shell
ctest -N
Test project /Users/ninjin/src/github.com/launchableinc/rocket-car-googletest
  Test #1: FooTest.Bar
  Test #2: FooTest.Baz
  Test #3: FooTest.Foo
  Test #4: */ParameterizedTest.Bar/*

Total Tests: 4
```

CTest sample is here https://github.com/launchableinc/rocket-car-googletest/pull/4

## GoogleTest standalone output sample
```shell
$ ./test_a --gtest_list_tests 
FooTest.
  Bar
  Baz
  Foo
```

### DIFF
--- a/launchable/test_runners/googletest.py
+++ b/launchable/test_runners/googletest.py
@@ -14,7 +14,8 @@ def make_test_path(cls, case) -> TestPath:
 def subset(client):
     # Read targets from stdin, which generally looks like //foo/bar:zot
     for label in sys.stdin:
-        #  -> //foo/bar & zot
+        # handle ctest -N output
+        # Test #1: FooTest.Bar -> (FooTest, Bar)
         result = re.match(r'^ *Test #\d+: ([^ ]+)$', label)
         if result:
             (cls, case) = result.group(1).split('.')
--- a/launchable/test_runners/googletest.py
+++ b/launchable/test_runners/googletest.py
@@ -1,0 +1,32 @@
+import sys
+import re
+import click
+
+from . import launchable
+from ..testpath import TestPath
+
+
+def make_test_path(cls, case) -> TestPath:
+    return [{'type': 'class', 'name': cls}, {'type': 'case', 'name': case}]
+
+
+@launchable.subset
+def subset(client):
+    # Read targets from stdin, which generally looks like //foo/bar:zot
+    for label in sys.stdin:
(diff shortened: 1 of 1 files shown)


launchable/test_runners/googletest.py
```diff
@@ -0,0 +1,46 @@
...
+from ..testpath import TestPath
+
+
+def make_test_path(cls, case) -> TestPath:
+    return [{'type': 'class', 'name': cls}, {'type': 'case', 'name': case}]
+
+
+@launchable.subset
+def subset(client):
+    # Read targets from stdin, which generally looks like //foo/bar:zot
```
kohsuke, line 15: I assume this is just a stale left-over comment

launchable/test_runners/googletest.py
```diff
@@ -0,0 +1,46 @@
+import sys
+import re
+import click
+
+from . import launchable
+from ..testpath import TestPath
+
+
+def make_test_path(cls, case) -> TestPath:
+    return [{'type': 'class', 'name': cls}, {'type': 'case', 'name': case}]
```
kohsuke, line 10: Is this path representation consistent with `CaseEvent.default_path_builder`? It doesn't look like it is to meas this uses type=case but the `default_path_builder` uses `type=testcase`.

I've said this before, but I think we need to rework the test path construction code in the report collection side given that two sides need to be aligned on the paths they create. Too many different test runners relying on the particular behavior of `default_path_builder` makes it really fragile. Aiming for post 1.0.

kohsuke (changes requested):
I believe the `make_test_path` function as it stands now breaks this test runner support because  test paths created by the `subset` command and `record tests` command are inconsistent.

I left a separate comment for how to fix this in a long run, but for the time being, I think the easiest fix is to have the `make_test_path` function produces `type=testcase`. Assuming Google Test doesn't report file names, that should do for now.

kohsuke (approved):
After a discussion today we decided to ship this code as-is to cut v1.0 and then issue a bug fix release in 2021.

PR #80, opened 2020-12-29
Run the release process when a release is published, not when it's cr…

kohsuke:
…eated

The 'create' trigger appears to kick in only when a release is created
from GitHub UI, and not when I create a tag locally and git-push it.

I believe this will change the workflow to trigger when the publish
button is pressed.

Reference: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -5,7 +5,7 @@ name: Upload Python Package
 
 on:
   release:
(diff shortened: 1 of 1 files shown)


ninjinkun (approved):
Looks good. Let's try whether it works when push the publish button.

PR #81, opened 2021-01-06
Move setup.py to setup.cfg

ninjinkun:
I heard modern Python projects use `setup.cfg` instead of `setup.py` https://www.python.org/dev/peps/pep-0390/. Following this way, we can remove requirements.txt and some lines. According to my research, it is compatible for Python 3.4 or later under using setuptools 30.3.0 or later.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


shuheiktgw:
@ninjinkun The job seems to fail 👀

shuheiktgw:
Ping me when the PR gets ready for review

ninjinkun:
@shuheiktgw Sorry, this PR is fixed now.

shuheiktgw (approved):
LGTM 👍

PR #83, opened 2021-01-07
[ST-657] shippable generic test runner support

kohsuke:
Made more robust by implementing a dedicated test path builder with error checks.

I also renamed 'generic' to 'file' to better represent what this code actually does. This also leaves room open to do more generic test runner support that allows more arbitrary test path construction.


launchable/test_runners/file.py
```diff
@@ -0,0 +1,49 @@
...
+    # read lines as test file names
+    for t in sys.stdin:
+        client.test_path(t)
+    client.run()
+
+
+@click.argument('reports', required=True, nargs=-1)
+@launchable.record.tests
+def record_tests(client, reports):
+    def path_builder(case: TestCase, suite: TestSuite, report_file: str) -> TestPath:
```
ninjinkun, line 22: I think it's better than `CaseEvent.default_path_builder`. Do you plan to copy this `path_builder` logic to `CaseEvent.default_path_builder` or remove `CaseEvent.default_path_builder`?

ninjinkun (approved):
This PR looks good.

launchable/test_runners/file.py
kohsuke: This logic builds a different looking test path compared to `default_path_builder`. I cannot replace the latter by the former without risking breaking every other test runner support. In effect, default path builder is untouchable.

I've mentioned to you in the past that the whole notion of default path builder is just fundamentally error prone because of that. So I'm +1 for removing the default path builder but for an entirely different reason.

I think we should talk. There's a disconnect here.

PR #84, opened 2021-01-08
[ST-648] Hides explicit session handling

ninjinkun:
To hide session ID from commands, and to handle multiple sessions, the CLI saves session ID into a file.

- Save a test session file into `~/.config/launchable/session.json`
- Remove `--session`  required on `subset` and `record tests`
  - `--session` is still remained. Users can specify it by themselvis.

### Session file format
```json
{ 
  "BUILD_ID:PID": "SESSION_ID"
}
```

### Basic flow
1. Start build, then remoe session file
2. Recrod session, store session ID into a session file
3. Subset tests with session ID from the session file
4. Record tests with session ID from the session file
5. Close session, then remove session file (This close session request is not yet implemented)


launchable/commands/record/session.py
```diff
@@ -33,7 +45,11 @@ def session(build_name):
 
         # what we print here gets captured and passed to `--session` in later commands
         click.echo("{}/{}".format(session_path, session_id))
-        return session_id
+
+        if save_session_file:
+            write_session(build_name, "{}/{}".format(session_path, session_id))
+        else:
+            return session_id
```
kohsuke, line 52: This caught my eyes. Why does the function stop returning a value when `--session-file` option is specified? What does the exit value from a `click.option` option do wrt Click? Nothing?

kohsuke (approved):
Looks good!

launchable/commands/record/session.py
ninjinkun: It was my previous mistake. It looks like nothing happens. I replace it with `click.echo`.

ninjinkun:
I'm working to modify this PR to run `record session` in `subset` and `recrod build`.

ninjinkun:
@kohsuke I added another code that runs `record session` in `subset` and `record tests`. 

* `subset` and `record tests`
  * If `--session` passed, use the param
  * If `--name` passed, search corresponding session from `~/.config/launchable/session.json`
    * If there is no session, run `record session` internally

Also, I added [responses](https://github.com/getsentry/responses) pip module to mock HTTP requests. Please review my additional code 🙏

launchable/commands/record/tests.py
```diff
@@ -26,17 +30,32 @@
...
-        return
+        if build_name:
+            try:
+                session_id = read_session(build_name)
+                if not session_id:
+                    session_id = create_session()
+            except SessionError as e:
+                session_id = create_session()
+        else:
+            raise click.UsageError(
```
kohsuke, line 57: Similarly it's lacking the error check for the case when both options are given.

launchable/commands/record/tests.py
```diff
@@ -26,17 +30,32 @@
...
+                'Missing option --build if you don\'t specify --session')
```
kohsuke, line 58: This may be a matter of taste, but I think it'd be slightly better to say "Either --build or --session has to be specified". Curios to hear a native speaker's take, like @awilkes 

Here is my motivation. You get this error when you didn't specify either options. This error message as it's currently stated makes me as a user feels like the first decision that I need to make is if I should specify the --session option or not.

launchable/utils/session.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class SessionError(Exception):
+    pass
+
+
+def read_session(build_name: str) -> Optional[str]:
+    if not session_file_path.exists():
+        raise SessionError(
+            "No session file. Please write session {}".format(session_file_path))
```
kohsuke, line 17: When I think about the situation in which the execution gets here, my mental picture is that somebody invoked `launchable subset --build` without ever calling `launchable record build`. I don't think asking users to "please write session $FILENAME" gets them on to the path of solving this problem.

To me, a better error message is "no session information recorded in $FILENAME. Did "launchable record  tests" run?"

kohsuke: How does this behavior compared to a situation when a file does exist but no session is recorded for the specified build name?

launchable/utils/session.py
```diff
@@ -0,0 +1,70 @@
...
+
+    with session_file_path.open('r') as json_file:
+        try:
+            data = json.load(json_file)
+            return data.get("{}:{}".format(build_name, os.getsid(os.getpid())))
+        except json.JSONDecodeError as e:
+            raise SessionError(
+                "Invalid session file. Please remove {}".format(session_file_path))
```
kohsuke, line 25: So again I'm thinking what situation would land here. Probably write has failed because of out of disk space? Any other situations? Maybe I'm missing that scenario, but I can't think of a situation where instructing users to remove this file is the right diagnosis. Especially given that this tends to happen on a CI system where such access is difficult.

Also, I suggest we chain the nested exception `e` so that it gets displayed. Imagine a random user dropping by our community and ask you he saw this error message and how he can solve it? Especially in case of an error, we need every bit of information to troubleshoot.

launchable/utils/session.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def write_session(build_name: str, session_id: str) -> None:
+    if session_file_path.exists():
+        with session_file_path.open('r') as json_file:
+            try:
+                data = json.load(json_file)
+            except json.JSONDecodeError as e:
+                raise SessionError(
+                    "Invalid session file. Please remove {}".format(session_file_path))
```
kohsuke, line 35: In this situation I feel a little more certain that the right thing to do is to simply clobber the file and write a new one. Maybe issue a warning at most.

launchable/utils/session.py
```diff
@@ -0,0 +1,70 @@
...
+    with session_file_path.open('r') as json_file:
+        try:
+            data = json.load(json_file)
+            return data.get("{}:{}".format(build_name, os.getsid(os.getpid())))
+        except json.JSONDecodeError as e:
+            raise SessionError(
+                "Invalid session file. Please remove {}".format(session_file_path))
+
+
+def write_session(build_name: str, session_id: str) -> None:
```
kohsuke, line 28: If the home directory doesn't exist (which happens typically in a Docker container), or if it's not writable, how does that error get reported? Does that give people enough clue how to solve it? Given how painful sometimes it is to change that, do we provide the escape hatch? Like switching to the explicit session passing with the `--session` option?

launchable/utils/session.py
```diff
@@ -0,0 +1,70 @@
...
+                raise SessionError(
+                    "Invalid session file. Please remove {}".format(session_file_path))
+    else:
+        session_file_path.parent.mkdir(parents=True, exist_ok=True)
+        session_file_path.touch()
+        data = {}
+
+    with session_file_path.open('w') as json_file:
+        data["{}:{}".format(build_name, os.getsid(os.getpid()))] = session_id
+        json.dump(data, json_file)
```
kohsuke, line 43: JSON is a fragile file format in the sense that failure to fully write it is gonna render the file completely unusable, so a more robust practice is to write to a temporary file, and replace it into the right name.

Actually, after examining this class more closely, I now think a single JSON file is a poor format mainly because of its fragility. The whole thing would be much more robust if each build name -> session mapping is written out as a separate file.

launchable/utils/session.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def remove_session(build_name: str) -> None:
+    """
+    Call it after closing a session
+    """
+    if not session_file_path.exists():
+        raise SessionError(
+            "No session file. Please write session {}".format(session_file_path))
```
kohsuke, line 52: I suspect this is a mistake. If you are trying to delete an entry that doesn't exist, shouldn't we just let that be no-op and move on?

launchable/utils/session.py
```diff
@@ -0,0 +1,70 @@
...
+    with session_file_path.open('r') as json_file:
+        try:
+            data = json.load(json_file)
+        except json.JSONDecodeError as e:
+            raise SessionError(
+                "Invalid session file. Please remove {}".format(session_file_path))
+
+    with session_file_path.open('w') as json_file:
```
kohsuke, line 60: Same atomic write & replace comment as above.

launchable/utils/session.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    if not session_file_path.exists():
+        raise SessionError(
+            "No session file. Please write session {}".format(session_file_path))
+    with session_file_path.open('r') as json_file:
+        try:
+            data = json.load(json_file)
+        except json.JSONDecodeError as e:
+            raise SessionError(
+                "Invalid session file. Please remove {}".format(session_file_path))
```
kohsuke, line 58: This also feels like we'd be better off just clobbering this file altogether. Discussed this more extensively above.

kohsuke (approved):
One last comment I leave from the perspective of "how/when does this design break and how do we troubleshoot that" is the problem of runaway data accumulation in the session file.

If somebody uses a static pre-provisioned build/test infrastructure, and if they fail to reliably close sessions, over time we accumulate a lot of dead entries into this JSON file. I think it's prudent to have code that garbage collect old entries. Say if we are using one entry per file format, then we can touch the file every time we read/write. If we haven't touched file for, say, 14 days, it's probably safe to delete that file.

I'll leave it up to you to decide how much of these robustness problem you tackle today vs tomorrow, but like I wrote in the CLI design principle, how something behaved when going gets tough plays a key role in shaping an impression, so I'll keep waving this "robust design, better error handling, self-supporting product" flag high.

launchable/utils/session.py
ninjinkun: Right. I agree JSON is fragile. I modified it to use multiple text files instead of a single JSON file.

launchable/utils/session.py
```diff
@@ -0,0 +1,70 @@
...
+    if session_file_path.exists():
+        with session_file_path.open('r') as json_file:
+            try:
+                data = json.load(json_file)
+            except json.JSONDecodeError as e:
+                raise SessionError(
+                    "Invalid session file. Please remove {}".format(session_file_path))
+    else:
+        session_file_path.parent.mkdir(parents=True, exist_ok=True)
+        session_file_path.touch()
```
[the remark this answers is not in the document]
ninjinkun: Okay. I removed this line because I modified `session.py` to write Session ID to file directly.

launchable/utils/session.py
ninjinkun: I modified it to no-op when the session file does not exist.

launchable/utils/session.py
ninjinkun: I added `LAUNCHABLE_SESSION_DIR` environment variable to allow to change the session directory. And notify how to use the variable as an exception message. Do you like this idea?

ninjinkun:
@kohsuke I modified storing session to use multiple text files instead of a single JSON file. Also `record build` removes the session directory on each run. I think it is more robust than the previous impl. Please final review!

ninjinkun:
Oh I found a problem
> Also record build removes the session directory on each run. 
It will remove ongoing sessions on where runs multi build run on the same instance.

ninjinkun:
Fixed

launchable/utils/session.py
```diff
@@ -0,0 +1,60 @@
...
+SESSION_DIR_KEY = 'LAUNCHABLE_SESSION_DIR'
+DEFAULT_SESSION_DIR = '~/.config/launchable/sessions/'
+
+
+def _session_file_dir() -> Path:
+    return Path(os.environ.get(SESSION_DIR_KEY) or DEFAULT_SESSION_DIR).expanduser()
+
+
+def _session_file_path(build_name: str) -> Path:
+    return _session_file_dir() / "{}:{}.txt".format(build_name, os.getsid(os.getpid()))
```
kohsuke, line 18: Operator overloading in Python, huh.

launchable/utils/session.py
kohsuke: Wow, thank you! I went ahead and tweaked the message a bit.

launchable/utils/session.py
kohsuke, line 18: Just one last thing, hopefully. build name is a user supplied string and there's no guarantee that it's safe as a file name.

Different ways to solve this, but my suggestion is to hash the build name.

kohsuke:
Let's ship this, except for that one thing where the build name might not be file system safe.

ninjinkun:
Alright, I fixed it to hash the filename. It's time to merge!

PR #86, opened 2021-01-10
Add `record tests minitest` command test

ninjinkun:
Testing `record tests minitest` with request mocking. 

- Run command via `CliRunner`
- Read test JUnitXML files on the repo
- Hijack request payload through request mock
- Compare request payload and expected result


tests/commands/test_record_tests_minitest.py
```diff
@@ -0,0 +1,31 @@
...
+    def setUp(self):
+        os.environ['LAUNCHABLE_TOKEN'] = self.launchable_token
+
+    @mock.patch('requests.request')
+    def test_record_test_minitest(self, mock_post):
+        runner = CliRunner()
+        result = runner.invoke(main, ['record', 'tests',  '--session', self.session, 'minitest', str(self.test_files_dir) + "/"])
+        eq_(result.exit_code, 0)
+        # TODO: uncompress data, then compare them as Python object
+        zipped_payload = b''.join(mock_post.call_args.kwargs['data'])
```
ninjinkun, line 27: @kohsuke I want to uncompress the gzipped payload to compare with an expected JSON. However, I can't uncompress that data with `zlib.uncompress`. Do you know how to uncompress the data compressed with `gzippen.py`?

tests/commands/test_record_tests_minitest.py
```diff
@@ -0,0 +1,31 @@
...
+
+        with self.result_file_path.open() as json_file:
+            json_obj = json.load(json_file)
+            eq_(zipped_payload, b''.join(compress((json.dumps(json_obj).encode(),))))
```
ninjinkun, line 32: I tried to compress an expected JSON data with `gzippen`. But its result is a bit different from the payload.

kohsuke: I think a better assertion is not to compare compressed content bit by bit, which makes the test sensitive to a lexical change in JSON (say, a whitespace after `.`), but instead decompress `zipped_payload` and tests equality of JSON trees (which hopefully is cheap enough in Python)

kohsuke: See https://github.com/launchableinc/cli/pull/87 for a test case that demonstrates this

ninjinkun: I see! `gzip. uncompress` works well.

tests/commands/test_record_tests_minitest.py
```diff
@@ -0,0 +1,32 @@
...
+class MinitestTest(TestCase):
+    launchable_token='v1:launchableinc/mothership:auth-token-sample'
+    session = '/intake/organizations/launchableinc/workspaces/mothership/builds/123/test_sessions/16'
+    test_files_dir = Path(__file__).parent.joinpath('../data/minitest/').resolve()
+    result_file_path = test_files_dir.joinpath('record_test_result.json')
+
+    def setUp(self):
+        os.environ['LAUNCHABLE_TOKEN'] = self.launchable_token
+
+    @mock.patch('requests.request')
```
kohsuke, line 22: You need to teach me what black magic happens behind this!!

kohsuke (approved):

ninjinkun:
Test passed!

tests/commands/test_record_tests_minitest.py
ninjinkun: Black magic is black magic🪄 https://docs.python.org/3/library/unittest.mock.html

PR #87, opened 2021-01-12
Added a quick test case to show gzipgen is working

kohsuke:
Might as well keep this now that I wrote it.

### DIFF
--- a/tests/utils/test_gzipgen.py
+++ b/tests/utils/test_gzipgen.py
@@ -1,4 +1,5 @@
 from launchable.utils.gzipgen import compress
+from nose.tools import eq_
 import gzip
 
 def test_compress():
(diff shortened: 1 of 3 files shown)


tests/utils/test_gzipgen.py
```diff
@@ -0,0 +1,9 @@
+from launchable.utils.gzipgen import compress
+import gzip
+
+def test_compress():
+    """Basic sanity test of """
+    encoded=b''.join(compress([b'Hello',b' ',b'world']))
+    msg = gzip.decompress(encoded)
+    print(msg)
+    assert msg==b'Hello world'
```
ninjinkun, line 9: Please use `eq_` from `from nose.tools import eq_` instead of `assert`.

kohsuke: So the difference between `eq_` and `assert` is that the latter might get optimized away with some compilation option?

ninjinkun (approved):

PR #88, opened 2021-01-12
Add Go test plugin

ninjinkun:
# Usage
## subset
`subset` command receives test names from stdin
```shell
go test -list . | launchable subset --name '#123' go_test > subset.txt
go test -run $(cat subset.txt | perl -anle 'print(join q{|}, map { q{^} . quotemeta . q{$} } @F)')
```

## record tests
`record tests` command read JUnit XML files.

```shell
go test -v 2>&1 | go-junit-report >  report.xml # convert output via go-junit-report 
launchable record tests --name '#123' go_test .
```

# Disucussion
Should the plugin command name be `go_test` or `go-test`?

### DIFF
--- a/launchable/test_runners/go_test.py
+++ b/launchable/test_runners/go_test.py
@@ -2,13 +2,14 @@ import sys
 import click
 from . import launchable
 
-
 @launchable.subset
 def subset(client):
     for case in sys.stdin:
         # Avoid last line such as `ok      github.com/launchableinc/rocket-car-gotest      0.268s`
         if not ' ' in case:
             client.test_path({'type': 'testcase', 'name': case.rstrip('\n')})
+
+    client.formatter = lambda t: "^{}$|".format(t)
     client.run()
 
 
--- a/launchable/test_runners/go_test.py
+++ b/launchable/test_runners/go_test.py
@@ -9,7 +9,8 @@ def subset(client):
         if not ' ' in case:
             client.test_path({'type': 'testcase', 'name': case.rstrip('\n')})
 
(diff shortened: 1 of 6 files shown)


kohsuke (approved):
Exciting!!

launchable/test_runners/go_test.py
```diff
@@ -0,0 +1,20 @@
+import sys
+import click
+from . import launchable
+
+
+@launchable.subset
+def subset(client):
+    for case in sys.stdin:
+        # Aboid last line such s `ok      github.com/launchableinc/rocket-car-gotest      0.268s`
```
awilkes, line 9:
```suggestion
        # Avoid last line such as `ok      github.com/launchableinc/rocket-car-gotest      0.268s`
```

ninjinkun: 😱 Thank you!

ninjinkun:
@kohsuke I found one more discussion point. This `subset` command outputs with white spaces such as `ExampleTest1 ExampleTest2 ExampleTest3`. However `go test -run` receive regexp only. It means customers needs. It means they need to convert the output to `^ExampleTest1$ |^ExampleTest2$|^ExampleTest3$` via shell script. Do you think we need to integrate the conversion into the CLI?

ninjinkun:
I changed `subset` output to regexp `^ExampleTest1$ |^ExampleTest2$|^ExampleTest3$`.

kohsuke:
The less the users have to do the better, so I think this makes sense :+1:

PR #92, opened 2021-01-13
go_test -> go-test

kohsuke:
After discussing with @awilkes , we came to the agreement that more conventional command name is probably `go-test`.

While this is technically a breaking change, we haven't really announced the Go test support yet, so I think it's OK to sneak this in and call it v1.1.2


ninjinkun (approved):
This is really nice 👍

PR #97, opened 2021-01-18
[ST-700] add unit test for googletest record tests

Konboi:
- Add googletest result files
- Add record test's unit test for googletest


ninjinkun (approved):
Looks good. Please merge it. Then, let's test `launchable subset googletest` command on other PR!

PR #98, opened 2021-01-19
Introduce Pipenv for development environment management

ninjinkun:
To improve the CLI tests, I need more pip development module dependencies to this repo (For example, I want to add [responses](https://github.com/getsentry/responses)). However, we don't have any way to force install new dependency to developers. I think it's time to introduce a development environment management.

[Pipenv](https://pipenv.pypa.io/en/latest/) is a popular module that provides development environment management and dependency management. It can create a virtual environment on your local and install dependencies there. If you want to know how to use it, please see new lines on `README.md`.


Pipfile
```diff
@@ -0,0 +1,24 @@
...
+
+[packages]
+launchable = {editable = true,path = "."}
+
+[scripts]
+build = "python setup.py sdist bdist_wheel"
+install = "pip install -U ."
+test = "python -m nose tests"
+lint = "flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics"
+lint-warn = "flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics"
```
Konboi, line 23: How did you decide these values?

Pipfile
```diff
@@ -0,0 +1,24 @@
...
+flake8 = "*"
+setuptools = ">=30.3.0"
+setuptools-scm = "*"
+wheel = "*"
+autopep8 = "*"
+nose = "*"
+importlib-metadata = "*"
+
+[packages]
+launchable = {editable = true,path = "."}
```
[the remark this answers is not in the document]
ninjinkun: We have already defined non-dev dependencies at `setup.cfg`. IIUC, they are used to PyPI installation and `pipenv install`.

Pipfile
ninjinkun: I copied these rules from a [GithubActions template](https://github.com/actions/starter-workflows/blob/main/ci/python-package.yml). We already have been using it on our Github Actions. I think we should consider what rule we should use later.

kohsuke (approved):
I believe we use `pipenv` in a part of mothership.

kohsuke:
Nice to see the development process & environment getting increasingly standardized.

Konboi (approved):
LGTM

PR #106, opened 2021-01-25
Replace mock with responses

ninjinkun:
SSIA


Konboi (commented):
I could confirm `pipenv run test` command be successful.
when I tried to test a specific test file, it failed, 
e.g) `python3 -m unittest tests/test_runners/test_googletest.py` 

Is not correct to test a specific file in this way? 
I sometimes want to test a specific file, so I want to know that.

ninjinkun:
At previous `pipenv run test`, you couldn't run a single test with the command becase I specified a test directory as `nosetests tests`. Just now, I merged main branch that includes replaced test runner from `nosetests` to `unittest` and I didn't specify a directory. In the latest commit, you can run `pipenv run test tests/test_runners/test_googletest.py`.

Konboi (commented):
I could check it! Thank you 🙏 .

LGTM 💯

PR #107, opened 2021-01-25
Replace nose with unittest

ninjinkun:
We decided to use unittest as this CLI's test framework. I removed nose dependencies and add some unittest codes.


Konboi (approved):
LGTM 🎉

PR #108, opened 2021-01-25
[ST-716] add debug option to print payload

Konboi:
I added debug option to print payloads for the record tests command.

I thought I would like to add debug option to global options, but the debug option only need for record tests.
So, I add debug options as the record tests command's option.


kohsuke (approved):
Thank you for taking care of this so quickly!!

ninjinkun (approved):

PR #110, opened 2021-01-26
[ST-726] fix handling for test cases result data

Konboi:
Even though `junitparser.TestCase.result` is a type of [List](https://github.com/weiwei/junitparser/blob/master/junitparser/junitparser.py#L668-L669), compared TestResult with the result that is a List directly, so always status was a Success status before.

I'm wondering that should we set success status to status's initial value. 
Will you please comment on it?

### DIFF
--- a/launchable/commands/record/case_event.py
+++ b/launchable/commands/record/case_event.py
@@ -44,8 +44,13 @@ class CaseEvent:
     def from_case_and_suite(cls, path_builder: TestPathBuilder, case: TestCase, suite: TestSuite, report_file: str, data: Dict = None) -> Dict:
         "Builds a JSON representation of CaseEvent"
 
-        status = CaseEvent.TEST_FAILED if case.result and any(isinstance(case.result, s) for s in (
-            Failure, Error)) else CaseEvent.TEST_SKIPPED if isinstance(case.result, Skipped) else CaseEvent.TEST_PASSED
+        status = CaseEvent.TEST_PASSED
+        for r in case.result:
+            if any(isinstance(r, s) for s in (Failure, Error)):
+                status = CaseEvent.TEST_FAILED
+                break
+            elif isinstance(r, Skipped):
+                status = CaseEvent.TEST_SKIPPED
 
         return {
             "type": cls.EVENT_TYPE,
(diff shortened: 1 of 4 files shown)


launchable/commands/record/case_event.py
```diff
@@ -44,8 +44,13 @@ def f(case: TestCase, suite: TestSuite, report_file: str) -> TestPath:
     def from_case_and_suite(cls, path_builder: TestPathBuilder, case: TestCase, suite: TestSuite, report_file: str, data: Dict = None) -> Dict:
         "Builds a JSON representation of CaseEvent"
 
-        status = CaseEvent.TEST_FAILED if case.result and any(isinstance(case.result, s) for s in (
-            Failure, Error)) else CaseEvent.TEST_SKIPPED if isinstance(case.result, Skipped) else CaseEvent.TEST_PASSED
+        status = CaseEvent.TEST_PASSED
```
ninjinkun, line 47: [IMO] You can refactor this code more readable. For example,
```python
def result_type(case):
  if any(any(isinstance(r, s) for s in (Failure, Error)) for r in case.result):
    return CaseEvent.TEST_FAILED
  elif any(isinstance(r, Skipped) for r case.result):
    return CaseEvent.TEST_SKIPPED
  else:
    return CaseEvent.TEST_PASSED
```

ninjinkun: > I'm wondering that should we set success status to status's initial value.
> Will you please comment on it?

Does this code resolve your concern?

ninjinkun:
This is a very important bug fix. I appreciate you.

launchable/commands/record/case_event.py
Konboi: I don't familiar with the Python way, but I want to avoid `if any(any(isinstance(r, s) for s in (Failure, Error)) for r in case.result):`

Because there are 2 loops and  `if` case into a line, it's difficult to understand the first time, I think.
What do you think about it?

ninjinkun: Yes, it wastes one more loop. But I think its performance does not matter in most cases. If we regard `for s in (Failure, Error)` as a constant, this is a `n + n` vs `n` problem, not `n^2`. At this time, you should choose a more readable and robust one. But if you think my idea is not good, please skip this. IMO less side effect is better in most cases.

kohsuke (approved):
Static typed languages win by the type check to catch errors like this pre-execution. Dynamic typed languages win by the ease of testing. I think we need to keep investing in testing as a team!

launchable/commands/record/case_event.py
Konboi: Sorry, I decided to keep this because I want to keep code and logic simple.
But, we have to keep consider the default value of status, so I'll add a comment about it and merge.

ninjinkun:
@Konboi Please add skip test case. I recommend splitting type detection function, then test it.

PR #111, opened 2021-01-26
[ST-723] Change gradle plugin output with `--tests` prefix

ninjinkun:
In Gradle plugin, `subset` command needs to convert the output such as 
```shell
./gradlew test $(cat subset.txt | perl -anle 'print map { sprintf(" --tests %s", $_) } @F')
```

To remove the above Perl one-liner, I added `--enable-format` option. It add `--tests` prefix to each test name like`--tests test_name1 --tests test_name2`.


ninjinkun:
See my discussion on the Jira ticket.

kohsuke (approved):
Thanks for a quick turn around.

This comment is largely a matter of taste, but I'd like to explore if there can be a better name. I'm approaching this from the perspective of "when somebody reads this command line option does it help them understand what this is". That is somebody trying to integrate Launchable to their Gradle project, and that is developers in their projects later trying to build some cursory understanding of what this Launchable thing is doing.

The problem with "enable" is that it doesn't really add any meaningful information, because every option is normally enabling something. Think about `--verbose` vs `--enable-verbose`. So in my mind that presence has a questionable value.

"format" is better but still pretty generic and vague. My first reaction to hearing "format" is that I get to control how the data is printed, which this option doesn't. It doesn't really say anything about how it's formatted (the crux of the vagueness), yet that is the raison d'etre of this option, namely to format the class name for Gradle command line.

My suggestions:
* `--for-command-line`
* `--command-line`
* `--with-prefix`
* `--prefixed`

@awilkes any thoughts?

Ultimately I believe the decision is  @ninjinkun's so I'm marking this PR as approved.

ninjinkun:
@kohsuke I removed `--enable-format` and changed default behavior because our customers haven't integrated `record tests gradle` yet. 

But on the other hand, I welcome naming discussion. Your `--verbose` vs `--enable-verbose` example is mind-blowing. Yes, it's redundant. `--with-prefix` and `--prefixed` sounds good to me.

kohsuke:
Best outcome!!

Issue #103, opened 2021-01-21
TypeError when file/glob does not match any file


astnbomb:
Running launchable record with a glob that does not match any file throws a TypeError:

```
$ launchable record tests --session /intake/organizations/one-two-eight/workspaces/128T_Pipeline/builds/jenkins-128T_Pipeline-11994/test_sessions/2411 googletest 'latest/gtest_*_results.xml'

Traceback (most recent call last):
  File "/usr/share/virtualenvs/ci_py36/bin/launchable", line 8, in <module>
    sys.exit(main())
  File "/usr/share/virtualenvs/ci_py36/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/share/virtualenvs/ci_py36/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/share/virtualenvs/ci_py36/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/share/virtualenvs/ci_py36/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/share/virtualenvs/ci_py36/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/share/virtualenvs/ci_py36/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/share/virtualenvs/ci_py36/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/share/virtualenvs/ci_py36/lib/python3.6/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/usr/share/virtualenvs/ci_py36/lib/python3.6/site-packages/launchable/test_runners/launchable.py", line 72, in record_tests
    click.echo("No matches found: " % root, err=True)
TypeError: not all arguments converted during string formatting
```

astnbomb:
Will add some logic to avoid calling Launchable when no test results exist as a workaround.

ninjinkun:
Thanks for your report! I'm going to research what causes the error. 

[...]

PR #113, opened 2021-01-27
Fix TypeError when file/glob does not match any file

ninjinkun:
https://github.com/launchableinc/cli/issues/103

Python's `"some_format" % some_string` needs `%s` in the format template such as `"some_format: %s" % some_string`. This is what caused #103 error.


Konboi (approved):
LGTM 👍

PR #114, opened 2021-01-27
Migrate docs to this repository

awilkes:
I followed [this guide](https://medium.com/@ayushya/move-directory-from-one-repository-to-another-preserving-git-history-d210fa049d4b) to do this.

Let me know if this is too disruptive and if I should just copy the directory over without retaining history.


kohsuke (approved):
History preservation is great. I checked and there appears to be no reference to customers, so we are good.

PR #116, opened 2021-01-28
Add trap command lines to documents

ninjinkun:
I found an issue. Most of CI system exits its run process when a test is failed. It means `record tests` after the test run doesn't run that time. To avoid this problem, customers must configure to run `record tests` after the test failed.

I propose the following setting. `trap` handles shell command signal. `0 1` are normal exit 0 and failure 1 (I think we can expect all test frameworks use this signal). 

```bash
trap "launchable record tests <BUILD NAME> [OPTIONS]" 0 1
bundle exec rails test -v $(cat tests.txt)
trap -
```

I want to add these instructions to the documents. I added some lines on this PR, but I think it's not enough. Please give me some idea about the following discussion point. 

# Discussions
## Other methods
There are other always run commands that depend on CI system. For example, Jenkins has [`post { always { ... } }`](https://www.jenkins.io/doc/book/pipeline/syntax/#post), GithubAction has [`if: ${{ always() }}`](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always), CircleCI has [`when: always`](https://circleci.com/docs/2.0/configuration-reference/#the-when-attribute). Do we introduce those solutions in the doc? If do this, what page do we edit?

## Other pages
I added some lines to `getting-started.md`. Do we need to add those lines to other pages? I think we should add lines to integration pages and `cli-reference.md`, but I'm not sure how I write those pages. Any idea is welcome.

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -13,6 +13,7 @@
   * [Nose \(Python\)](integrations/nose-python.md)
   * [Generic file based test runner](integrations/file.md)
   * [Dealing with custom test report format](integrations/convert-to-junit.md)
+  * [Always record tests](integrations/always-run.md)
 * [Security](security/README.md)
   * [Data Privacy and Protection](security/data-privacy-and-protection.md)
   * [Security Policies](security/information-security.md)
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -155,7 +155,7 @@ launchable record tests [OPTIONS] TESTRUNNER ...
 | `--session SESSIONID` | ID of the test session \(see `record session`\) | One of `--build` or `--session` is required |
 | `--base DIR` | See the discussion of `launchable subset --base` option. | No |
 
-This command reads JUnit XML report files produced by test runners and sends them to Launchable. This command require to run either tests succeed / failed
+This command reads JUnit XML report files produced by test runners and sends them to Launchable.
 
 Exactly how this command generates the subset and what's required to do this depends on test runners. For available supported `TESTRUNNER`, see [Integrations](integrations/)
 
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -155,7 +155,7 @@ launchable record tests [OPTIONS] TESTRUNNER ...
 | `--session SESSIONID` | ID of the test session \(see `record session`\) | One of `--build` or `--session` is required |
 | `--base DIR` | See the discussion of `launchable subset --base` option. | No |
 
-This command reads JUnit XML report files produced by test runners and sends them to Launchable.
+This command reads JUnit XML report files produced by test runners and sends them to Launchable. This command require to run either tests succeed / failed
 
 Exactly how this command generates the subset and what's required to do this depends on test runners. For available supported `TESTRUNNER`, see [Integrations](integrations/)
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -24,15 +24,13 @@ bundle install
 # ask Launchable which tests to run for this build
 launchable subset --build <BUILD NAME> [OPTIONS] > tests.txt
 
-# set trap to send test results to Launchable for this build either tests succeed/fail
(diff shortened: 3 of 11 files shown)


kohsuke (commented):
Thanks, this is one of the problems in the doc that I noticed while prepping for the demo, so it's great to see getting this addressed.

A few questions/suggestions:

* IIUC, signal `0` is an alias for `EXIT` trap, and I find it more readable to write it out as `EXIT` not `0`.
* What's the significant of signal 1? That appears to be `SIGHUP` but I'm not sure when that matters
* It seems to me that now if the `bundle exec rails test` is successful tests are not reported. Am I wrong?
* Passing the `launchable record tests` invocation as a string complicates the quotation if they need to be used. For example, I found it better to always doublequote `<BUILD_NAME>` to make it robust when the script runs outside CI. Combine this with the above, I ended up defining a function. WDYT?
* `trap -` is a neat trick! I didn't know that.

ninjinkun:
- trap "launchable record tests <BUILD NAME> [OPTIONS]" EXIT SIGHUP
- This script also records successful cases because If `bundle exec rails test` is succeeded, they output EXIT signal
  - I tested it at `rocket-car-minites`
- Okay. I understand the string problem. Do you mean we should add the following function, right? Pushing this idea further, I'm considering how I can do this in the CLI internal.
```bash
trap_always() {
  trap "$1" EXIT SIGHUP
}
```
- `trap -` removes all traps

kohsuke:
```shell
trap_always() {
  trap "$1" EXIT SIGHUP
}
```
No. The problem of double quotation is that you'd have to write something like `trap "launchable \"$BUILD_NAME\" ..." EXIT` so the function is to wrap the whole launchable invocation:

```
function record {
  launchable record tests --build "$BUILD_NAME" ...
}
trap record EXIT
```

Then,
> This script also records successful cases because If bundle exec rails test is succeeded, they output EXIT signal.  I tested it at rocket-car-minitest
I can't help but think there's some key disconnect in how trap works, but you say you tested this, so I'm confused. My understanding is that `trap` traps a signel received by the shell process, not the child processes forked by the shell. EXIT trap is thus when the shell process is exiting, not when a child process is exiting.

I tested this as follows:
```
% cat test.sh 
#!/bin/bash
trap "echo caught" EXIT
sleep 3
trap - EXIT
```
If I let this run to completion, I get no output, as I expect:
```
% ./test.sh 
```
If I hit Ctrl+C while `sleep` is running, I see the trap handler kick in:
```
% ./test.sh 
^Ccaught
```

Then on the question of what signals should be processed, I found [this](https://unix.stackexchange.com/questions/57940/trap-int-term-exit-really-necessary) to be relevant for portability. Probably prudent to trap SIGINT and SIGTERM.

Finally, with my `bash` I noticed that `trap -` doesn't work. Relevant info for you:
```
% cat foo.sh 
#!/bin/bash -ex
trap -

% ./foo.sh 
+ trap -
trap: usage: trap [-lp] [[arg] signal_spec ...]

% /bin/bash --version
GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

% cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"
```

ninjinkun:
From today's discussion with @kohsuke, we decided to recommend CIs' always run option as a first choice. Then we introduce `trap` as a second choice.

ninjinkun:
@kohsuke I modified the documents. Please review again.

kohsuke (approved):

ninjinkun:
@awilkes Do you have any suggestions about the content and English?

ninjinkun:
I'm going to merge it. You can modify the docs after merge.

PR #118, opened 2021-02-01
[ST-732] Maven plugin

Konboi:
SSIA

I wrote the document in another PR https://github.com/launchableinc/cli/pull/119, so please see it if you know the useage.

### DIFF
--- a/launchable/test_runners/maven.py
+++ b/launchable/test_runners/maven.py
@@ -18,7 +18,6 @@ def subset(client, source_roots):
     for root in source_roots:
         client.scan(root, '**/*', file2test)
 
-    client.separator = ','
     client.run()
 
 
--- a/launchable/test_runners/maven.py
+++ b/launchable/test_runners/maven.py
@@ -1,3 +1,25 @@
+import os
+import click
 from . import launchable
+
+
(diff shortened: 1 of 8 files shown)


launchable/test_runners/maven.py
```diff
@@ -0,0 +1,24 @@
+import os
+import click
+from . import launchable
+
+
+@click.argument('source_roots', required=True, nargs=-1)
+@launchable.subset
+def subset(client, source_roots):
+    def file2test(f: str):
+        if f.endswith('.java'):
```
ninjinkun, line 10: Do you think we need to support `.kt` and `.scala` like our Gradle plugin, don't you? IIRC, Maven also can handle Scala and Kotlin.

Konboi: You're right. I hadn't considered that.
the maven is described in the [kotilin official documentation](https://kotlinlang.org/docs/reference/using-maven.html), so I'll fix the same with the gradle plugin

Konboi: fixed: https://github.com/launchableinc/cli/pull/118/commits/240905b6a645bb23bf26d4240be26f3a99513041

ninjinkun (approved):
LGTM

PR #119, opened 2021-02-02
[ST-732] doc for the maven plugin

Konboi:
SSIA

### DIFF
--- a/docs/integrations/maven.md
+++ b/docs/integrations/maven.md
@@ -2,7 +2,7 @@
 
 ## Recording test results
 
-Apache Maven is a software project management and comprehension tool. See [Maven – Welcome to Apache Maven](https://maven.apache.org/)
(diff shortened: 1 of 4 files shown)


docs/integrations/maven.md
```diff
@@ -0,0 +1,30 @@
+# Maven
+
+## Recording test results
+
+Apache Maven is a software project management and comprehension tool. See [Maven – Welcome to Apache Maven](https://maven.apache.org/)
```
awilkes, line 5: This doc should probably mention Surefire, even if it is the default.

docs/integrations/maven.md
```diff
@@ -0,0 +1,30 @@
...
+
+After running tests, point to the directory that contains all the generated test report XML files. You can specify multiple directories, for example if you do multi-project build:
```
awilkes, line 7: This line says `for example if you do a multi-project build` but the example script includes only one directory. We want to make it clear that multiple directories are supported, so it would be good to update the example.

awilkes (approved):
Ship it!

docs/integrations/maven.md
Konboi: OK, I'll update

Konboi:
Thank you for your review, I'll ship it!

PR #125, opened 2021-02-03
Add ctest plugin

ninjinkun:
I added ctest plugin. The doc is here https://github.com/launchableinc/cli/pull/126

PR #126, opened 2021-02-03
Add CTest doc

ninjinkun:
This is a doc about CTest plugin https://github.com/launchableinc/cli/pull/125


docs/integrations/ctest.md
```diff
@@ -0,0 +1,39 @@
+# CTest
+
+## Recording test results
+
+CTest does not natively produce JUnit compatible test report files, so you should use an extra tool such as [`xsltproc`](http://xmlsoft.org/xslt/xsltproc.html) and [`ctest-to-junit`](https://github.com/rpavlik/jenkins-ctest-plugin/blob/master/ctest-to-junit.xsl) to convert them for use with Launchable.
```
kohsuke, line 5: As we all know, JUnit report files produced must contain key bits of information in the right way in order for the subset to work. Yet I notice that the documentation is somewhat open-ended about how to produce JUnit report files. Isn't that inviting a trouble, or did I miss something that makes this process actually more robust than I thought?

* A quick look at the referenced XSLT stylesheet shows test names are having their `-` replaecd to `_`. If we are matching with test names, that creates a problem, I'd think.
* Given the error prone nature of this process, I wouldn't use a phrase like "such as", which implies other tools are OK. Or perhaps be more specific about what this format needs to satisfy.

docs/integrations/ctest.md
```diff
@@ -0,0 +1,39 @@
...
+
+After running tests, point to the directory that contains all the generated test report XML files:
+
+```bash
+# run the tests however you normally do
+# `-T test` option output own format XML file
+ctest -T test --no-compress-output
+
+# convert CTest's XML format to JUnit format with XSLT converter
+xsltproc ctest-to-junit.xsl Testing/**/Test.xml > Testing/**/junit.xml
```
kohsuke, line 15: Pretty sure `> Testing/**/junit.xml` doesn't work.

kohsuke (changes requested):
And finally, this is just a thought, but given the error-proneness in converting to JUnit, and if CTest already has its native XML format, to me it makes sense to just switch to parsing that format directly.

It saves the integration smoother by saving users one extra step, it makes CSM easier by eliminating this problem entirely, and in terms of the implementation effort, there's SAX-based parsing code that I created some time ago and, I'm hoping that can be reused quickly.

I'll leave it up to you to decide how best to proceed -- you know the best what are the other priorities and the effort you can spend on this one.

From my perspective, the only "show stopper" change is a broken `xsltproc` invocation.

ninjinkun:
I agree to support CUnit's XML format. Instead of handle the XML format directory, I propose to use XSLT in the python code because we can continue to use `junitparser`:
```python
import lxml.etree as ET

dom = ET.parse(input)
xslt = ET.parse(xslt_file)
transform = ET.XSLT(xslt)
converted = transform(dom)
```

However, I notice that this XSLT feature requires `lxml` module depends on `libxml2`. Should we avoid `libxml2` dependency?

ninjinkun:
I fixed it to support CTest's XML!

kohsuke (approved):
Great to see that it's gotten better so quickly!

PR #133, opened 2021-02-05
Remove CTest support from GoogleTest plugin

ninjinkun:
We have the new CTest plugin. To avoid confusion, I remove CTest support in the GoogleTest plugin. IIRC, the feature is undocumented and no customer uses it.


Konboi:
oh, test failed 😢

ninjinkun:
Fixed!

Konboi (approved):
LGTM

PR #136, opened 2021-02-06
User friendliness improvements

kohsuke:
- If we are reading stdin but it's not piped from a file/process, issue a warning. In a rare case this is intended (someone interactively debugging), but more often this is just a misunderstanding of the way CLI works, and an user mistakenly thinks that the CLI is hanging.
- If the server reports an error and we fall back, tell users we are doing that, again to avoid a surprise.

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -1,6 +1,7 @@
 import click
 import json
 import os
+import sys
 from os.path import *
 import glob
 import gzip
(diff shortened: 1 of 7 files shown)


ninjinkun (approved):

ninjinkun (approved):
This `stdin` method aggregation is good.

ninjinkun:
I'm going to merge and release it.

PR #137, opened 2021-02-06
Set a time out

kohsuke:
... so that a call will never hang and block people's builds/tests.

The connection time out is set aggressively, given that we have the
acess terminated at ALB and it's hard to imagine the pick up taking
longer than milliseconds order.

The read time out needs to be lax at the moment, mainly for `subset`
that might have to wait for a model to load, which is currently costly.


ninjinkun (approved):

ninjinkun:
I'm going to merge and release it.

PR #138, opened 2021-02-08
Fix CTest to handle skipped case

ninjinkun:
According to our customer, CTest XML format don't have `<NamedMeasurement name="Execution Time" />` element in skipped tests. I fixed CTest plugin to handle missing that element. Also, I added a converter that handles a `notrun` case as a skipped case.

### DIFF
--- a/launchable/test_runners/ctest.py
+++ b/launchable/test_runners/ctest.py
@@ -52,32 +52,43 @@ def record_tests(client, source_roots):
         testsuite = ET.Element("testsuite", {"name": "CTest"})
         test_count = 0
         failure_count = 0
+        skip_count = 0
 
         for test in original_tree.findall("./Testing/Test"):
             test_name = test.find("Name")
-            duration = test.find(
-                "./Results/NamedMeasurement[@name=\"Execution Time\"]/Value")
-            measurement = test.find("Results/Measurement/Value")
-            testcase = ET.SubElement(testsuite, "testcase", {
-                                     "name": test_name.text, "time": duration.text, "system-out": measurement.text})
+            if test_name != None:
+                duration = test.find(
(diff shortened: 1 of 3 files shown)


ninjinkun:
I've tested this change with the customer's data. But I can't share it in this public repository.

Konboi:
> I've tested this change with the customer's data. But I can't share it in this public repository.

I agree with you. 
But we should add a unit test for skip case and we can add a common skip test to a rocket car repo, can't you?

ninjinkun:
No, I couldn't find the way to skip Google Test in the rocket-car-googletest.

ninjinkun:
Okay, I'm going to add a test with anonymizing the customer's data.

launchable/test_runners/ctest.py
```diff
@@ -52,32 +52,44 @@ def parse_func(p: str) -> ET.Element:
...
-                "./Results/NamedMeasurement[@name=\"Execution Time\"]/Value")
-            measurement = test.find("Results/Measurement/Value")
-            testcase = ET.SubElement(testsuite, "testcase", {
-                                     "name": test_name.text, "time": duration.text, "system-out": measurement.text})
+            if test_name != None:
+                duration = test.find(
+                    "./Results/NamedMeasurement[@name=\"Execution Time\"]/Value")
+                measurement = test.find("Results/Measurement/Value")
+                stdout = measurement.text if measurement != None else ''
+                duration = duration.text if duration != None else '0'
```
Konboi, line 64: It's safe 👍

ninjinkun:
@Konboi I modified an existing test case to skip case based on the customer's data.

Konboi (approved):
LGTM 🙆‍♀️

PR #139, opened 2021-02-08
Fix test path order

ninjinkun:
Fix correct test path order from `class, file, testcase` to `file, class, testcase`


shuheiktgw (approved):
👍

PR #142, opened 2021-02-12
[ST-788] Split test results uploading 

ninjinkun:
Launchable API responded 460 status code because of many recording test case request. To resolve this error, split test results uploading.

Included features are:
- Suppress no test cases request
- Defer XML parsing exceptions to unblock sending other XMLs


launchable/commands/record/tests.py
```diff
@@ -140,47 +144,75 @@ def payload():
...
                 for d in f:
                     print(d)
                     yield d
 
-            payload = printer(payload()) if debug else payload()
-
-            # str -> bytes then gzip compress
-            payload = (s.encode() for s in payload)
-            payload = compress(payload)
+            def send(payload: Generator[TestCase, None, None]) -> None:
```
kohsuke, line 178: Love those typing hints as generators can get confusing pretty quickly for me.

launchable/commands/record/tests.py
```diff
@@ -140,47 +144,75 @@ def payload():
...
+                }
+                client = LaunchableClient(token)
                 res = client.request(
                     "post", "{}/events".format(session_id), data=payload, headers=headers)
                 res.raise_for_status()
-                res = client.request("patch", "{}/close".format(session_id), headers=headers)
+
+            try:
+                splitted_cases = splitter(printer(testcases(
+                    self.reports)), CASE_POST_CHUNK) if debug else splitter(testcases(self.reports), CASE_POST_CHUNK)
```
kohsuke, line 194: Nit: at least `splitter(printer(testcases(self.reports)) if debug else testcases(self.reports), CASE_POST_CHUNK)` for a strength reduction. Beyond that is subjective, but to me the following code is more readable:
```
tc = testcases(self.reports)
if debug: tc=printer(tc)
for chunk in splitter(tc):
    ...
```

tests/test_runners/test_minitest.py
```diff
@@ -19,3 +20,23 @@ def test_record_test_minitest(self):
 
         expected = self.load_json_from_file(self.result_file_path)
         self.assert_json_orderless_equal(expected, payload)
+
+    @responses.activate
+    def test_record_test_minitest_chunked(self):
+        # override max chunk
+        module = sys.modules['launchable.commands.record.tests']
+        module.CASE_POST_CHUNK = 5
```
kohsuke, line 28: Don't we need to restore this value at the end of the test so as not to influence other tests?

Alternatively, given that this is only used locally, maybe easier to turn this into a hidden option. That way a test will avoid a side effect, and the option might be marginally useful as a last resort in some desparate situations, too.

kohsuke (approved):
Looks good. A few minor comments left separately.

I think this is the right time to use sessions in requests: https://requests.readthedocs.io/en/master/user/advanced/

With this change, a typical client will end up making 10s or even 100s of HTTPS connections to our endpoint. Between the chatiness of HTTPS connection setup and the TCP slow start, in a large latency network, I think this is going to add a significant overhead to the data transfer.

IIUC, using session in request would bring in connection pooling, which will make this basically go as fast as the "send everything over a singlle request" case.

ninjinkun:
@kohsuke Thanks for your good suggestion! I created a new PR for integrating reqests's sessions as you said https://github.com/launchableinc/cli/pull/143

PR #143, opened 2021-02-15
Send HTTP requests with sessions

ninjinkun:
From the latest change https://github.com/launchableinc/cli/pull/142, `record tests` command becames running many HTTP requests. To reduce those requests' overheads, introduce [requests's sessions](https://requests.readthedocs.io/en/master/user/advanced/).

This PR also includes Cypress tests refactoring because [`responses`](https://pypi.org/project/responses/) automatically mocks the session requests.

### DIFF
--- a/launchable/commands/record/tests.py
+++ b/launchable/commands/record/tests.py
@@ -8,6 +8,8 @@ from junitparser import JUnitXml, TestSuite
 import xml.etree.ElementTree as ET
 from typing import Callable, Generator, List, Union
 from itertools import repeat, starmap, takewhile, islice
+
+import requests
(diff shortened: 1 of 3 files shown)


kohsuke (commented):
I think a better abstraction is to make session an internal concern of `LaunchableClient`, and reuse `LaunchableClient` instead of creating new one for each request.

That way, every invocation of LaunchableClient automatically benefits from sessions, such as implicit test session creation.

ninjinkun:
@kohsuke Got it. I modified the line.

kohsuke (approved):
Perfect!

PR #144, opened 2021-02-17
Set subset request timeout longer

ninjinkun:
Temporarily extend the timeout because the subset API response has been slow.


shuheiktgw (approved):
👍

PR #145, opened 2021-02-17
Add proxy support in `record commit` command

ninjinkun:
This PR will propose after Ingerster PR merged.


kohsuke:
Oh I can actually see now that that omission was intentional, but I'm gonna leave it in because it's eyebrow raising, and `os.system` would be OK with there being exttra space.

kohsuke:
@awilkes wants to get this shipped today so I'm going to take this over

PR #147, opened 2021-02-18
Fix `record commit` command

ninjinkun:
My fault, `os.environ` isn't callable. I use `os.getenv()` instead of that.


shuheiktgw (approved):
👍

PR #149, opened 2021-02-18
Run Mypy on CI

ninjinkun:
[Mypy](http://mypy-lang.org/) is a type checker for Python. To force checking types on PRs, run Mypy on GithubAction.


Konboi (approved):
LGTM

Good Work 👍

PR #152, opened 2021-02-19
Add example POST bodies

awilkes:
These were captured by running commands locally through a proxy. Soon we will have an audit log; in the meantime, this is intended to answer infosec questions about what we collect.


kohsuke (approved):
Looks good, thanks for doing that.

Do we intend to keep this maintained up to date? If so that's something we need to be mindful of.

awilkes:
Indeed, that is definitely something to be mindful of. Perhaps we can delete this page after audit logging ships: we could export audit logs for each GitHub Actions run in the `examples` repository, which would provide examples that are always up-to-date!

PR #153, opened 2021-02-22
[ST-796] Introduce logging logger and print audit log

Konboi:
### abstract

I introduced a logger used by logging and set log levels default and audit. add a global option `--log-level` to change level for printing logs. when we only set the log level, the logger print log. 

I'd like to replace the @command.echo log output with this logger in the future.

#### example

when we set `log-level`, logs are printed like below,

```
$ LAUNCHABLE_TOKEN=XXX launchable --log-level audit subset --build sample-1  --target 10% maven ./src/test/java/com/example/app/App*
AUDIT:launchable:send request method:post path:/intake/organizations/<ORG>/workspaces/<WORKSPACE>/builds/sample-1/test_sessions headers:{'Content-Type': 'application/json'} args:{'timeout': (5, 60)}
AUDIT:launchable:send request method:post path:/intake/organizations/<ORG>/workspaces/<WORKSPACE>/subset headers:{'Content-Type': 'application/json', 'Content-Encoding': 'gzip'} args:{'data': b'\x1f\x8b\x08\x00C^3`\x02\xff\xabV*I-.\tH,\xc9(V\xb2R\x88\x8e\xd5QP*I,JO-\x01\xf2\x0c\xf4\x0c\x81\xdc\xe2\xd4\xe2\xe2\xcc\xfc< \xbfZ)3\x05H)\x19\x9b+\xd5\xd6\x02\x00\xd1bG\xaa9\x00\x00\x00', 'timeout': (5, 180)}
```

#### about log level

I was going to set the highest level to the audit log level. However, when I use some times, I found it a little noisy that if we set the error to log level, the audit logs were printed.
Also, when we want to show the audit logs, there weren't any errors usually. so error log level and audit log level behavior don't conflict.
So I changed log levels from audit>critical>error>warn>info>debug to critical>error>warn>audit>info>debug.

### DIFF
--- a/launchable/__main__.py
+++ b/launchable/__main__.py
@@ -15,7 +15,7 @@ from .utils import logger
 @click.option(
     '--log-level',
     'log_level',
-    help='Set logger log level.\nlevels are CRITICAL > ERROR > WARNING > AUDIT > INFO > DEBUG.',
+    help='Set logger\'s log level (CRITICAL, ERROR, WARNING, AUDIT, INFO, DEBUG).',
     type=str,
     default=logger.LOG_LEVEL_DEFAULT_STR,
 )
--- a/launchable/__main__.py
+++ b/launchable/__main__.py
@@ -15,7 +15,7 @@ from .utils import logger
 @click.option(
     '--log-level',
     'log_level',
-    help='Set logger log level.\nlevels are AUDIT > CRITICAL > ERROR > WARNING > INFO > DEBUG.',
+    help='Set logger log level.\nlevels are CRITICAL > ERROR > WARNING > AUDIT > INFO > DEBUG.',
     type=str,
     default=logger.LOG_LEVEL_DEFAULT_STR,
 )
--- a/launchable/__main__.py
+++ b/launchable/__main__.py
@@ -1,16 +1,28 @@
 from .version import __version__
 import click
 import importlib
+import logging
 from os.path import dirname, basename, join
 from glob import glob
 from .commands.record import record
 from .commands.subset import subset
 from .commands.verify import verify
+from .utils import logger
 
 
 @click.group()
 @click.version_option(version=__version__, prog_name='launchable-cli')
-def main():
+@click.option(
+    '--log-level',
+    'log_level',
+    help='Set logger log level.\nlevels are AUDIT > CRITICAL > ERROR > WARNING > INFO > DEBUG.',
+    type=str,
+    default=logger.LOG_LEVEL_DEFAULT_STR,
+)
+def main(log_level):
+    level = logger.get_log_level(log_level)
+    logging.basicConfig(level=level)
+
     # load all test runners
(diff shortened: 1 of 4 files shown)


Konboi:
🤔 🤔 🤔 

```
$ pipenv run test tests/utils/test_logger.py
..
Ran 2 tests in 0.001s

OK
```

```
$ pipenv run test
...............................b'Hello world'
.F.
FAIL: test_log_level_audit (tests.utils.test_logger.LoggerTest)
  File "/Users/yabuki-ryosuke/.brew/Cellar/python@3.9/3.9.1_6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/unittest/mock.py", line 1337, in patched
    return func(*newargs, **newkeywargs)
  File "/Users/yabuki-ryosuke/src/github.com/launchableinc/cli/tests/utils/test_logger.py", line 31, in test_log_level_audit
    self.assertEqual(stderr.getvalue(
AssertionError: '' != 'AUDIT:launchable:audit\nCRITICAL:launchab[58 chars]rn\n'
+ AUDIT:launchable:audit
+ CRITICAL:launchable:critical
+ ERROR:launchable:error
+ WARNING:launchable:warn


Ran 34 tests in 0.303s

FAILED (failures=1)
```

launchable/__main__.py
```diff
@@ -1,16 +1,28 @@
...
+@click.option(
+    '--log-level',
+    'log_level',
+    help='Set logger\'s log level (CRITICAL, ERROR, WARNING, AUDIT, INFO, DEBUG).',
+    type=str,
+    default=logger.LOG_LEVEL_DEFAULT_STR,
+)
+def main(log_level):
+    level = logger.get_log_level(log_level)
+    logging.basicConfig(level=level)
```
ninjinkun, line 24: @Konboi  I found this line affects test failure! This is because this line runs to set the config globally before `test_logger.py` tests run.

Konboi: Ohhh, Thank you so much!!! I'll try to avoid this method.

tests/utils/test_logger.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def setUp(self):
+        self.default_root_handler = logging.root
+
+        # reset logging handelr for test
+        for handler in logging.root.handlers[:]:
+            logging.root.removeHandler(handler)
+
+    def tearDown(self):
+        logging.root = self.default_root_handler
```
ninjinkun, line 21: I guess this line can't reset the root handler because you copied the root's reference. `self.default_root_handler` may be removed by the above lines. You should deep copy `logging.root` or `logging.root.handlers`.

Konboi: thank you, I'll check it

ninjinkun:
I left one comment. All others look good.

ninjinkun (approved):
Looks good!

PR #158, opened 2021-03-03
Avoid JUnitParser's value escaping 

ninjinkun:
JUnitPerser escape value in the JUnit XML by default. So, we need non escaped values in test names. Therefore, I modified the value handling to retrieve element value directly.


Konboi (commented):
Looks good!! But, I think that we should add escaped strings to some test data and check it isn't escaped to avoid reverting this change.

What do you think?

ninjinkun:
I agree. I'm going to add that case.

ninjinkun:
@Konboi I added two test names including escapable charactors.

Konboi (approved):
LGTM

PR #159, opened 2021-03-05
[ST-780] Removing the line number field

kohsuke:
We discussed and decided while back that this perhaps belong to the data bag but not in the test path.


Konboi:
Humm, JSON's format was changed so need to fix tests. 
but, it's a little trouble because if the request(or response) format will change, we need to fix all tests request(or response) every time.

kohsuke:
Oops looks like I broke tests left & right

PR #160, opened 2021-03-08
support when base has magic

Konboi:
before cli didn't expand the base path that includes magic words e.g `*`, `?`, etc
so, if the user sets the path that includes magic words, subset path will be broken.

### before

```
$ launchable subset <SUBSET ARGS> '**/src/test/java'

--tests .com.launchableinc.rocket_car_gradle.AppTest2 --tests .com.launchableinc.rocket_car_gradle.AppTest --tests .com.launchableinc.rocket_car_gradle.sub.AppTest3 --tests ava.com.launchableinc.rocket_car_gradle.utils.UtilsTest
```

### after

```
$ launchable subset <SUBSET ARGS> '**/src/test/java'

--tests com.launchableinc.rocket_car_gradle.AppTest2 --tests com.launchableinc.rocket_car_gradle.AppTest --tests com.launchableinc.rocket_car_gradle.sub.AppTest3 --tests com.launchableinc.rocket_car_gradle.utils.UtilsTest
```

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -159,7 +159,8 @@ def subset(context, target, session_id, base_path: str, build_name: str):
 
             for b in glob.iglob(base):
                 for t in glob.iglob(join(b, pattern), recursive=True):
-                    path = path_builder(t[len(b)+1:])  # type: ignore
+                    if path_builder:
+                        path = path_builder(t[len(b)+1:])
                     if path:
                         self.test_paths.append(self.to_test_path(path))
 
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -157,16 +157,9 @@ def subset(context, target, session_id, base_path: str, build_name: str):
                     return file_name
                 path_builder = default_path_builder
 
-            for t in glob.iglob(join(base, pattern), recursive=True):
-                if path_builder:
-                    if glob.has_magic(base):
-                        for b in glob.glob(base):
-                            if b in t:
-                                path = path_builder(t[len(b)+1:])
-                                break
-                    else:
-                        path = path_builder(t[len(base)+1:])
-
+            for b in glob.iglob(base):
+                for t in glob.iglob(join(b, pattern), recursive=True):
+                    path = path_builder(t[len(b)+1:])  # type: ignore
                     if path:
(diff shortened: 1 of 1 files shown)


launchable/commands/subset.py
```diff
@@ -159,7 +159,14 @@ def default_path_builder(file_name):
 
             for t in glob.iglob(join(base, pattern), recursive=True):
                 if path_builder:
-                    path = path_builder(t[len(base)+1:])
+                    if glob.has_magic(base):
+                        for b in glob.glob(base):
```
kohsuke, line 163: This fores `glob.glob(base)` to be evaluated repeatedly which is costly and unnecessary.

How about we evaluate `glob.igob(base)` then for each path it returns we evaluate `glob.iglob(x,pattern)` which minimizes directory traversal?

Konboi: Yes, indeed.
I'll fix it with your approach

launchable/commands/subset.py
```diff
@@ -157,9 +157,9 @@ def default_path_builder(file_name):
                     return file_name
                 path_builder = default_path_builder
 
-            for t in glob.iglob(join(base, pattern), recursive=True):
-                if path_builder:
-                    path = path_builder(t[len(base)+1:])
+            for b in glob.iglob(base):
+                for t in glob.iglob(join(b, pattern), recursive=True):
+                    path = path_builder(t[len(b)+1:])  # type: ignore
```
ninjinkun, line 162: Why did you put `type: ignore`? Although I assume `path_builder` is not nullable in this line, you should put a type guard such as `if path_builder is not None` or refactor code itself to prepend destruction in the future.

Konboi: > Although I assume path_builder is not nullable in this line

I think so. because in [line#148](https://github.com/launchableinc/cli/pull/160/files#diff-4c62e35fb4e3b2b6bff8593107cdddd1fd47eff90293edc9d838c9b2f3cff94aR148) define path_builder if it is None.
`pipenv run type` failed. so I wrote `# type: ignore`

but I'll try an another way to solve it.

ninjinkun (approved):
LGTM

PR #163, opened 2021-03-09
Demonstration of the plugin loading mechanism

kohsuke:
To allow some customers to maintain custom test runner support more easily, define a mechanism to load test runner support from arbitrary user specified directory. This is also handy for rapid experiments.


Konboi (approved):
LGTM 👍 

Awsome feature!!

PR #165, opened 2021-03-11
Add Python 3.9 to testing matrix

kohsuke:
3.9.2 appears to be the current latest, so it should be in the test matrix.


Konboi (approved):
LGTM

PR #167, opened 2021-03-15
[ST-831] subset diff

Konboi:


Konboi:
I adopted split option
https://github.com/launchableinc/cli/pull/169

PR #169, opened 2021-03-16
[ST-831] support subset --split option

Konboi:
`--split` option helps to introduce subsetting. if test framework support excluding, you should use it and subsetting list.
but some test frameworks don't support excluding so I add this option.

e.g) I tried the `--split` option to a maven project.

there are two test files
```sh
$ tree ./src/test/java/com/launchableinc/rocket_car_maven
./src/test/java/com/launchableinc/rocket_car_maven
├── App2Test.java
└── AppTest.java
```

first, tried to subset without the split option to check the subsetting result

```sh
$ launchable subset --build maven-1 --target 10% maven ./src/test/**
com.launchableinc.rocket_car_maven.AppTest
```

next, tried to subset with the split option and check results.

```sh
$ launchable subset --build maven-1 --target 10% --split subset.txt remainder.txt maven ./src/test/**
$ cat subset.txt
com.launchableinc.rocket_car_maven.AppTest
$ cat remainder.txt
com.launchableinc.rocket_car_maven.App2Test
```

that's it

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -216,7 +216,7 @@ def subset(context, target, session_id, base_path: str, build_name: str, rest: s
 
             # regardless of whether we managed to talk to the service
             # we produce test names
-            if rest != "":
+            if rest is not None:
                 rests = []
 
                 subset = [self.formatter(t) for t in output]
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -226,8 +226,8 @@ def subset(context, target, session_id, base_path: str, build_name: str, rest: s
                         rests.append(p)
 
                 if len(rests) == 0:
-                    click.echo(click.style(
-                        "Warning: remainder files don't exists.", fg='yellow'),  err=True)
+                    # no tests will be in the "rest" file. but add a test case to avoid failing tests using this
+                    rests.append(subset[-1])
 
                 open(rest, "w+").write(self.separator.join(rests))
 
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -45,8 +45,15 @@ from .record.session import session
     type=str,
     metavar='BUILD_NAME'
 )
+@click.option(
+    '--split',
+    'split',
+    help='output subset and remainder to file',
+    nargs=2,
+    type=str,
+)
 @click.pass_context
-def subset(context, target, session_id, base_path: str, build_name: str):
+def subset(context, target, session_id, base_path: str, build_name: str, split):
     token, org, workspace = parse_token()
 
     if session_id and build_name:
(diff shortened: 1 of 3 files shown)


awilkes:
Hey @Konboi, thanks for working on this so quickly. Please accept my apologies for repeated design changes, but @kohsuke and I discussed this interface today and thought that a `--rest <file>` option (e.g. `launchable subset --build [build] --rest remainder.txt [profile] [input] > subset.txt`) would fit into our existing implementation model more seamlessly.

This is because we want to recommend `--rest` as the first step of rolling out Launchable. Then, once sufficient confidence has been achieved, they can simply remove the `--rest` option from their `subset` call.

Could you make that change?

Konboi:
Thank you for your comment.

No, problem, because I feel your new suggested command option is better than I implemented and it's not difficult to fix!

I'll chage it 👍

launchable/commands/subset.py
```diff
@@ -199,18 +205,33 @@ def run(self):
...
                         click.echo(e, err=True)
-                    click.echo(click.style("Warning: the service failed to subset. Falling back to running all tests", fg='yellow'), err=True)
-
+                    click.echo(click.style(
+                        "Warning: the service failed to subset. Falling back to running all tests", fg='yellow'), err=True)
 
             # regardless of whether we managed to talk to the service
             # we produce test names
-            click.echo(self.separator.join(self.formatter(t) for t in output))
+            if rest != "":
```
kohsuke, line 219: A little surprising that the default value is empty string, not `None`!

launchable/commands/subset.py
```diff
@@ -199,18 +205,33 @@ def run(self):
...
+
+                subset = [self.formatter(t) for t in output]
+                for test_path in self.test_paths:
+                    p = self.formatter(test_path)
+                    if p not in subset:
+                        rests.append(p)
+
+                if len(rests) == 0:
+                    click.echo(click.style(
+                        "Warning: remainder files don't exists.", fg='yellow'),  err=True)
```
kohsuke, line 230: Hmm, it's really interesting what we should do in this case!

Producing an empty file does make sense, but then when I think about, say, Gradle, if this file is empty we'd end up running all the tests, as opposed to none. Maybe we issue a warning and then put one test case in here, as an approximation of "empty test set"?

The point of issuing a message is to drive some actions, but in this case it's not clear what we want the user to do in reaction to this. IOW, what would be an actionable message? If there's no action to be taken, is there any point in telling them?

Either way, the message needs to change. It's not the files that do not exist as they will be created anyway. The warning is that there's no test in the "rest" set. So maybe "Warning: no tests will be in the "rest" file {}"?

I'm inclined to:
- move the last test case from subset to the rest to avoid empty file
- remove the warning
- maybe set something in the subset request so that we can deal with this situation on the server side to understand the extent of this problem, if any.

kohsuke (approved):
I'm leaving +1 for this to avoid blocking the delivery of this feature, but there are some interesting corner cases to think about, not necessarily now.

Konboi:
Thank you for review, but CI failed so I'll merge after fix it.

launchable/commands/subset.py
Konboi: it is my miss understanding and test failed 🙇

launchable/commands/subset.py
Konboi: > move the last test case from subset to the rest to avoid empty file
> remove the warning


I agree that.
I felt a little strange before but I’ll be changed back to avoid test fail.

PR #170, opened 2021-03-18
[ST-828] robot plugin

Konboi:


kohsuke (approved):
Thanks. This is great.

I think this example shows that we are embedding JUnit as an assumption too deep in our system. You are converting XML of one format (Robot) into another (JUnit), only to have it converted into the final form we want (CaseEvent).

We should start thinking about tweaking the abstraction so that the abstraction is to parse files into CaseEvents, where the default implementation is to parse JUnit XML, but different parsers can be plugged into this to parse this format, for example.

Konboi:
thank you for your review.

> We should start thinking about tweaking the abstraction so that the abstraction is to parse files into CaseEvents, where the default implementation is to parse JUnit XML, but different parsers can be plugged into this to parse this format, for example.

I agree with your opinion, now record/tests.py depends JUnitXML format but Just as you said that we only need CaseEvents or Object that can transform to testPath finally.
anyway, I keep this in my mind.

PR #173, opened 2021-03-22
add rspec plugin

yoshiori:
Added the Rspec plugin, which is almost the same as minitest.
I found a bug in minitest and fixed it.


Konboi (approved):
LGTM 🚀

PR #175, opened 2021-03-23
[ST-830] add time option for absolute time-based subsetting

Konboi:
I support absolute time-based subsetting.

e.g)
```sh
launchable subset --time 30m <SUBSET ARGS>
```

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -6,7 +6,7 @@ from os.path import *
 import glob
 import gzip
 from typing import Callable, Union, Optional
-from ..utils.click import PERCENTAGE
+from ..utils.click import PERCENTAGE, DURATION
 from ..utils.env_keys import REPORT_ERROR_KEY
 from ..utils.http_client import LaunchableClient
 from ..utils.token import parse_token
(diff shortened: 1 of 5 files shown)


launchable/commands/subset.py
```diff
@@ -170,6 +175,25 @@ def default_path_builder(file_name):
...
+                }
+            }
+
+            if target is not None:
+                payload["target"] = target
+            elif duration is not None:
+                payload["goal"] = {
+                    "type": "subset-by-absolute-time",
+                    "duration": duration,
+                }
```
kohsuke, line 193: The new `goal` property is capable of representing every kind of goal, including the percentage based target, so for consistency don't we want to just get rid of `payload["target"]` altogether and always stick to `payload["goal"]` ?

launchable/utils/click.py
```diff
@@ -51,3 +68,27 @@ def emoji(s: str, fallback: str = ''):
     Returns 's' in an environment where stdout can deal with emojis, but 'fallback' otherwise.
     """
     return s if EMOJI else fallback
+
+
+def convert_to_seconds(s: str):
```
kohsuke, line 73: Nice touch!

kohsuke (approved):

launchable/commands/subset.py
Konboi: I thought that, and there might be no problem, but changing request field is a little big change.
so I thought that we should release gradually.

Next, I'll change this field and release.

PR #176, opened 2021-03-24
[ST-839] set test framework  name to header

Konboi:
set test-runner name when cli requests to the API.

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -198,7 +198,8 @@ def subset(context, target, session_id, base_path: str, build_name: str, rest: s
                     path = "/intake/organizations/{}/workspaces/{}/subset".format(
                         org, workspace)
 
-                    client = LaunchableClient(token)
+                    client = LaunchableClient(
+                        token, test_runner=context.invoked_subcommand)
                     # temporarily extend the timeout because subset API response has become slow
                     # TODO: remove this line when API response return respose within 60 sec
                     timeout = (5, 180)
--- a/launchable/utils/http_client.py
+++ b/launchable/utils/http_client.py
@@ -41,6 +41,6 @@ class LaunchableClient:
         }
 
         if self.test_runner != "":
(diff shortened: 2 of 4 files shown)


launchable/utils/http_client.py
```diff
@@ -34,7 +35,12 @@ def request(self, method, path, **kwargs):
...
     def _headers(self):
-        return {
+        h = {
             "User-Agent": "Launchable/{} (Python {}, {})".format(__version__, platform.python_version(), platform.platform()),
-            'Authorization': 'Bearer {}'.format(self.token)
+            'Authorization': 'Bearer {}'.format(self.token),
         }
+
+        if self.test_runner != "":
+            h["Launchable-Test-Runner"] = self.test_runner
```
Konboi, line 44: What do you think this field name?

yoshiori: I recommend you to use `launchable-test-runner`.
I want to use a name that can also be used for http/2 for future reference.

https://tools.ietf.org/html/rfc7540#section-8.1.2
>  A request or response containing uppercase  header field names MUST be treated as malformed

Konboi: OK, I thought that I should arrange the name with other headers. but I'll use http2 rule

launchable/commands/subset.py
```diff
@@ -198,7 +198,8 @@ def run(self):
                     path = "/intake/organizations/{}/workspaces/{}/subset".format(
                         org, workspace)
 
-                    client = LaunchableClient(token)
+                    client = LaunchableClient(
+                        token, test_runner=context.invoked_subcommand)
```
Konboi, line 202: Is there any other way to get a test-runner name?

ninjinkun (approved):
Looks good. I prefer to add a test to assert the header on this or another PR

Konboi:
Thank you for your review! I added a unit test and it passed so, I'll merge and ship it

PR #180, opened 2021-03-26
add doc for RSpec plugin

yoshiori:
SSIA

### DIFF
--- a/docs/test-runners/rspec.md
+++ b/docs/test-runners/rspec.md
@@ -1,0 +1,99 @@
+# RSpec
+
+## Getting started
+
+First, follow the steps in the [Getting started](../getting-started.md) guide to install the Launchable CLI, set your API key, and verify your connection.
+
+Then return to this page to complete the three steps of implementation:
+
+1. Recording builds
+2. Subsetting test execution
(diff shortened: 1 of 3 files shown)


docs/test-runners/rspec.md
```diff
@@ -0,0 +1,99 @@
...
+bundle exec rspec $(cat launchable-subset.txt)
+
+bundle exec rspec $(cat launchable-remainder.txt)
+```
+
+You can remove the second part after we've let you know that the model is sufficiently trained. Once you do this, make sure to continue running the full test suite at some stage. Run `launchable record build` and `launchable record tests` for those runs to continually train the model.
+
+## Recording test results
+
+Use [rspec_junit_formatter](https://github.com/sj26/rspec_junit_formatter) to output the test results to a file in RSpec. If you already have a CI server storing your test results it may already be installed.
```
yoshiori, line 85: @awilkes I think `minitest` also needs `minitest-ci`, but it was not written.  Do you think I should fix that one separately?

awilkes: Yes, sounds good.

PR #182, opened 2021-04-01
[ST-872] add flavor option for create test session

Konboi:
API starts to support flavor option for create test session, so cli supports it

### DIFF
--- a/launchable/commands/record/session.py
+++ b/launchable/commands/record/session.py
@@ -1,5 +1,6 @@
 import click
 import os
+import json
 
 from ...utils.http_client import LaunchableClient
 from ...utils.token import parse_token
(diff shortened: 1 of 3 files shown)


launchable/commands/record/session.py
```diff
@@ -38,12 +47,18 @@ def session(build_name: str, save_session_file: bool, print_session: bool = True
         "Content-Type": "application/json",
     }
 
-    client = LaunchableClient(token)
+    flavor_dict = {}
+    for f in flavor:
+        flavor_dict[f[0]] = f[1]
```
kohsuke, line 52: Extra credit: flag and reject duplicate keys since that'd be an user error.

kohsuke (approved):

launchable/commands/record/session.py
Konboi: yes indeed, so I'll add todo comment and merge this branch

PR #183, opened 2021-04-01
[ST-863] show warning message when request gets 404

Konboi:
I have made 2 changed in this PR

- show an error message when the request gets 404 error
- the commonality to setup session

## show an error message when the request gets 404 error

If we request to create a session before record a build, got 404 error. But unfortunately, the error message isn't easy to understand what is the cause of the error now. So, I changed to show the message that what should we do.
Also, sometimes the subset command and the record tests get a 404 error. 
But I only added to the create session command and the record tests command, because the subset command gets 404 when the subset model doesn't create yet. https://github.com/launchableinc/cli/pull/183/commits/2c49478f38bc8d92c951398a483e57499562b685

#### before 

```sh
$ LAUNCHABLE_TOKEN=<TOKEN> launchable record session  --build teeeeeeeeeeeest
404 Client Error:  for url: https://api.mercury.launchableinc.com/intake/organizations/<org>/workspaces/<workspace>/builds/teeeeeeeeeeeest/test_sessions
```

#### after 

```sh
$ LAUNCHABLE_TOKEN=<TOKEN> launchable record session  --build teeeeeeeeeeeest
Build teeeeeeeeeeeest was not found. Make sure to run `launchable record build --name teeeeeeeeeeeest` before
404 Client Error:  for url: https://api.mercury.launchableinc.com/intake/organizations/<org>/workspaces/<workspace>/builds/teeeeeeeeeeeest/test_sessions
```

note: the create test session commands will be called by the subset command and the record tests command

## the commonality to setup session

the subset command and the record tests command build the session_id used by session or build_name on head of these methods. 
these logics are similar, so I optimize it and added test code.

### DIFF
--- a/launchable/commands/helper.py
+++ b/launchable/commands/helper.py
@@ -1,0 +1,27 @@
+import click
+from typing import Optional
+from .record.session import session as session_command
+from ..utils.session import read_session
+
+
+def find_or_create_session(context, session: Optional[str], build_name: Optional[str], flavor=[]) -> Optional[str]:
+    if session and build_name:
+        raise click.UsageError(
+            'Only one of --build or --session should be specified')
+
+    if session is None and build_name is None:
+        raise click.UsageError(
+            'Either --build or --session has to be specified')
+
+    if session:
+        return session
+    elif build_name:
+        session_id = read_session(build_name)
+        if session_id:
+            return session_id
+        else:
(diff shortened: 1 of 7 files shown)


kohsuke (approved):
Love this attention to the details!

PR #188, opened 2021-04-08
add check wildcard characters are contained or not

Konboi:
before this change, if the test_path method receives test path strings which are contained wildcards, this method sets test_paths directly.

```sh
$ launchable subset --build 1 --target 20% minitest "test/**/*.rb"
test/**/*.rb

$ launchable subset --build 1 --target 20% minitest test/**/*.rb
test/application_system_test_case.rb
test/channels/application_cable/connection_test.rb
test/controllers/user_test.rb
test/models/admin/user_test.rb
test/models/open_class_user_test.rb
test/models/user_copy_test.rb
test/models/user_test.rb
test/test_h*ge.rb
test/test_helper.rb
```

so I fixed the test_path method expands wildcards 
```sh
$ launchable subset --build 1 --target 20% minitest "test/**/*.rb"
test/test_h*ge.rb
test/application_system_test_case.rb
test/test_helper.rb
test/models/open_class_user_test.rb
test/models/user_copy_test.rb
test/models/user_test.rb
test/models/admin/user_test.rb
test/controllers/user_test.rb
test/channels/application_cable/connection_test.rb
```

SHELLs usually expand wildcard, but when we combine some commands, SHELLs realize the wildcards as strings and don't expand, I think

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -118,6 +118,9 @@ def subset(context, target, session: Optional[str], base_path: Optional[str], bu
         def test_path(self, path: TestPathLike):
             if isinstance(path, str) and any(s in path for s in ('/**', '**/', '*.', "/*")):
                 for i in glob.iglob(path, recursive=True):
+                    if os.path.isdir(i):
+                        continue
+
                     self.test_path(i)
                 return
 
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -116,7 +116,7 @@ def subset(context, target, session: Optional[str], base_path: Optional[str], bu
             self._separator = s
 
         def test_path(self, path: TestPathLike):
-            if isinstance(path, str) and any(s in path for s in ('/**', '**/', '*.')):
+            if isinstance(path, str) and any(s in path for s in ('/**', '**/', '*.', "/*")):
                 for i in glob.iglob(path, recursive=True):
                     self.test_path(i)
                 return
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -116,6 +116,11 @@ def subset(context, target, session: Optional[str], base_path: Optional[str], bu
             self._separator = s
 
         def test_path(self, path: TestPathLike):
+            if isinstance(path, str) and any(s in path for s in ('/**', '**/', '*.')):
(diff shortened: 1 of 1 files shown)


launchable/commands/subset.py
```diff
@@ -116,6 +116,11 @@ def separator(self, s: str):
             self._separator = s
 
         def test_path(self, path: TestPathLike):
+            if any(s in path for s in ('/**', '**/', '*.')):
```
Konboi, line 119: Are there other cases?

launchable/commands/subset.py
```diff
@@ -116,6 +116,11 @@ def separator(self, s: str):
             self._separator = s
 
         def test_path(self, path: TestPathLike):
+            if isinstance(path, str) and any(s in path for s in ('/**', '**/', '*.')):
```
ninjinkun, line 119: Do we need to consider `*`, `/*`case?

Konboi: if add `*`, `test/test_h*ge.rb` match this if statement and happen infinity loop error.
so I'll add only `/*`.

(Ofcource, I know `test/test_h*ge.rb`is rare case)

Konboi: @ninjinkun I added https://github.com/launchableinc/cli/pull/188/commits/2e2b75762e12131e9dcf468d09d58fbcb7391fb9

ninjinkun (approved):
Looks good!

launchable/commands/subset.py
kohsuke: Help me understand why this is not `s in path for s in ('*', '?')` and instead there are restrictions about where the wildcard appears. This feels like a rough edge waiting to bite somebody. I can totally imagine somebody doing something like `foo/*/*_test.py` somewhere.

Why would this result in an infinite loop? And even if it does, shouldn't we resolve that by not doing recursion?

Konboi: thank you. your point is right.

before I added this change (https:/github.comlaunchableincclipull188commits521288834e1f0c26ee6fdd512b6fcd58027704b6), an infinite loop error had happened.
but now, we checked the path is a directory or not, so it doesn't happen.

I'll fix it.

PR #191, opened 2021-04-13
[ENG-61] change to flavor option format

Konboi:
before this change, the flavor option provides a `--flavor key value` format.

but the source option in the record build command provides a `--source repo=name` format.
IOW, the source option already supports a `key=value` style.

so I changed to keep the `key=value` format

### DIFF
--- a/launchable/commands/record/session.py
+++ b/launchable/commands/record/session.py
@@ -31,8 +31,6 @@ LAUNCHABLE_SESSION_DIR_KEY = 'LAUNCHABLE_SESSION_DIR'
     "--flavor",
     "flavor",
     help='flavors',
-    nargs=2,
-    type=(str, str),
     multiple=True,
 )
 def session(build_name: str, save_session_file: bool, print_session: bool = True, flavor=[]):
@@ -50,7 +48,18 @@ def session(build_name: str, save_session_file: bool, print_session: bool = True
     # TODO: check duplicate keys
     flavor_dict = {}
     for f in flavor:
-        flavor_dict[f[0]] = f[1]
+        if "=" not in f:
+            click.echo(
+                "Skip to set --flavor {}. Make sure to set `--flavor key=value`.".format(f), err=True)
+            continue
+
+        k, v = f.split('=')
+        if not k or not v:
+            click.echo(
+                "Skip to set --flavor {}. Make sure to set `--flavor key=value`.".format(f), err=True)
+            continue
+
+        flavor_dict[k] = v
 
     client = LaunchableClient(token)
     try:
--- a/launchable/commands/record/session.py
+++ b/launchable/commands/record/session.py
@@ -6,7 +6,7 @@ from ...utils.http_client import LaunchableClient
 from ...utils.token import parse_token
 from ...utils.env_keys import REPORT_ERROR_KEY
 from ...utils.session import write_session
-
+from ...utils.click import KeyValueType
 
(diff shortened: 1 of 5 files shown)


launchable/commands/record/session.py
```diff
@@ -50,7 +48,18 @@ def session(build_name: str, save_session_file: bool, print_session: bool = True
     # TODO: check duplicate keys
     flavor_dict = {}
     for f in flavor:
-        flavor_dict[f[0]] = f[1]
+        if "=" not in f:
```
ninjinkun, line 51: I think we can support both `key=val` and `key val` with `nargs=-1` option and [custom type](https://click.palletsprojects.com/en/7.x/parameters/#implementing-custom-types). Don't you think we need to support them?

Konboi: Oh, really? thank you for your advice.

I'll read it and try to support both

Konboi: @ninjinkun I tried it but couldn't. sorry,
because the `nargs` option only provides click.argument not supports click.option.

I tried to use a [click.argument](https://click.palletsprojects.com/en/7.x/arguments/) for the flavor option, but the click.argument can't set option name and some problems to implement it.

so unfortunately, I decided to keep this change

Konboi: but I just found it https://stackoverflow.com/questions/48391777/nargs-equivalent-for-options-in-click
so I tried it

Konboi:
nargs=-1 is only supported custom attribute so I can't use it to a custom type.

so I defined KeyValueType for the click option and use it.
https://github.com/launchableinc/cli/pull/191/files#diff-739a2fcee2511af52af764b3a3bf8dcc9f26ff0dfc754f4996c8867092797634R51

Konboi:
@ninjinkun I fixed so, could you review again?

launchable/utils/click.py
```diff
@@ -48,6 +48,52 @@ def convert(self, value: str, param, ctx):
...
 
+class KeyValueType(click.Option):
+    def __init__(self, *args, **kwargs):
+        super(KeyValueType, self).__init__(*args, **kwargs)
+        self._previous_parser_process = None
+        self._key_value_parser = None
+
+    def add_to_parser(self, parser, ctx):
+        def parser_process(value, state):
+            errorMessage = "Expected key-value like --option kye=value or --option key value. but got '{}'"
```
ninjinkun, line 59: This variable looks static. I think it should be a constant variable or a class variable.

ninjinkun:
Basically, it's awesome! I left one comment.

launchable/utils/click.py
Konboi: I fixed https://github.com/launchableinc/cli/pull/191/commits/5d45cfab687cfb94130a1465637a79885eb41f43. you mean it, right?

launchable/utils/click.py
```diff
@@ -48,6 +48,53 @@ def convert(self, value: str, param, ctx):
             value), param, ctx)
 
 
+class KeyValueType(click.Option):
+    errorMessage = "Expected key-value like --option kye=value or --option key value. but got '{}'"
```
ninjinkun, line 52: Please use snake case for variables

Konboi: 🙆‍♀️

ninjinkun (approved):
Approved to unblock to merge

PR #192, opened 2021-04-14
Run verify command on Windows

ninjinkun:
To run `launchable verify` on Windows, I modified the code to use `shutil.which`.

Here is the command result.

```shell
PS C:\Users\IEUser> launchable verify
Organization: ninjinkun2
Workspace: oreore
Platform: Windows-10-10.0.17763-SP0
Python version: 3.9.4
Java command: java
launchable version: 1.13.3.dev7+g360b92b
Your CLI configuration is successfully verified

### DIFF
--- a/launchable/utils/java.py
+++ b/launchable/utils/java.py
@@ -1,10 +1,8 @@
 import os
-import subprocess
-
+import shutil
 
 def get_java_command():
-    res = subprocess.run(["which", "java"], stdout=subprocess.DEVNULL)
-    if res.returncode == 0:
+    if shutil.which("java"):
         return "java"
 
     if os.access(os.path.expandvars("$JAVA_HOME/bin/java"), os.X_OK):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: launchable/utils/java.py [hunk 1]
```diff
@@ -1,13 +1,11 @@
 import os
-import subprocess
-
+import shutil
 
 def get_java_command():
-    res = subprocess.run(["which", "java"], stdout=subprocess.DEVNULL)
-    if res.returncode == 0:
+    if shutil.which("java"):
```
Konboi, line 5: 🆒

Konboi (approved):
LGTM!

Move Fast 🏃‍♂️ 

___

anyway, I think we should add Windows to ci 

refs: 
https://github.com/launchableinc/cli/blob/b91d6754d98199f42dd5cd100c1c5a3b37196676/.github/workflows/python-package.yml#L15

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

PR #194, opened 2021-04-15
Windows support

ninjinkun:
Enable to run `launchable subset` and `launchable record test` on Windows.
* Use [Win32_LogonSession](https://powershell.one/wmi/root/cimv2/win32_logonsession) as a session id
* Canonicalize path name in TestPath with `replace("\\", "/")`
* Add Windows runner on GitHub Action
* Fixed some tests

### DIFF
--- a/launchable/utils/session.py
+++ b/launchable/utils/session.py
@@ -75,7 +75,7 @@ def clean_session_files(days_ago: int = 0) -> None:
                 child.stat().st_mtime)
 
             if sys.platform == "win32":
-            # Windows sometimes misses to delete session files at Unit Test
+                # Windows sometimes misses to delete session files at Unit Test
                 microseconds = -10
             else:
                 microseconds = 0
--- a/launchable/utils/session.py
+++ b/launchable/utils/session.py
@@ -20,7 +20,7 @@ def _session_file_path(build_name: str) -> Path:
 
 def _get_session_id():
     if sys.platform == "win32":
-        import wmi
+        import wmi # type: ignore
         c = wmi.WMI()
         wql = "Associators of {{Win32_Process='{}'}} Where Resultclass = Win32_LogonSession Assocclass = Win32_SessionProcess".format(os.getpid())
         res = c.query(wql)
--- a/launchable/utils/session.py
+++ b/launchable/utils/session.py
@@ -73,7 +73,14 @@ def clean_session_files(days_ago: int = 0) -> None:
         for child in _session_file_dir().iterdir():
             file_created = datetime.datetime.fromtimestamp(
                 child.stat().st_mtime)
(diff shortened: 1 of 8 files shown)


launchable/utils/session.py
```diff
@@ -20,7 +19,14 @@ def _session_file_path(build_name: str) -> Path:
 
 
 def _get_session_id():
-    id = os.getsid(os.getpid())
+    if sys.platform == "win32":
+        import wmi # type: ignore
+        c = wmi.WMI()
+        wql = "Associators of {{Win32_Process='{}'}} Where Resultclass = Win32_LogonSession Assocclass = Win32_SessionProcess".format(os.getpid())
+        res = c.query(wql)
+        id = res[0].LogonId
```
ninjinkun, line 27: Whoops, this line makes test flaky. Sometimes it seems to return the same id.

ninjinkun: Sorry, I misunderstood. This line isn't bad.

launchable/utils/session.py
```diff
@@ -20,7 +19,14 @@ def _session_file_path(build_name: str) -> Path:
 
 
 def _get_session_id():
-    id = os.getsid(os.getpid())
+    if sys.platform == "win32":
```
kohsuke, line 22: I don't know the concept and the practical use of a logon session well enough to be able to say if this is good or bad. I assume you looked into this somewhat in order to get to this code. Is there any journal around that effort?

I'd think one key test is to see if two different shell invocations in GitHub Action would produce the same id under this scheme. If not, we have a problem.

launchable/utils/session.py
```diff
@@ -20,7 +19,14 @@ def _session_file_path(build_name: str) -> Path:
...
+        import wmi # type: ignore
+        c = wmi.WMI()
+        wql = "Associators of {{Win32_Process='{}'}} Where Resultclass = Win32_LogonSession Assocclass = Win32_SessionProcess".format(os.getpid())
+        res = c.query(wql)
+        id = res[0].LogonId
+    else:
+        id = os.getsid(os.getpid())
```
kohsuke, line 29: I realize now that I'm looking around this code that the type of `id` will be either `int` or `str` based on the code path. Where I came from, a statically typed language, I'd be inclined make them a consistent type, I'd think that's preferrable here too, unless you feel otherwise.

kohsuke (approved):

launchable/utils/session.py
ninjinkun: I haven't written any journal about it. I referred [a CPAN module](https://metacpan.org/pod/Win32::Monitoring::Session).

Testing with two different shell invocations is a good idea. I'm going to try it.

PR #199, opened 2021-04-23
Update the PyPI link to point to the launchable page

shuheiktgw:
Intuitively, I want the link to point to the launchable PyPI page instead of PyPI home page otherwise I feel the link is kind of useless.


awilkes (approved):
Heh, good catch!

PR #201, opened 2021-04-27
print warning message when output list is empty

Konboi:
- if the users run subset command in miss place, the cli show error message
- fix to trim base path when the user set base option in subset command

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -117,19 +117,26 @@ def subset(context, target, session: Optional[str], base_path: Optional[str], bu
             self._separator = s
 
         def test_path(self, path: TestPathLike):
+            def rel_base_path(path: str):
+                p = path
+                if base_path:
+                    p = normpath(relpath(path, start=base_path))
+
+                return p
+
             if isinstance(path, str) and any(s in path for s in ('*', "?")):
                 for i in glob.iglob(path, recursive=True):
                     if os.path.isfile(i):
-                        test_path = i
-                        if base_path:
-                            test_path = normpath(
-                                relpath(i, start=base_path))
-
-                        self.test_paths.append(self.to_test_path(test_path))
+                        self.test_paths.append(
+                            self.to_test_path(rel_base_path(i)))
                 return
 
             """register one test"""
-            self.test_paths.append(self.to_test_path(path))
+            p = path
+            if base_path and isinstance(p, str):
+                p = rel_base_path(p)
+
+            self.test_paths.append(self.to_test_path(p))
 
         def stdin(self):
             """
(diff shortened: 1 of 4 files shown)


launchable/commands/subset.py
```diff
@@ -117,14 +117,26 @@ def separator(self, s: str):
             self._separator = s
 
         def test_path(self, path: TestPathLike):
+            def rel_base_path(path: str):
```
Konboi, line 120: test_runners sometimes call test_path but the test_path method hasn't considered the base option.
so I fixed it.

Let me make sure, just in case. 
we should trim the base path when the user set base option, is this right?

launchable/commands/subset.py
```diff
@@ -239,6 +251,11 @@ def run(self):
                     click.echo(click.style(
                         "Warning: the service failed to subset. Falling back to running all tests", fg='yellow'), err=True)
 
+            if len(output) == 0:
```
Konboi, line 254: this is the main change in this PR

launchable/commands/subset.py
```diff
@@ -117,14 +117,26 @@ def separator(self, s: str):
             self._separator = s
 
         def test_path(self, path: TestPathLike):
+            def rel_base_path(path: str):
+                p = path
```
ninjinkun, line 121: You don't need to copy the string because it's an immutable value. You can simply write it like the following code.

```python3
return normpath(relpath(path, start=base_path)) if base_path else path
```

launchable/commands/subset.py
```diff
@@ -117,14 +117,26 @@ def separator(self, s: str):
...
                     if os.path.isfile(i):
-                        self.test_paths.append(self.to_test_path(i))
+                        self.test_paths.append(
+                            self.to_test_path(rel_base_path(i)))
                 return
 
             """register one test"""
-            self.test_paths.append(self.to_test_path(path))
+            p = path
+            if base_path and isinstance(p, str):
```
ninjinkun, line 136: Why did you check `isinstance(p, str)` here, but not in the `rel_base_path()` function? If you needed it in `rel_base_path()`, you could integrate it into the function.

ninjinkun:
Your key idea looks good. I left some minor comments.

launchable/commands/subset.py
Konboi: Sounds nice, I'll fix it.

Konboi:
@ninjinkun I fixed your feedback. could you review again?

launchable/commands/subset.py
```diff
@@ -117,14 +117,21 @@ def separator(self, s: str):
             self._separator = s
 
         def test_path(self, path: TestPathLike):
+            def rel_base_path(path):
+                if isinstance(path, str):
```
ninjinkun, line 121: Almost good. But it's still a little bit hard to understand because there are two conditions. Let's gather them to one if statement.
```python3
return normpath(relpath(path, start=base_path)) if isinstance(path, str) and base_path else path
```
or
```python3
if isinstance(path, str) and base_path:
  return normpath(relpath(path, start=base_path))
else:
  teturn path
```

ninjinkun (approved):
I left one comment. But it's a nit. Unblock this PR.

PR #202, opened 2021-04-28
Ant support

ninjinkun:
# Usage
## Subsetting
According to [this Ant tutorial](https://ant.apache.org/manual/tutorial-HelloWorldWithAnt.html), Ant (or old Java project?) put test files into an `src` directory instead of `src/test/java/...`. That is why I thought the plugin needs to support root path + glob like `src/**/*Test.java`.

```shell
launchable subset --build 123 ant 'src/**/*Test.java' > check-launchable.txt
ant junit
```

Users need to edit their `build.xml` to import the subset text file like below (complete `build.xml` is in https://github.com/launchableinc/rocket-car-ant).

```xml
   <target name="check-launchable">
      <available file="launchable-subset.txt" property="launchable"/>
    </target>

    <target name="junit" depends="jar,check-launchable">
        <mkdir dir="${report.dir}"/>
        <junit printsummary="yes">
            <classpath>
                <path refid="classpath"/>
                <path refid="application"/>
            </classpath>

            <formatter type="xml"/>

            <batchtest fork="yes" todir="${report.dir}">
              <fileset dir="${src.dir}" >
                <includesfile name="launchable-subset.txt" if="${launchable}" />
                <include name="**/*Test.java" unless="${launchable}" />
            </fileset>
            </batchtest>
        </junit>
    </target>
```

## Recording tests
```shell
launchable record tests --build 123 ant  build/junitreport/TESTS-TestSuites.xm
```

### DIFF
--- a/launchable/test_runners/ant.py
+++ b/launchable/test_runners/ant.py
@@ -1,0 +1,32 @@
+import os
+from typing import List
+import click
+from . import launchable
+from ..utils.file_name_pattern import jvm_test_pattern
+
+
+@click.argument('source_roots', required=True, nargs=-1)
+@launchable.subset
+def subset(client, source_roots: List[str]):
+    def file2test(f: str):
+        if jvm_test_pattern.match(f):
+            f = f[:f.rindex('.')]   # remove extension
+            # directory -> package name conversion
+            cls_name = f.replace(os.path.sep, '.')
+            return [{"type": "class", "name": cls_name}]
+        else:
+            return None
+    
+    for root in source_roots:
+        # split "src/**/*Test.java" to "src/" and "**/*Test.java"
+        glob = "**/*"
+        if '*' in root:
+          idx = root.find('*')
+          glob = root[idx:]
+          root = root[:idx]
+        client.scan(root.rstrip('/'), glob, file2test)
+    
+    client.run()
+
+
+record_tests = launchable.CommonRecordTestImpls(__name__).report_files()
--- a/launchable/test_runners/ant.py
+++ b/launchable/test_runners/ant.py
@@ -18,13 +18,7 @@ def subset(client, source_roots: List[str]):
             return None
     
     for root in source_roots:
-        # split "src/**/*Test.java" to "src/" and "**/*Test.java"
-        glob = "**/*"
-        if '*' in root:
-          idx = root.find('*')
-          glob = root[idx:]
-          root = root[:idx]
-        client.scan(root.rstrip('/'), glob, file2test)
+        client.scan(root.rstrip('/'), "**/*Test.java", file2test)
     
     client.run()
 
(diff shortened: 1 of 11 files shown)


launchable/test_runners/ant.py
```diff
@@ -0,0 +1,32 @@
...
+        else:
+            return None
+    
+    for root in source_roots:
+        # split "src/**/*Test.java" to "src/" and "**/*Test.java"
+        glob = "**/*"
+        if '*' in root:
+          idx = root.find('*')
+          glob = root[idx:]
+          root = root[:idx]
```
kohsuke, line 26: I think this is actually problematic.

The easiest case to think of would be what if the input is `src/test/java/com/launchableinc/integration_tests/**/*Test.java`
It fails to produce the correct fully qualified class names.

This is why a fileset has the "dir" part and includes/excludes part. A file set isn't just a collection of files. It's a base directory, plus a collection of files within it.

One way to solve this is to restrict `source_roots` to be base directories. Then define `--includes PATTERN` and `--excludes PATTERN`, somewhat like how `tar` does it, which has the same structure of the modeling as we do. That is, `launchable subset ant src/test/java --includes 'com/launchableinc/integration_test/**/*Test.java'`

I recognize in order to support multiple source roots, this would have to make `--includes` and `--excludes` behave positionally, which pushes the envelope of how we use Click. This also requires expanding `Optimize` class with this new way of specifying things. This will be worth making it reusable as it won't be unique to Ant, which makes me wonder how we did other Java test runners, which generally has the same problem.

If that's a little too much to take on for Ant, which is arguably a minority test runner in 2021, maybe our starting point is to limit `source_roots`  to be directories and we hard code the default GLOB mask. That gets the ball moving until this limited mode of the support becomes inadequate.

I'd love to hear if you see other ways to solve this.

kohsuke (changes requested):
Design problem that has future compatibility implications.

launchable/test_runners/ant.py
ninjinkun: Got it. In my understanding, recent Java projects use `src/test/java/` file path but old projects use `src/` directly, am I right? If so, we need to support both in Java projects. 

Unfortunately, I don't much time today. I fix it with the easiest way you pointed out `**/*Test.java`. But I'm going to create a ticket for supporting these file paths.

ninjinkun:
@kohsuke Any thought about this?

launchable/test_runners/ant.py
kohsuke: When Ant was in active use, no such directory name convention was established. People used all sorts of random directory names to house the source roots, so I certainly wouldn't treat `src/test/java` or `src` any differently.

(BTW, `src/test/java` is the convention established by Maven and I think Gradle followed that somewhat)

kohsuke (approved):
I think this is a great starting point. Thanks for accommodating my concern.

PR #205, opened 2021-05-07
[ENG-111] decompress data when request payload is compressed

Konboi:
e.g) subset request

before

```
AUDIT:launchable:send request method:post path:/intake/organizations/org/workspaces/ws/subset headers:{'Content-Type': 'application/json', 'Content-Encoding': 'gzip'} args:{'data': b'\x1f\x8b\x08\x00\x9e\xdb\x94`\x02\xff\xabV*I-.\tH,\xc9(V\xb2R\x88\x8e\xd5QP*N-.\xce\xcc\xcf\x03r\xab\x952S\x80\x94\x92\xa1\xa5\x85R-P\xa6$\xb1(=\xb5\x04(b\xa0gX\x0b\x00\xb5\x04\x9c\xb2:\x00\x00\x00', 'timeout': (5, 180)}
```

after 

```
AUDIT:launchable:send request method:post path:/intake/organizations/org/workspaces/ws/subset headers:{'Content-Type': 'application/json', 'Content-Encoding': 'gzip'} args:{'data': '{"testPaths": [[{"type": "class", "name": "com.example.package.App2Test"}], [{"type": "class", "name": "com.example.package.AppTest"}]], "session": {"id": "198"}, "target": 0.3}', 'timeout': (5, 180)}
```

### DIFF
--- a/launchable/utils/http_client.py
+++ b/launchable/utils/http_client.py
@@ -1,6 +1,8 @@
 import requests
 import os
 import platform
+import gzip
+import copy
 from launchable.version import __version__
 from .logger import Logger
 
@@ -28,9 +30,16 @@ class LaunchableClient:
 
         logger = Logger()
         try:
+            log_kwargs = copy.copy(kwargs)
+            if any("gzip" in h for h in headers.values()):
+                log_kwargs["data"] = gzip.decompress(
+                    log_kwargs["data"]).decode("utf-8")
+
             logger.audit(
-                "send request method:{} path:{} headers:{} args:{}".format(method, path, headers, kwargs))
-            response = self.session.request(method, url, headers={**headers, **self._headers()}, **kwargs)
+                "send request method:{} path:{} headers:{} args:{}".format(method, path, headers, log_kwargs))
+
+            response = self.session.request(
+                method, url, headers={**headers, **self._headers()}, **kwargs)
             logger.debug(
                 "received response status:{} message:{} headers:{}".format(
                     response.status_code, response.reason, response.headers)
--- a/launchable/utils/http_client.py
+++ b/launchable/utils/http_client.py
@@ -1,13 +1,8 @@
 import requests
 import os
 import platform
-import gzip
-import copy
-import types
-import itertools
-import logging
 from launchable.version import __version__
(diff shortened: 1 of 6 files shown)


launchable/utils/http_client.py
```diff
@@ -28,9 +33,25 @@ def request(self, method, path, **kwargs):
...
         try:
-            logger.audit(
-                "send request method:{} path:{} headers:{} args:{}".format(method, path, headers, kwargs))
-            response = self.session.request(method, url, headers={**headers, **self._headers()}, **kwargs)
+            if LOG_LEVEL_AUDIT >= logging.root.level:
+                log_kwargs = copy.copy(kwargs)
+                if any("gzip" in h for h in headers.values()) and "data" in log_kwargs:
+                    data = log_kwargs.pop("data")
+                    if isinstance(data, types.GeneratorType):
+                        generator, _generator = itertools.tee(data)
```
Konboi, line 41: https://docs.python.org/3/library/itertools.html#itertools.tee

> Once tee() has made a split, the original iterable should not be used anywhere else; otherwise, the iterable could get advanced without the tee objects being informed.

kohsuke (commented):
I think we got the layering wrong.

Auditing should be above content encoding, layer wise. So if we are to make `LaunchableClient` responsible for auditing, which makes sense, then it also needs to take on the responsibility of content encoding. That way, it can do things in the right order.

See for example how `record/tests.py` does the debug printing and sending in the right order, which keeps the code much more straightforward.

`itertools.tee` is also problematic becaues it is clearly buffering the content. So it eliminates the point of using the generator. Better to do it like the abovementioned `record/tests.py` in which data is consumed in a streaming manner.

Konboi:
Thank you for your comment.

I was also wondering which layer I should use to show the log.
I'll change the place to the print log.

if the LaunchableClient supports some encoding formats, we don't have to care of the print log for each method. unfortunately, it will be big changes. 
I want to avoid it and I'll set log each method.

Konboi:
@kohsuke I fixed so could you review again?

kohsuke (approved):
I see that we now restored the proper layering in the other direction, by moving the auditing responsibility out of LaunchableClient

PR #207, opened 2021-05-07
Reorder instructions to make sure tests get recorded first

awilkes:
...Then come back to subsetting.


awilkes:
This is just a first pass. I will make the flow better later.

PR #208, opened 2021-05-10
[ENG-113] use os.path.relpath

Konboi:
to avoid operate path as an array

I checked used by https://github.com/launchableinc/examples/tree/master/rspec

## before 

```
$ launchable subset --target 50% --build mm rspec spec/
spec/odels/article_spec.rb
spec/odels/comment_spec.rb
spec/equests/articles_spec.rb
spec/equests/comments_spec.rb
```

## after 

```
$ launchable subset --target 50% --build mm rspec spec/
spec/models/article_spec.rb
spec/models/comment_spec.rb
spec/requests/articles_spec.rb
spec/requests/comments_spec.rb
```


kohsuke (approved):
:+1: looks more kosher indeed

PR #210, opened 2021-05-13
Add max-days option

shuheiktgw:
Added max-days option so that users can collect commits older than 30 days.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,7 +17,7 @@ packages = find:
 setup_requires =
     setuptools_scm
 install_requires =
-    click>=7.0
+    click~=7.0
     requests
     junitparser>=2.0.0
     setuptools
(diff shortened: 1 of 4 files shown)


Pipfile
```diff
@@ -15,6 +15,7 @@ mypy = "*"
 
 [packages]
 launchable = {editable = true,path = "."}
+click = "~=7.1"
```
shuheiktgw, line 18: As we discussed in Slack, Click v8.0.0 was released yesterday and some tests does not work with the version

setup.cfg
```diff
@@ -17,7 +17,7 @@ packages = find:
 setup_requires =
     setuptools_scm
 install_requires =
-    click>=7.0
+    click~=7.0
```
shuheiktgw, line 20: As we discussed in Slack, Click v8.0.0 was released yesterday and some tests does not work with the version

ninjinkun (approved):

shuheiktgw:
Thank you!

PR #211, opened 2021-05-13
Option to just add class names without '--tests' prefix

kohsuke:
Intended for more advanced scripted integration with Gradle

### DIFF
--- a/launchable/test_runners/gradle.py
+++ b/launchable/test_runners/gradle.py
@@ -5,9 +5,15 @@ import click
 from . import launchable
 from ..utils.file_name_pattern import jvm_test_pattern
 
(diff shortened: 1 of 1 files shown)


launchable/test_runners/gradle.py
```diff
@@ -5,9 +5,15 @@
 from . import launchable
 from ..utils.file_name_pattern import jvm_test_pattern
 
+@click.option('--bare',
```
awilkes, line 8: Maybe `--list`? No strong opinion.

kohsuke:
OK, I'm gonna go with the `--bare` option on the idea that `--tests a --tests b ...` is also listing something.

PR #214, opened 2021-05-19
[ENG-115] to check report file timestamp

Konboi:
In some environments, the report files have been staying. then the record tests command sends old test results to Launchable.
So we changed to check that timestamp of the report file is newer than the time of record build start at.


ninjinkun (approved):
LGTM

Konboi:
I'll check a few things and then merge.

PR #215, opened 2021-05-24
Add build-event-json option to the bazel plugin

Konboi:
the bazel doesn't update the timestamp of the test report files if there is a cache of the test.
However, Launchable needs only reports that were tested so we want to ignore old reports.

so I added `--build-event-json` option to use build event json file of bazel (refs: [build event protocol](https://docs.bazel.build/versions/4.0.0/build-event-protocol.html))


tests/data/bazel/build_event.json
```diff
@@ -0,0 +1,36 @@
+{"id":{"started":{}},"children":[{"progress":{}},{"unstructuredCommandLine":{}},{"structuredCommandLine":{"commandLineLabel":"original"}},{"structuredCommandLine":{"commandLineLabel":"canonical"}},{"structuredCommandLine":{"commandLineLabel":"tool"}},{"buildMetadata":{}},{"optionsParsed":{}},{"workspaceStatus":{}},{"pattern":{"pattern":["src/test/java/com/ninjinkun:mylib_test1","src/test/java/com/ninjinkun:mylib_test2","src/test/java/com/ninjinkun:mylib_test5"]}},{"buildFinished":{}}],"started":{"uuid":"97af8ec6-c78b-441d-9b4f-f673b6d587e4","startTimeMillis":"1621841949774","buildToolVersion":"4.0.0-homebrew","optionsDescription":"--build_event_json_file\u003d./build_event_with_subset.json","command":"test","workingDirectory":"/Users/yabuki-ryosuke/src/github.com/launchableinc/rocket-car-bazel/java","workspaceDirectory":"/Users/yabuki-ryosuke/src/github.com/launchableinc/rocket-car-bazel/java","serverPid":"14147"}}
```
Konboi, line 1: this is generated by below
```
$ /bin/cat subset.txt
src/test/java/com/ninjinkun:mylib_test1
src/test/java/com/ninjinkun:mylib_test2
src/test/java/com/ninjinkun:mylib_test5
$ bazel test  --build_event_json_file=./build_event.json $(/bin/cat subset.txt)
```

launchable/test_runners/bazel.py
```diff
@@ -56,6 +58,35 @@ def f(case: TestCase, suite: TestSuite, report_file: str) -> TestPath:
...
+    try:
+        f = open(path)
+    except Exception as e:
+        click.echo(click.style(
+            "Can not open build event json file {}".format(path), "yellow"), err=True)
+        return build_list
+
+    for line in f:
+        d = json.loads(line)
+        if "id" in d:
```
Konboi, line 84: there is a full path of the report file in a testResult field.
Although I wonder that should I use it, I decided to use the label field.
because it is difficult to prepare test data

```json
{
  "id": {
    "testResult": {
      "attempt": 1,
      "configuration": {
        "id": "86e57527e1c8bb10edc853e734cb858f8159d8f3e0a4df9ceb16f80aad784b93"
      },
      "label": "//src/test/java/com/ninjinkun:mylib_test6",
      "run": 1,
      "shard": 1
    }
  },
  "testResult": {
    "executionInfo": {
      "strategy": "darwin-sandbox"
    },
    "status": "PASSED",
    "testActionOutput": [
      {
        "name": "test.log",
        "uri": "file:///private/var/tmp/_bazel_yabuki-ryosuke/3f9f29672c16363f9fe4cad29f21ed46/execroot/__main__/bazel-out/darwin-fastbuild/testlogs/src/test/java/com/ninjinkun/mylib_test6/test.log"
      },
      {
        "name": "test.xml",
        "uri": "file:///private/var/tmp/_bazel_yabuki-ryosuke/3f9f29672c16363f9fe4cad29f21ed46/execroot/__main__/bazel-out/darwin-fastbuild/testlogs/src/test/java/com/ninjinkun/mylib_test6/test.xml"
      }
    ],
    "testAttemptDurationMillis": "2577",
    "testAttemptStartMillisEpoch": "1621924928775"
  }
}
```
```

File: launchable/test_runners/bazel.py [hunk 3]
```diff
@@ -28,8 +29,9 @@ def subset(client):
 
 
 @click.argument('workspace', required=True)
+@click.option('--build-event-json', 'build_event_json', help="set file path generated by --build_event_json_file", type=str, required=False)
```
kohsuke, line 32: `type=click.Path(exists=True)` would ensure this is a file that exists.

launchable/test_runners/bazel.py
```diff
@@ -56,6 +58,35 @@ def f(case: TestCase, suite: TestSuite, report_file: str) -> TestPath:
...
+
+    for line in f:
+        d = json.loads(line)
+        if "id" in d:
+            if "testResult" in d["id"]:
+                if "label" in d["id"]["testResult"]:
+                    label = d["id"]["testResult"]["label"]
+                    # replace //foo/bar:zot to /foo/bar/zot
+                    label = label.lstrip("/").replace(":", "/")
+                    build_list.append(label)
```
kohsuke, line 90: I was gonna say using a generator would be more Pythonesque but since the whole JSON is parsed eagerly I guess there's no point.

(I do wonder how big this JSON file gets and if we should be parsing this in streaming but we can take that on in the future)

launchable/test_runners/bazel.py
```diff
@@ -56,6 +58,35 @@ def f(case: TestCase, suite: TestSuite, report_file: str) -> TestPath:
 
     client.path_builder = f
     client.check_timestamp = False
-    client.scan(base, '**/test.xml')
+
+    if build_event_json:
+        lists = parse_build_event_json(build_event_json)
+        for l in lists:
+            path = Path(base).joinpath(l).resolve()
+            client.scan(path, "**/test.xml")
```
kohsuke, line 66: Given that `path` here represents one Bazel package it seems odd to me that we are scanning `**/test.xml`.

Shouldn't it be just `client.report(Path(base).joinpath(l,'test.xml'))` ?

For that matter, for my own education, what is the point of `resolve()` call?

kohsuke (approved):
It's great that this change came in this quickly. Left a few comments about some loose ends.

launchable/test_runners/bazel.py
Konboi: TIL, cool

launchable/test_runners/bazel.py
Konboi: thank you for your advice, I changed to use Generator

launchable/test_runners/bazel.py
Konboi: > Shouldn't it be just client.report(Path(base).joinpath(l,'test.xml')) ?

yes right, so I fixed it. thank you

> For that matter, for my own education, what is the point of resolve() call?
In my understanding, `resolve()` resolves symlinks and eliminate “..” components.

PR #216, opened 2021-05-25
[ENG-115] add error handling to parse the time string

Konboi:
the CLI should continue to report the results if the CLI can't parse the time string.
so I add an error handling when parsing the time string.

### DIFF
--- a/launchable/commands/record/tests.py
+++ b/launchable/commands/record/tests.py
@@ -23,6 +23,7 @@ from http import HTTPStatus
 from ...utils.click import KeyValueType
 from ...utils.logger import Logger, AUDIT_LOG_FORMAT
 import datetime
+from dateutil.parser import parse
 
 
 @click.group()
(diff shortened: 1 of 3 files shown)


launchable/commands/record/tests.py
```diff
@@ -289,5 +289,14 @@ def get_record_start_at(token: str, org: str, workspace: str, build_name: Option
...
         return datetime.datetime.now()
 
+    return parse_launchable_timeformat(res.json()["createdAt"])
+
+
+def parse_launchable_timeformat(t: str) -> datetime.datetime:
     # e.g) "2021-04-01T09:35:47.934+00:00"
-    return datetime.datetime.strptime(res.json()["createdAt"], "%Y-%m-%dT%H:%M:%S.%f%z")
+    try:
+        return datetime.datetime.strptime(t, "%Y-%m-%dT%H:%M:%S.%f%z")
```
ninjinkun, line 298: I found the Python official document that recommends `python-dateutil` module to parsing ISO8601 https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat . Of course, there is a trade-off between increasing dependency and reusability. You can decide it make sense or not.

Konboi: Thank you for your advice! I'll add `python-dateutil` and use it

ninjinkun (approved):
LGTM

PR #217, opened 2021-05-25
NUnit support + generic XML parsing support

kohsuke:
My attempt to tackle NUnit turned into a little bigger change, so I'm requesting code review from both of you.

The change mainly consists of two parts:

* Ability to parse arbitrary XML file format other than JUnit, and create `CaseEvent`s directly from them.
* NUnit support built on top of it.

`tests.py` and its `junitxml_parse_func` was retrofitted to sit on top of this new abstraction.

If this change is accepted, I'll go back and rework some of the existing test runners to use this mechanism. I believe it'll result in more straightforward code.


tests/test_runners/test_nunit.py
```diff
@@ -0,0 +1,37 @@
+from pathlib import Path
+import responses  # type: ignore
+import json
+import gzip
+from tests.cli_test_case import CliTestCase
+
+
+class NUnitTest(CliTestCase):
```
Konboi, line 8: I'm not familiar with NUnit. So could you add comments how to output these XML files or add README.md?

launchable/test_runners/nunit.py
```diff
@@ -0,0 +1,77 @@
...
+from launchable.testpath import TestPath
+
+# common code between 'subset' & 'record tests' to build up test path from nested <test-suite>s
+def build_path(e: Element):
+    pp = [] # type: TestPath
+    if e.parent:
+        pp = e.parent.tags.get('path') or []    # type: ignore
+    if e.name == "test-suite":
+        # <test-suite>s form a nested tree structure so capture those in path
+        e.tags['path'] = pp + [{'type': e.attrs['type'], 'name': e.attrs['name']}]
```
Konboi, line 18: TIL:  add elements to array use by `+` operator in Python

Konboi (approved):
LGTM 

___

I commented on one request.

ninjinkun (approved):
Looks good!

launchable/test_runners/nunit.py
ninjinkun: Because Python allows operator overloading. Especially Numpy uses this feature a lot.

PR #221, opened 2021-05-25
Set required python version to the minimum version we support

kohsuke:
... to help catch problems earlier.


kohsuke:
Given how this PR is going, I'm starting to get the sense that we have never actually tested with different Python versions before

.github/workflows/python-package.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install pipenv
-        pipenv install --dev
+        pipenv install --dev --python ${{ matrix.python-version }}
```
Konboi, line 31: TIL: https://pipenv.pypa.io/en/latest/advanced/#pipenv.environments.PIPENV_DEFAULT_PYTHON_VERSION

Konboi (approved):
LGTM

Nice work!!

ninjinkun:
Thanks!

PR #222, opened 2021-05-26
Remove `from __future__ import annotations`. because it is available starting is 3.7

yoshiori:


launchable/utils/sax.py
```diff
@@ -1,11 +1,12 @@
-from __future__ import annotations
 from xml.sax import *
 from xml.sax.handler import ContentHandler
 from typing import List, Dict, Callable
 import sys
 import re
 import click
 
+class Element:
+    pass
```
yoshiori, line 9: Self-reference resolution

PR #226, opened 2021-05-31
Add ADB plugin

ninjinkun:


tests/test_runners/test_adb.py
```diff
@@ -0,0 +1,63 @@
+from pathlib import Path
+import responses  # type: ignore
+import json
+import gzip
+import itertools
+from launchable.utils.session import read_session
+from tests.cli_test_case import CliTestCase
+
+
+class BazelTest(CliTestCase):
```
Konboi, line 10: ADBTest is correct?

Konboi (approved):
LGTM

___

But I commented a little bit

tests/test_runners/test_adb.py
ninjinkun: You are right!

PR #229, opened 2021-06-02
Up-to-date check improvement & bug fix

kohsuke:
- If for some reasons we fail to obtain a valid timestamp, we should err
  on the side of disabling up-to-date checks, as opposed to not collecting
  anything, which is what `now()` was doing.

- Added additional debug & warning messages that situations like this
  will be flagged more proactively, or else the user will never notice
  what's going on.

See: https://launchableinc.atlassian.net/l/c/AmddErcw


Konboi (approved):
LGTM

Konboi:
Thank you for fixing

PR #230, opened 2021-06-03
NUnit parsing error fix

kohsuke:
A customer reported that their NUnit report file didn't parse. Upon
inspection, theirs only had the start-time attribute at the root
element.

I suspect NUnit2 produces that kind of output but not entirely sure the
origin of it. In any case, the robustness feels worthwhile


Konboi (approved):
LGTM

PR #233, opened 2021-06-04
Change the default log level to warning

shuheiktgw:
The current default logging level is too high and even warnings and errors do not show up to stdout by default. I believe it is a common practice in CLI that shows logs the level above warning by default. For example, [the Python logging library](https://docs.python.org/ja/3/howto/logging.html) sets the default log level to the warning. 

Also, the `--debug` option contradicts the `--log-level` option and I believe a user should control it by `--log-level` option.

### DIFF
--- a/launchable/utils/http_client.py
+++ b/launchable/utils/http_client.py
@@ -31,7 +31,7 @@ class LaunchableClient:
 
         headers = self._headers(compress)
 
-        Logger().audit(AUDIT_LOG_FORMAT.format("post", url, headers, payload))
+        Logger().audit(AUDIT_LOG_FORMAT.format(method, url, headers, payload))
 
         data = _build_data(payload, compress)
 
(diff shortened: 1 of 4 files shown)


launchable/utils/http_client.py
```diff
@@ -31,7 +31,7 @@ def request(self, method, sub_path, payload=None, timeout=(5, 60), compress=Fals
 
         headers = self._headers(compress)
 
-        Logger().audit(AUDIT_LOG_FORMAT.format("post", url, headers, payload))
```
shuheiktgw, line 34: The method should be `method`, not `"post"`. This is a bug I incorporated in my previous PR...

Konboi (approved):
LGTM

But we need to create a guideline that how we separate the methods(click.echo and Logger) we use depending on when we logging.

shuheiktgw:
Thank you! I'll prepare documentation on it later

PR #234, opened 2021-06-07
Promote reuse of test runner support code

kohsuke:
What we discussed today -- promote more code reuse between different test runners


yoshiori (approved):
Thank you!

PR #235, opened 2021-06-08
Set GitHub-Pr-Head-Sha header only when it exists

shuheiktgw:
I've changed the authentication_headers method and set the `GitHub-Pr-Head-Sha` header only when it exists in the environment. Otherwise, the method sets an empty string to the `GitHub-Pr-Head-Sha` header and that could be confusing to the API.


Konboi:
Please merge after fixing test :pray:

shuheiktgw:
Hmm, let's see...

PR #237, opened 2021-06-08
Scope a session to a Jenkins build if we are running in one.

kohsuke:
We just found out that Jenkins pipeline assigns a new Unix session for every process it launches.

### DIFF
--- a/launchable/utils/session.py
+++ b/launchable/utils/session.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from typing import Optional
 import datetime
 import hashlib
+from .logger import Logger
 
 SESSION_DIR_KEY = 'LAUNCHABLE_SESSION_DIR'
 DEFAULT_SESSION_DIR = '~/.config/launchable/sessions/'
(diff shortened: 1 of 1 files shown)


launchable/utils/session.py
```diff
@@ -18,6 +19,25 @@ def _session_file_path(build_name: str) -> Path:
 
 
 def _get_session_id() -> str:
+    # CircleCI changes unix session id each steps, so set non change variable
+    # https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
+    if os.environ.get("CIRCLECI") is not None:
+        id = os.environ.get("CIRCLE_WORKFLOW_ID")
+        if id is not None:
+             return id
+        Logger().warning("CIRCLECI environment variable is set but not CIRCLE_WORKFLOW_ID")
```
kohsuke, line 28: If this branch is taken there's no need to do WMI nor Unix session, so better to have this at the top rather than the bottom

kohsuke, line 28: If CIRCLECI is set but CIRCLE_WORKFLOW_ID is not set, that's an unanticipated situation so we'd be better off flagging that instead of silently returning "", which leads to undesirable behaviours.

PR #245, opened 2021-06-21
Improved the error diagnostics

kohsuke:
From a recent interaction with an user who struggled through getting the proxy setting right came this change.

- I'm generally a fun of more information when things go wrong. When the verify command fails, the user is clearly hitting some blocking problem and so added details can save a lot of frustrating time. So when the verify command fails, report a full stack trace.
- Report the proxy that we are using, to help people verify that their proxy setting is correctly recognized by us.


shuheiktgw (approved):
Thank you!

PR #250, opened 2021-06-24
[ENG-135] Add support for PyTest

yoshiori:
- pytest uses `test_*.py` or `*_test.py as` the test file.
- <del>longer included in the Test Report. I think this is a trap, so I made the error message thicker.</del>
  - <del>https://docs.pytest.org/en/latest/deprecations.html#junit-family-default-value-change-to-xunit2</del>
- Added support for not having to change the formatter.

## Task
- [x] Write test.
- [x] Write Document.


awilkes:
I don't think we can ask customers to change their report format, since it could break their existing toolchain. (Plus, we're asking them to use an _older_ version, which is not a good look.)

yoshiori:
@awilkes Thanks for the input! 
I thought it was good that the only difference between xunit1 and xunit2 is the file attribute and line attribute.


But your opinion is correct. I wouldn't want to be told to use legacy either.
I'll fix it.

docs/languages/python.md
```diff
@@ -0,0 +1,8 @@
+# Python
```
yoshiori, line 1: @awilkes Python is now supported by two languages, so I created a language page.

awilkes: Awesome!

docs/subset-your-tests.md
```diff
@@ -0,0 +1,19 @@
+# Subset your test runs
```
awilkes, line 1: I think you might need to merge in `main`, because this file has been renamed to `subsetting-your-tests.md` since you created your branch.

yoshiori: Nice catch! Thank you!

tests/test_runners/test_pytest.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import Path
+import responses  # type: ignore
+import json
+import gzip
+import os
+from unittest import mock
+from tests.cli_test_case import CliTestCase
+
+
+class PytestTest(CliTestCase):
```
ninjinkun, line 10: Could you add a subsetting test?

launchable/test_runners/pytest.py
```diff
@@ -0,0 +1,27 @@
...
+          cls_name = f.replace(os.path.sep, '.')
+          client.test_path([{"type": "class", "name": cls_name}])
+
+    for root in source_roots:
+      for b in glob.iglob(root):
+        if isdir(b):
+          for t in glob.iglob(join(b, '**/*.py'), recursive=True):
+            add(t)
+        else:
+          add(b)
```
ninjinkun, line 23: Please add a new line below

tests/test_runners/test_pytest.py
yoshiori: Added.

launchable/test_runners/pytest.py
yoshiori: Fixed!

ninjinkun (approved):
LGTM

PR #252, opened 2021-06-29
Add no-submodules option to the record build command

shuheiktgw:
One of our customers had trouble running the git submodule command due to their environmental problem. To circumvent the problem, I added a new `--no-submodules` option to the record build command to disable information collection from Git Submodules (and skip running the git submodule command internally).


ninjinkun (approved):

shuheiktgw:
Thank you!

PR #254, opened 2021-06-29
Integrate Launchable service into this repo

ninjinkun:
I configured Launhchabe service into this repo as "dogfooding." I referred to [the OSS integration guide](https://docs.launchableinc.com/resources/integration-for-open-source).

* Replace default separator to a white space from a newline
  * This is because the newline separator failed runs on Windows. Bash on Windows expands command substitution results including newlines in a strange way. See [this result](https://github.com/launchableinc/cli/runs/2942077915#step:9:68).
  * But I'm not confident that it doesn't affect other customers. @kohsuke Do you think is this okay? I forget why we chose newline as a separator.
* Use `shell: bash` option to run the script on Windows

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -85,7 +85,7 @@ def subset(context, target, session: Optional[str], base_path: Optional[str], bu
             self.test_paths = []
             # TODO: robustness improvement.
             self._formatter = Optimize.default_formatter
-            self._separator = " "
+            self._separator = "\n"
 
         @staticmethod
         def default_formatter(x: TestPath):
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -85,7 +85,7 @@ def subset(context, target, session: Optional[str], base_path: Optional[str], bu
             self.test_paths = []
             # TODO: robustness improvement.
             self._formatter = Optimize.default_formatter
-            self._separator = "\n"
+            self._separator = " "
 
         @staticmethod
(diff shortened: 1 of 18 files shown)


launchable/commands/subset.py
```diff
@@ -85,7 +85,7 @@ def __init__(self):
             self.test_paths = []
             # TODO: robustness improvement.
             self._formatter = Optimize.default_formatter
-            self._separator = "\n"
+            self._separator = " "
```
kohsuke, line 88: Is this change safe? At a casual glance, this looks like it's going to impact every test runner support

kohsuke (approved):

kohsuke:
Looking forward to seeing this experiece summarized & captured so that we can improve the onboarding experience

shuheiktgw (approved):
👍

launchable/commands/subset.py
ninjinkun: Yeah, this is what I worried about. So, I'm going to modify it to limit to the file plugin instead of `subset.py`. Then, If the file plugin is okay at all customers environments, I will change default operator to white space, like a staged rollout.

ninjinkun:
On the Windows runner, Launchable CLI outputs `\r\n` as a subset separator by Python’s auto-convert feature. But the bash runner can’t handle `\r\n`. Finally, I converted `\r\n` to `\n` with `tr '\r\n' '\n'` on the shell.

PR #257, opened 2021-07-01
Retry when a request failed

shuheiktgw:
The same with https://github.com/launchableinc/nose-launchable/pull/44 and change it to retry requests when it fails.

### DIFF
--- a/tests/test_runners/test_gradle.py
+++ b/tests/test_runners/test_gradle.py
@@ -26,7 +26,7 @@ class GradleTest(CliTestCase):
         # TODO: we need to assert on the request payload to make sure it found test list all right
         self.assertEqual(result.exit_code, 0)
         output = '--tests com.launchableinc.rocket_car_gradle.App2Test --tests com.launchableinc.rocket_car_gradle.AppTest --tests com.launchableinc.rocket_car_gradle.sub.App3Test --tests com.launchableinc.rocket_car_gradle.utils.UtilsTest'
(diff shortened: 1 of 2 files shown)


tests/test_runners/test_gradle.py
```diff
@@ -26,7 +26,7 @@ def test_subset_without_session(self):
         # TODO: we need to assert on the request payload to make sure it found test list all right
         self.assertEqual(result.exit_code, 0)
         output = '--tests com.launchableinc.rocket_car_gradle.App2Test --tests com.launchableinc.rocket_car_gradle.AppTest --tests com.launchableinc.rocket_car_gradle.sub.App3Test --tests com.launchableinc.rocket_car_gradle.utils.UtilsTest'
-        self.assertEqual(result.output.rstrip('\n'), output)
+        self.assertIn(output, result.output.rstrip('\n'))
```
shuheiktgw, line 29: `method_whitelist=` outputs a warning message so I changed the assertion from equal to in.

Konboi (approved):

shuheiktgw:
Thank you!

awilkes:
What is the context driving this change, @shuheiktgw?

shuheiktgw:
First, it is a pretty common practice to retry requests a few times when it fails. Second, our API is still not fully reliable and returns 5xx (even though it becomes pretty rare recently) so I want our clients to retry requests to increase reliability!

awilkes:
Great, makes sense 👍 Thanks for driving this change

PR #258, opened 2021-07-02
split subset

Konboi:
## Context 

Launchable will support split subsetting so I implemented the cli to support it.

## Useage

```
# main
$ launchable subset --split --target 60%  --session <SESSION> bazel .
subset/123

...

# worker 1
$ launchable split-subset --subset-id subset/123 --bin 1/3 bazel --rest rest.txt > subset.txt
$ bazel test $(cat subset.txt)
$ launchable record tests --subset-id subset/123 .


# worker 2
$ launchable split-subset --subset-id subset/123 --bin 2/3 bazel --rest rest.txt > subset.txt
$ bazel test $(cat subset.txt)
$ launchable record tests --subset-id subset/123 .

# worker 3
$ launchable split-subset --subset-id subset/123 --bin 3/3 bazel --rest rest.txt > subset.txt
$ bazel test $(cat subset.txt)
$ launchable record tests --subset-id subset/123 .

```

## Changes

- Add split-subset subcommand
- Defined `TestPathWrite` to optimize the common logic of the subset command and the split-subset command
- Add `--subset-id` option to the record test command to support split subset


kohsuke:
If not done so already, it'd be great to hear from @awilkes about the proposed CLI UX. Command names, option names, etc.

kohsuke (commented):
So I note that this change touches every test runner support. And that begs the question if we can do any better.

It seems to me that what a test runner support needs to provide for the `subset` command and the `split-subset` command feels very similar, so much so that it looks feasible to just reuse the same `subset` implementation of each test runner, supply a slightly modified `client` object, and support both subset commands with zero changes to test runners.

What's behind this trade off is the idea that we generally want to treat test runner support as the "user land" and we want to keep that simpler thant we'd otherwise normally do, even at the expense of a little more complexity on the "kernel land".

Have that been considered?

awilkes:
Regarding UX - I think everything looks good, except for one question:

`--subset-id subset/123`: is `subset/` required here? I would have expected just the ID, e.g. `--subset-id 123`.

Konboi:
> --subset-id subset/123: is subset/ required here? I would have expected just the ID, e.g. --subset-id 123.

@awilkes Not required. I alined like the session-id style 
e.g) `builds/<BUILD NAME>/test_sessions/<TEST SESSION ID>` 

Should I change it?

Konboi:
https://github.com/launchableinc/cli/pull/258#pullrequestreview-701589410

if I don't have to implement each test runner to realize the split-subset, I want to do that. Because adding a new command to each test runner raise a cost to maintain too.

I'll try to change it again.

Konboi:
> it looks feasible to just reuse the same subset implementation of each test runner, supply a slightly modified client object, and support both subset commands with zero changes to test runners.

@kohsuke  Sorry, I don't understand it correctly so I have a question

you mean, below right?

- Add options for split subset command to subset.py
- Fix subset.py to be able to handle subset and split-subset
- Set alias `split-subset` to subset command e.g) `main.add_command(subset, "split-subset")` in __init__.py

If so, I worry that the commands will have unnecessary options.

awilkes:
> > --subset-id subset/123: is subset/ required here? I would have expected just the ID, e.g. --subset-id 123.
> 
> @awilkes Not required. I alined like the session-id style
> e.g) `builds/<BUILD NAME>/test_sessions/<TEST SESSION ID>`
> 
> Should I change it?

Oh, I understand now: the customer isn't going to type this value, it will be populated from a file! We do not need to change it 😄

Konboi:
I need more time to fix it https://github.com/launchableinc/cli/pull/258#pullrequestreview-701589410
I'll fix it another PR after release

PR #268, opened 2021-07-14
Remove file path prefixies in NUnit XML

ninjinkun:
We found some NUnit XML contains file path prefix like `<test-suite type="Assembly" name="C:\Works\Project1\calc.ddl}" …>`.  But subset command in the NUnit plugin doesn't include those file paths. To resolve this miss much, I implemented removing file prefixes for `record build … nunit …` command.


kohsuke (commented):
OK, so it's interesting that looking at the test data the `name` attribute is supposed to be already free of the directory prefix. I guess this means the customer's JUnit report has some idiosyncracies.

Given that, it's important that we leave some trails behind as to what data should look like (that this is supposed to be a file name only field), what varieties we've encountered in the wild (that some people had it full path -- ideally we'd track down what tool produced that file), and why a particular processing (in this case path removal) was deemed necessary.

A year down the road there will be 5 more variants and the code will look pretty cryptic, and when somebody tries to clean it up they'd appreicate all that audit trail behind.)

launchable/test_runners/nunit.py
```diff
@@ -15,9 +15,13 @@ def build_path(e: Element):
...
         # <test-suite>s form a nested tree structure so capture those in path
-        e.tags['path'] = pp + [{'type': e.attrs['type'], 'name': e.attrs['name']}]
+        pp = pp + [{'type': e.attrs['type'], 'name': e.attrs['name']}]
     if e.name == "test-case":
-        e.tags['path'] = pp + [{'type': 'TestCase', 'name': e.attrs['name']}]
+        pp = pp + [{'type': 'TestCase', 'name': e.attrs['name']}]
+
+    if len(pp) > 0:
+        # remove file path prefix
+        e.tags['path'] = [{**path, 'name': path['name'].split(os.sep)[-1]} for path in pp]
```
kohsuke, line 24: This is too broad for my comfort. IThis processing kicks in for every name, not just assembly name. I'm specifically worried about parameterized tests where names seem to include parameter values. I'd imagine it'd be possible for that to contain a path separator.

When we restrict this to just assembly names, it'd be also prudent to split not just by `os.sep` but either `\` or `/` -- in a complicated CI setup people often transfer test reports from one location to another, for example when parallelization is involved, so it's conceivable that some tests might run on Windows and then collected in Linux, and vice versa. Accepting both gracefully therefore would be a little more forgiving.

ninjinkun: Okay, it's a rough way. I'm going to fix this to focus on `Assembly` path, and to support both `/` and `\` regardless of OS.

ninjinkun:
@kohsuke I fixed it. Please review again.

launchable/test_runners/nunit.py
```diff
@@ -15,9 +16,26 @@ def build_path(e: Element):
...
+            # Supports Linux and Windows
+            if '/' in path:
+                return path.split('/')
+            else:
+                return path.split('\\')
+
+        # "Assembly" type containts full path at a cutomer's environment
+        # remove file path prefix in Assembly
+        e.tags['path'] = [
+            {**path, 'name': split_filepath(path['name'])[-1]}
```
kohsuke, line 34: TIL. I guess `path['name']` in `**path` gets overridden by the `'name': exp`

Looks like `{'foo':3, 'foo':5}` evaluates to `{'foo':5}` which makes sense but somehow I thought that would result in an error.

kohsuke (approved):

launchable/test_runners/nunit.py
ninjinkun: It's a common pattern in lightweight languages. For example, JavaScript has `{...obj}`, Ruby has `{**obj}`.

PR #270, opened 2021-07-20
Add a --file-attribute option to the File plugin

ninjinkun:
This is for a customer who records file path as `classname` in the result XML.

* Add `--file-attribute` option to the File plugin
  * It adds arbitrary XML attribute to the file attribute list such as `['file', 'filepath']`
* Add a test for the option


shuheiktgw (approved):
👍

shuheiktgw:
Do you need to update the documentaiton? Or are you going to handle it in another PR?

ninjinkun:
Sorry, this feature is no longer needed.

PR #272, opened 2021-07-30
--time option supports human readable suffix

kohsuke:
Better to have computers do the unit conversion than humans.


awilkes:
Can you update the CLI reference page too?

kohsuke:
Done.

awilkes (approved):
w00t, this is much better UX!

PR #276, opened 2021-08-04
Enclose Jar path including white space with double quotes

ninjinkun:
In our customer's environment, a path for `exe_deploy.jar` contained white space such as `c:\program files\...`. It causes a failure in `launchable record commit` command. To resolve it, I enclose the path with double-quotes.


ninjinkun:
I tested it on my local Mac and Windows VM

shuheiktgw (approved):
👍

PR #277, opened 2021-08-06
Contains test runner name to User Agent

Konboi:
Now, it is difficult to check that which test runner is used. This change helps us to resolve it.

### DIFF
--- a/tests/utils/test_http_client.py
+++ b/tests/utils/test_http_client.py
@@ -23,6 +23,5 @@ class LaunchableClientTest(TestCase):
         cli = LaunchableClient("/test", test_runner="dummy")
         self.assertEqual(cli._headers(False), {
             'Content-Type': 'application/json',
-            "User-Agent": "Launchable/{} (Python {}, {})".format(__version__, platform.python_version(), platform.platform()),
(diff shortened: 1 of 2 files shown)


tests/utils/test_http_client.py
```diff
@@ -23,6 +23,5 @@ def test_header(self):
         cli = LaunchableClient("/test", test_runner="dummy")
         self.assertEqual(cli._headers(False), {
             'Content-Type': 'application/json',
-            "User-Agent": "Launchable/{} (Python {}, {})".format(__version__, platform.python_version(), platform.platform()),
-            "launchable-test-runner": "dummy"
+            "User-Agent": "Launchable/{} (Python {}, {}) (TestRunner: {})".format(__version__, platform.python_version(), platform.platform(), "dummy"),
```
shuheiktgw, line 26: Sorry to nitpicking but `(Python {}, {})` does not contain any `:` but `(TestRunner: {})` does. I think either you should remove `:` from test runner or add `:` to the Python output.

Konboi: OK, I remove `:`

shuheiktgw (approved):
Looks good 👍

PR #284, opened 2021-08-17
Add an option to skip the commit collection

draftcode:
When a repository is shallow-cloned, a revision walk is not possible
further than the shallow depth. Add an option to the record build
command to skip the commit collection and let the user upload the commit
data in a separate workflow.

### DIFF
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -34,8 +34,19 @@ from ...utils.session import clean_session_files
     help="stop collecting information from Git Submodules",
     default=False
 )
+@click.option('--no-commit-collection',
+    is_flag=True,
+    help="""do not collect commit data.
+
+    This is useful if the repository is a shallow clone and the RevWalk is not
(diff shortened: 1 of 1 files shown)


yoshiori (approved):

awilkes:
I think we still need to support "tagging" the build with a commit hash (for the purpose of computing CUT later), right?

draftcode:
> I think we still need to support "tagging" the build with a commit hash (for the purpose of computing CUT later), right?

Yep. This doesn't stop that part (see the code after the commit collection). It gathers the submodules' and superproject's commit hashes.

awilkes:
> > I think we still need to support "tagging" the build with a commit hash (for the purpose of computing CUT later), right?
> 
> Yep. This doesn't stop that part (see the code after the commit collection). It gathers the submodules' and superproject's commit hashes.

Ah, of course! Thank you. I'll stay in my lane now 😄

kohsuke (approved):

PR #286, opened 2021-08-18
rm bracket

Konboi:
SSIA

### DIFF
--- a/launchable/utils/http_client.py
+++ b/launchable/utils/http_client.py
@@ -76,7 +76,7 @@ class LaunchableClient:
 
         if self.test_runner != "":
             h["User-Agent"] = h["User-Agent"] + \
-                " (TestRunner {})".format(self.test_runner)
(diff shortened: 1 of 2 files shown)


ninjinkun:
@Konboi I also proposed to add a slash like `TestRunner/gradle` at the Slack post.

Konboi:
@ninjinkun I fixed. Could you review again?

ninjinkun (approved):
LGTM!

awilkes:
Do we need to do anything special for 'custom' profiles? We could end up with a conflict if a customer named their profile the same as one of ours (unlikely, but you never know)

Perhaps `TestRunner#gradle` for standard ones and `TestRunner#{org}/{ws}/{name}` for custom? Just an idea...

Konboi:
I think don't need it.
We should protect not to register custom plugins as general plugin names than change UA names.

What do you think?

awilkes:
Sounds good! Good idea.

PR #287, opened 2021-08-18
Guess the base path for relativizing test file paths 

draftcode:
Some test runners produce absolute paths. This makes a problem when we make statistical data for tests as those absolute paths can vary per machine or per run.

The CLI has an option to specify a base path to make a relative path from those absolute paths. This change teaches the CLI to guess the base path based on the Git repository root directory.

### DIFF
--- a/launchable/commands/record/case_event.py
+++ b/launchable/commands/record/case_event.py
@@ -3,7 +3,7 @@ import os
 import sys
 from typing import Callable, Dict
 from junitparser import Failure, Error, Skipped, TestCase, TestSuite  # type: ignore
-from ...testpath import TestPath
+from ...testpath import TestPath, FilePathNormalizer
 
 
 class CaseEvent:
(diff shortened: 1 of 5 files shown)


draftcode:
I guess the failure is caused by the commit author identity. Fixing...

launchable/commands/record/case_event.py
```diff
@@ -17,7 +17,7 @@ class CaseEvent:
     TestPathBuilder = Callable[[TestCase, TestSuite, str], TestPath]
 
     @staticmethod
-    def default_path_builder(base_path) -> TestPathBuilder:
+    def default_path_builder(file_path_normalizer: FilePathNormalizer) -> TestPathBuilder:
```
ninjinkun, line 20: It's a good design 👍 At first I couldn't understand why you designed FilePathNormalizer as a class, not a function. But later, I was convinced that the class holds basepaths and sets the instance to default path builder.

ninjinkun (approved):
Looks good. Clap for your fight with Windows 👏

Konboi (approved):
LGTM

PR #290, opened 2021-09-08
Add a pre-commit hook for the style check

draftcode:
This change introduces a git pre-commit hook configuration for the style
check.

[pre-commit](https://pre-commit.com) is a utility for Git pre-commit
hook. This can be installed with `pip install pre-commit`, and a Git
pre-commit hook based on this utility can be installed with `pre-commit
install`. Once installed, it checks the styles based on
.pre-comit-config.yaml. Currently, it checks the line-endings of the end
of files and applies autopep8.

The Git pre-commit hook is trigged upon `git commit`. When there's a
style change needed, autopep8 automatically fixes it and cancels the
`git commit` command. You need to check the modified files, add them
with `git add`, and run `git commit` again. If there's no change needed
by autopep8, it'll just continue.


draftcode:
I think it's better to split the python version update part.

ninjinkun:
This is great! I have an additional idea: how about adding `pre-commit` module and `pre-commit install` command to `Pipfile`? It will make a little bit easier to configure the module.

ninjinkun (approved):

draftcode:
> This is great! I have an additional idea: how about adding `pre-commit` module and `pre-commit install` command to `Pipfile`? It will make a little bit easier to configure the module.

Done.

PR #291, opened 2021-09-08
Update the minimal Python version to 3.6

draftcode:
Python 3.5 is already end of life. [pre-commit](https://pre-commit.com)
doesn't work with it. Update to 3.6.

This includes pyproject.toml, which is required for PEP 518. The context
is well described in the setuptools document:

> The landscape of Python packaging is shifting and Setuptools has
> evolved to only provide backend support, no longer being the de-facto
> packaging tool in the market. All python package must provide a
> pyproject.toml and specify the backend (build system) it wants to use.
>
> https://setuptools.readthedocs.io/en/latest/userguide/quickstart.html#python-packaging-at-a-glance

Based on this and the setup instruction on
https://github.com/pypa/setuptools_scm, pyproject.toml is checked in and
setup.cfg and setup.py are modified.

Pipfile's python_version becomes X.Y per
https://pipenv.pypa.io/en/latest/basics/#general-recommendations-version-control


ninjinkun (approved):
I agree to use `pyproject.toml`.

PR #297, opened 2021-09-16
[ENG-315] Change pytest from class base to file name base

yoshiori:
- pytest has changed its default test report format from **xunit1** to **xunit2** since version 6.
  -  https://docs.pytest.org/en/latest/deprecations.html#junit-family-default-value-change-to-xunit2
- The **xunit2** format no longer includes file names.
  - The **xunit1** format includes the file name.
  - It is possible to output in xunit1 format by specifying `junit_family=legacy`.

I chose to generate the class name from the file name so that I wouldn't have to change the format. However, it was not possible to create a complete class name from a file name. So I'm going to change it to file name based.

### DIFF
--- a/launchable/test_runners/pytest.py
+++ b/launchable/test_runners/pytest.py
@@ -6,17 +6,18 @@ import glob
 from . import launchable
 from ..utils.file_name_pattern import pytest_test_pattern
 
-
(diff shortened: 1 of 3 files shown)


awilkes (approved):

Konboi (approved):
LGTM

PR #298, opened 2021-09-17
Print the system information first

draftcode:
This makes sure that we can get the system information even if there's
an error. The values are quoted by repr. This converts NoneType into a
str and shows non-printing characters if there's any.


Konboi (approved):
👍

PR #303, opened 2021-09-24
Add an option for ctest to write regexp into split files

draftcode:
CTest has a size limit in the regexp engine
(https://gitlab.kitware.com/cmake/cmake/-/blob/2974ea7c3e9cbf9952d6c361bbcecb5027f69605/Source/kwsys/RegularExpression.cxx#L362).
Because of this, we cannot specify regexp for test subsets.

This change teaches the ctest profile to split the regexp and output
files in the specified directory. The caller can opt-in to this new
behavior with --output-regex-files. With this, the CLI writes out the
regexp to multiple files, limiting each file size so that it fits to
the CTest regexp upperbound.


Konboi (approved):
LGTM

Nice Work 👍

PR #307, opened 2021-10-07
upgrade pytest support

yoshiori:
Support the `--collect-only` option to get detailed information.

### DIFF
--- a/launchable/test_runners/pytest.py
+++ b/launchable/test_runners/pytest.py
@@ -12,19 +12,53 @@ from ..utils.file_name_pattern import pytest_test_pattern
 # - The xunit2 format no longer includes file names.
 #   - It is possible to output in xunit1 format by specifying junit_family=legacy.
 #   - The xunit1 format includes the file name.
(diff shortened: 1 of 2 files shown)


launchable/test_runners/pytest.py
```diff
@@ -1,33 +1,98 @@
...
+    >
+    > 2 tests collected in 0.02s
+
+    result.xml(junit)
+    <testcase classname="tests.fooo.func4_test" name="test_func6" file="tests/fooo/func4_test.py" line="0" time="0.000" />
+    <testcase classname="tests.test_mod.TestClass" name="test__can_print_aaa" file="tests/test_mod.py" line="3" time="0.001" />
+    '''
+    def _add_testpaths(lines: List[str]):
+        for label in lines:
+            label = label.rstrip()
```
Konboi, line 35: Sorry for the detail. I think this `for line in lines:` is better, I think.
But you can skip my comment

yoshiori: Fixed. thanks

PR #315, opened 2021-10-15
Add raw profile

draftcode:
Record test is not implemented yet.


okajimad (approved):
I don't know much about this part, but let's go.

awilkes:
Great points. Thanks Masaya!

PR #316, opened 2021-10-15
Replace subprocess.check_output with a wrapper

draftcode:
See the comment in launchable.utils.subprocess. This fixes Windows
build.

### DIFF
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -1,6 +1,6 @@
 import re
 import click
-import subprocess
+from ...utils import subprocess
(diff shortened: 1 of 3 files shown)


kohsuke:
Oops did I push the change before verifying that the fix works on Window?

draftcode:
I still do not know why this surfaced on my PR (the raw profile PR). But it seems that this wasn't surfaced on your original PR.

kohsuke:
And of course the momeny I wrote it, I found it. It's `mock.side_effect`

PR #317, opened 2021-10-18
Change test session file format

Konboi:
We had used a UNIX process id as a hash key to saving a test session. But we sometimes couldn't get the same UNIX process id through a CI job in some environments. IOW, a UNIX process id was changed each step in a certain environment. 
Also, the cli has been created a new test session file when the UNIX process id changes, missing introduce the cli, don't need to create a new one but create and etc.

image

So I changed to stop using a UNIX process id and to use `<SESSION DIR>/.launchable` file to avoid miss integration.

If the user tries to create a new build after didn't run record tests due to an error or misconfiguration, the user will get an error.
It may not have been a problem in the previous version, but since this version, it is an error to make it easier to detect misconfigurations.

image

And tests of the cli has deleted test session files that was created in CI pipeline, so I changed to avoid deleting test session file.

### DIFF
--- a/launchable/utils/session.py
+++ b/launchable/utils/session.py
@@ -15,38 +15,7 @@ def _session_file_dir() -> Path:
 
 
 def _session_file_path(build_name: str) -> Path:
-    return _session_file_dir() / (hashlib.sha1("{}:{}".format(build_name, _get_session_id()).encode()).hexdigest() + ".txt")
-
-
-def _get_session_id() -> str:
-    # CircleCI changes unix session id each steps, so set non change variable
-    # https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
-    if os.environ.get("CIRCLECI") is not None:
-        id = os.environ.get("CIRCLE_WORKFLOW_ID")
-        if id is not None:
-            return id
-        Logger().warning("CIRCLECI environment variable is set but not CIRCLE_WORKFLOW_ID")
-
-    # Jenkins pipeline also launches every process as a new session, so better to scope this to a particular build.
-    # For freestyle job types that do not do this, session IDs do not distinguish different builds, so that is a problem, too.
-    #
-    # See: https://github.com/jenkinsci/lib-durable-task/blob/6e020747205cb5aca5af757bc0d2a302c42bdb79/src/cmd/bash/durable_task_monitor.go#L41
-    # See: https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables
-    if os.environ.get("JENKINS_URL") is not None:
-        id = os.environ.get("BUILD_URL")
-        if id is not None:
-            return id
-        Logger().warning("JENKINS_URL environment variable is set but not BUILD_URL")
-
-    if sys.platform == "win32":
-        import wmi  # type: ignore
-        c = wmi.WMI()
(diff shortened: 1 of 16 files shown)


ninjinkun (approved):

Konboi:
Sorry, I discussed with @kohsuke before and I realized need to fix some codes. 
So, this pr back to draft this PR and I'll ask for review again later

launchable/utils/session.py
```diff
@@ -14,47 +12,33 @@ def _session_file_dir() -> Path:
     return Path(os.environ.get(SESSION_DIR_KEY) or DEFAULT_SESSION_DIR).expanduser()
```
kohsuke, line 12: This might be a result of the evolution of thinking, but I notice that it's no longer tied to the current directory. Is that intentional?

This reads to me that it's using a known fixed location in user's home directory, which means if multiple builds are running with the same user ID (which is common in Jenkins) different builds will cross wire.

launchable/utils/session.py
```diff
@@ -14,47 +12,33 @@ def _session_file_dir() -> Path:
...
-            os.getpid())
-        res = c.query(wql)
-        return str(res[0].LogonId)
-    else:
-        return str(os.getsid(os.getpid()))
+def _session_file_path() -> Path:
+    return _session_file_dir() / ".launchable"
+
+
+def _parse_session_file(text: str):
```
kohsuke, line 19: It's generally preferrable that a file like this follows a reasonably well understood syntax that has room for extension.
This code seems to assume that the file is always in `a=b#c=d` and it's not actually even checking if `a` and `c` are `build` and `session`. And what if build name contains `#`? Doesn't seem like too unusual a character.

It'd be better for parsing logic like this to be a little more robust. Say assume arbitrary `k=v` each separated with NL.

How far you go would be up to you, but at least I'd suggest you make sure `a` and `c` are `build` and `session` as expected.

launchable/utils/session.py
```diff
@@ -14,47 +12,33 @@ def _session_file_dir() -> Path:
...
-    f = _session_file_path(build_name)
+    f = _session_file_path()
     try:
         if not f.exists():
             return None
-        return f.read_text()
+        build, session_id = _parse_session_file(f.read_text())
+
+        if build != build_name:
+            raise Exception("Build name is different between input:{} and saved:{}. Previous job might be failed.\nPlease confirm previous job result and run after removing {}".format(
```
kohsuke, line 38: How is this error message eventually get shown to users?

CLI error handling generally falls into two categories; if an error is anticipated, as in this case, then the error message should not be shown with a stack trace, which distracts users. If an error is unanticipated, then we want to see the entire stack trace so that we can understand where the problem happened and how.

My typical design pattern is to designate an exception type for anticipated errors, so that the code that's catching it can use the type as a hint to decide what to do. When I see a raw `Exception` being thrown like this, I don't think we can make the distinction above, which degrade the user experience.

This is our CLI design deficiency that goes beyond this PR, but we might as well fix it here (or in the next PR, I suppose)

kohsuke (commented):
My apologies for a delay to come back to this. I haven't kept up with the Confluence page that designs this, so this could very well be that I haven't seen the latest evolution of thoughts, but some of the behaviour was surprising to me, so I left those as comments.

Konboi:
I realized that I miss understanding from you and Al's feedbacks. 
While I was implementing this change, I overlooked something important in order to simplify the changes.
I'll get a chill out and implement it again

Konboi:
I made a new PR
https://github.com/launchableinc/cli/pull/342

PR #318, opened 2021-10-21
provide proper audit log

Songmu:
by passing -audit flag to jar when the log level is AUDIT


Konboi (approved):
LGTM

___

I could confirm new audit logs at my environment

PR #329, opened 2021-11-09
Jest plugin reacords class names

ninjinkun:
From a customer's feedback, we decided to record class names in the Jest plugin.


Konboi (approved):
LGTM

PR #331, opened 2021-11-10
Specify `more_itertools` version lower than v8.6 in Python 3.5

ninjinkun:
[more_itertools](https://more-itertools.readthedocs.io/en/stable/versions.html) dropped python 3.5 support since v8.6.


ninjinkun:
You can see it worked at https://github.com/launchableinc/cli/runs/4161202858

draftcode (approved):

PR #335, opened 2021-11-12
create temp dir and use it for session dir in tests

Konboi:
A session file that was created in this repo's CI was deleted by tests (by clean_test_session_files method)
Then, two test sessions were created 1) subset command, 2) record test command but it's not intented.

So I create a temp dir  and use it in test to avoid deleting a session file that this repository created in CI.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -35,7 +35,14 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
+    - name: Install specific dependencies in 3.6
+      if: matrix.python-version == '3.6'
+      run: |
+        python -m pip install --upgrade pip
+        pip install pipenv==2021.11.5
+        pipenv install --dev --python ${{ matrix.python-version }}
     - name: Install dependencies
+      if: matrix.python-version != '3.6'
       run: |
         python -m pip install --upgrade pip
         pip install pipenv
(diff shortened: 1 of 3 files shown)


tests/cli_test_case.py
```diff
@@ -48,6 +51,7 @@ def setUp(self):
 
     def tearDown(self):
         clean_session_files()
+        del os.environ[SESSION_DIR_KEY]
```
ninjinkun, line 54: With `tempfile.mkdtemp`, you need to clean up the temp file manually https://stackoverflow.com/questions/6884991/how-to-delete-a-directory-created-with-tempfile-mkdtemp/55871731

Konboi: I fixed it

ninjinkun (approved):
LGTM. Kudos for your python 3.6 workaround 👍

.github/workflows/python-package.yml
```diff
@@ -35,7 +35,14 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
+    - name: Install specific dependencies in 3.6
+      if: matrix.python-version == '3.6'
+      run: |
+        python -m pip install --upgrade pip
+        pip install pipenv==2021.11.5
```
Konboi, line 42: https://github.com/pypa/pipenv/issues/4865

PR #341, opened 2021-11-26
Add dependency caching on Github Actions

ninjinkun:
Recently, GitHub Actions started to support dependency caching for `pip` and `pipenv` https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/ .


ninjinkun:
We can't use this option because we don't fix module versions. `Pipfile.lock` is required.

PR #342, opened 2021-11-29
Change test session format

Konboi:


launchable/utils/session.py
```diff
@@ -1,102 +1,103 @@
...
+
+def read_build() -> Optional[str]:
+    f = _session_file_path()
+    try:
+        if not f.exists():
+            return None
+
+        with open(str(_session_file_path())) as session_file:
+            session = json.load(session_file)
+            return session["build"]
```
kohsuke, line 25: This line will raise `KeyError` if the session doesn't contain build, whiich is probably not what we want. I think `session["build"]` is what we want:
```
>>> x=json.loads('{"a":3}')
>>> x["b"]
  File "<stdin>", line 1, in <module>
KeyError: 'b'
>>> x.get("b")
>>> 
```

launchable/commands/helper.py
```diff
@@ -20,6 +20,14 @@ def find_or_create_session(context, session: Optional[str], build_name: Optional
     if session:
         return session
     elif build_name:
+        if not read_build():
+            raise click.UsageError(click.style(
+                "Make sure to run `launchable record build --name {}` before, \nOr this command was ran on another machine what was ran `launchable recrod build` command. So please use --session option to run".format(build_name), fg="yellow"))
+
+        if build_name != read_build():
```
kohsuke, line 27: NIT: reading the same data twice unnecessarily.

launchable/commands/helper.py
```diff
@@ -20,6 +20,14 @@ def find_or_create_session(context, session: Optional[str], build_name: Optional
     if session:
         return session
     elif build_name:
+        if not read_build():
+            raise click.UsageError(click.style(
+                "Make sure to run `launchable record build --name {}` before, \nOr this command was ran on another machine what was ran `launchable recrod build` command. So please use --session option to run".format(build_name), fg="yellow"))
```
kohsuke, line 25: TODO: note to myself; @Konboi and I discussed this part and there will be some change here, so review this message afterward.

launchable/utils/session.py
```diff
@@ -1,102 +1,103 @@
...
 
-        _session_file_path(build_name).write_text(session_id)
+        session = {}
+        session["build"] = build_name
+
+        with open(str(_session_file_path()), 'w') as session_file:
+            json.dump(session, session_file)
+
+    except Exception as e:
+        raise Exception("Can't write to {}. Perhaps set the {} environment variable to specify an alternative writable path?".format(
```
kohsuke, line 61: I might suggest "Can't write to {}. Is the path writable? If not, set the {} environment variable to specify an alternative directory for this file"

launchable/commands/helper.py
Konboi: I finished to change, so could you please review again?

launchable/commands/helper.py
```diff
@@ -1,31 +1,39 @@
...
+            "Given build name ({}) is different from when you ran `launchable record build --name {}`.\nMake sure to run `launchable record build --name {}` before.".format(build_name, saved_build_name, build_name), fg="yellow"))
+
     if session:
         return session
-    elif build_name:
-        session_id = read_session(build_name)
+    else:
+        if not saved_build_name:
+            raise click.UsageError(click.style(
+                "Have you run `launchable record build`?\nIf not, please do. If it was run elsewhere/earlier, please use the --session option", fg="yellow"))
```
kohsuke, line 31: I modified the message and in particular not to use `build_name` as that is optional and the value might not be present. More so in the future as `--build` option is legacy.

launchable/commands/record/tests.py
```diff
@@ -91,7 +92,7 @@ def tests(context, base_path: str, session: Optional[str], build_name: Optional[
     else:
         session_id = find_or_create_session(
             context, session, build_name, flavor)
-
+        build_name = read_build()
```
kohsuke, line 95: IIUC, this is problematic.

My first "hmm" was because this overwrites what the user has given. Upon further thoughts, the case where this goes wrong is when we are talking a distributed test and the user uses the `--session` option. We do not expect a context session file in that case, so `read_build()` will return `None`, even if the user specifies some valid value in `--build`. But the real problem in this case is, why would the user have to specify the build name?

What `get_record_start_at` is trying to do is to compute the timestamp of the build. Looking at the logic, it just needs a valid test session ID, which in this context we are guaranteed to have.

So I think the next line should be changed to `record_start_at = get_record-start_at(None, session_id)`

kohsuke (approved):
I went ahead and tweaked a few messages and pushed them. There's one line where I think the code is a tad problematic, and I left that as a comment. Otherwise this looks good.

PR #345, opened 2021-12-06
Add commit option

Konboi:
When the user uses `--no-commit-collection` option, the cli can't send commit hashes. 
So I add `--commit` option to be able to send repository name and commit hash.

```
$ launchable record build --name build1 --no-commit-collection --commit repoA:hash123456 --commit repoB:hash67890
Warning: Commit collection is turned off. The commit data must be collected separately.
Launchable recorded build build1 to workspace launchableinc/mothership with commits from 3 repositories:

| .      | .      | 6bf701222e2b56a935862a78eb8b40e5a8fcaeb2 |
| repoA  |        | hash123456                               |
| repoB  |        | hash67890                                |
```

### DIFF
--- a/launchable/utils/click.py
+++ b/launchable/utils/click.py
@@ -49,7 +49,7 @@ class DurationType(click.ParamType):
 
 
 class KeyValueType(click.Option):
-    error_message = "Expected key-value like --option kye=value or --option key value. but got '{}'"
+    error_message = "Expected key-value like --option kye=value, --option key:value or --option key value. but got '{}'"
 
     def __init__(self, *args, **kwargs):
         super(KeyValueType, self).__init__(*args, **kwargs)
--- a/launchable/utils/click.py
+++ b/launchable/utils/click.py
@@ -66,6 +66,14 @@ class KeyValueType(click.Option):
                         self.error_message.format(value))
 
                 value = tuple([kv[0].strip(), kv[1].strip()])
+            # case: --option key:value
+            elif ':' in value:
+                kv = value.split(':')
+                if len(kv) != 2:
+                    raise ValueError(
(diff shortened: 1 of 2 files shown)


launchable/utils/click.py
```diff
@@ -65,6 +65,14 @@ def parser_process(value, state):
                     raise ValueError(
```
draftcode, line 65: Change error_message above for the new case (I cannot comment on that line...)

draftcode (approved):

awilkes:
Can we validate that the commit exists in the database?

awilkes:
@draftcode and I discussed server-side versus client-side validation and decided that client-side validation (i.e. check that the input looks like `0-9a-fA-F`) is sufficient for now. The reason is that `--no-commit-collection` is already an edge case (so we aren't sure how much extra time we want to invest in it), there might be legitimate use cases where server-side validation will break things, and we can add it later, so we'll wait and see.

Konboi:
@draftcode I added client validation logic, so could you review it again, just in case?

launchable/utils/click.py
```diff
@@ -49,7 +49,7 @@ def convert(self, value: str, param, ctx):
 
 
 class KeyValueType(click.Option):
-    error_message = "Expected key-value like --option kye=value or --option key value. but got '{}'"
+    error_message = "Expected key-value like --option kye=value, --option key:value or --option key value. but got '{}'"
```
awilkes, line 52:
```suggestion
    error_message = "Expected a key-value pair formatted as --option key=value, --option key:value, or --option key value, but got '{}'"
```

draftcode (approved):

PR #349, opened 2021-12-14
Add a constraint on more_itertools in Pipfile

draftcode:
Apparently, one of those dev dependencies pulls latest more_itertools,
which conflicts with the main app's dependency. Pin the version in the
dev dependencies as well.


Konboi (approved):
Thank you

PR #350, opened 2021-12-14
use allowed_method option instead of method_whitelist

Songmu:
To suppress DeprecationWarning.

I explicitly specify the version of urllib3 that the request package depends on as 1.26 or later. This is the version in which `allowed_methods` was introduced.

- https://github.com/urllib3/urllib3/blob/1dd0613b610df1269dbc11e9780b0fe7debaff3c/CHANGES.rst#1260-2020-11-10
- https://github.com/urllib3/urllib3/pull/2000

Also, the requests library has been updated to support urllib3 v1.26 since version 2.25, so I specify that as well.

https://github.com/psf/requests/blob/main/HISTORY.md#2250-2020-11-11

These changes meet our version requirements to support python 3.5.

### DIFF
--- a/launchable/utils/http_client.py
+++ b/launchable/utils/http_client.py
@@ -39,7 +39,7 @@ class LaunchableClient:
         if session is None:
             strategy = Retry(
                 total=3,
-                method_whitelist=["GET", "PUT", "PATCH", "DELETE"],
+                allowed_methods=["GET", "PUT", "PATCH", "DELETE"],
                 status_forcelist=[429, 500, 502, 503, 504],
                 backoff_factor=2
             )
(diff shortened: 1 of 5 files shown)


launchable/utils/http_client.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, base_url: str = "", session: Session = None, test_runner: str
         if session is None:
             strategy = Retry(
                 total=3,
-                method_whitelist=["GET", "PUT", "PATCH", "DELETE"],
+                allowed_methods=["GET", "PUT", "PATCH", "DELETE"],
```
Konboi, line 42: 👍

Konboi (approved):
LGTM

PR #352, opened 2021-12-14
drop --executable docker option

Songmu:
It may be that no one uses the docker option. And the docker image is old, not following up with the recent new features of the cli such as audit and dry-run.

So, how about deleting it?

I leave the option as a hidden flag for now. I plan to remove it completely in the future.


Konboi (approved):
nice 🔪

ninjinkun (approved):
Let's go!

PR #353, opened 2021-12-15
Update exe_deploy.jar

draftcode:
This includes log4j2 update to 2.16.0.

Implications to the users
=========================

The binary contains log4j-api, but it doesn't include log4j-core. This
means that the binary wasn't affected by the recently discovered
vulnerabilities (CVE-2021-44228, CVE-2021-45046) in the first place.
This can be verified by checking the existence of JndiLookup.class as
instructed in https://www.cve.org/CVERecord?id=CVE-2021-44228 and
https://www.cve.org/CVERecord?id=CVE-2021-45046.

Based on this, the users of the old binaries should not be affected by
the forementioned vulnerabilities. However, we encourage users to keep
the launchable cli up-to-date, even if it's not affected.


Konboi (approved):
LGTM 

Thank you!

PR #355, opened 2021-12-15
Add/isort

shibuiwilliam:
# What
- add isort to sort imports
- run isort


Songmu (approved):
Thanks. I didn't know about isort.
Introducing this kind of toolchain is very helpful.

PR #357, opened 2021-12-17
update readme; add installation

shibuiwilliam:
# What
- add installation in readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,30 @@ pipenv install --dev some-what-module
 Create new release on Github, then Github Actions automatically uploads the
 module to PyPI.
 
+# Installing CLI
+
+You can install the `launchable` command from either source or [pypi](https://pypi.org/project/launchable/).
+
+## Prerequisite
+
+- >= Python 3.5
+- >= Java 8
+
+## Install from source
+
+```sh
+$ pwd
+~/cli
+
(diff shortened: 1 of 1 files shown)


Konboi:
I think we don't need to add this. Because this part already exists https://docs.launchableinc.com/getting-started.
We should keep README.md for developers, I think.
How about you??

draftcode (approved):

draftcode:
I'm ok with having both. Both anyway have the same content, right?

shibuiwilliam:
My point is to add `python setup.py install` to help a developer to install it with local changes.
It is true that users read the [docs](https://docs.launchableinc.com/getting-started) and that should be the main document rather than the github readme.

yoshiori:
[Off topic]
`pip3 install --editable .` 
I often forget to do this, so I should probably write this down somewhere, too.

shibuiwilliam:
> pip3 install --editable .

True, this is fine as well. I'm just use to run `python setup.py install`.
`pip3 install git+git@github.com:launchableinc/cli.git@${COMMIT_HASH}` may be fine as well.

Konboi:
> My point is to add python setup.py install to help a developer to install it with local changes.

I see, LGTM

PR #360, opened 2021-12-20
Fix/record tests error handling

shibuiwilliam:
# What
- updated error handling on `launchable record tests` command
- Added custom exceptions to manage cli-specific exceptions https://github.com/launchableinc/cli/pull/360/commits/ec46c668a8edb1d688be4a7d792f43cc58d87b8c
- added org and workspace verification at the top of the `tests` function to let the function fail before going deeper. same for requesting `GET verification`. https://github.com/launchableinc/cli/pull/360/commits/9d005e6730f8adc99fefaebc8df90f653cf69940
- fixed tests for `record tests` to follow the change of having first `GET verification` https://github.com/launchableinc/cli/pull/360/commits/97d6473b82630dc0b71e99da2eb3edc73f651f36

# Why
- it might be better to stop the cli  if the authorization fails
- Since there are multiple exception that are cli-specific and launchable-specific, I propose to make custom exceptions to handle those errors


launchable/commands/record/tests.py
```diff
@@ -80,19 +81,39 @@
 @click.pass_context
 def tests(context, base_path: str, session: Optional[str], build_name: Optional[str], post_chunk: int, subsetting_id: str,
           flavor, no_base_path_inference):
+    org, workspace = get_org_workspace()
+
+    if org is None or workspace is None:
```
shibuiwilliam, line 86: adding org and workspace verification beforehand

launchable/commands/record/tests.py
```diff
@@ -80,19 +81,39 @@
...
+        raise click.UsageError(click.style(
+            "Could not identify Launchable organization/workspace. "
+            "Please confirm if you set LAUNCHABLE_TOKEN or LAUNCHABLE_ORGANIZATION and LAUNCHABLE_WORKSPACE environment variables",
+            fg="red"))
+
+    client = LaunchableClient(test_runner=context.invoked_subcommand,
```
shibuiwilliam, line 92: initialize client at the top and reuse it through the function

launchable/commands/record/tests.py
```diff
@@ -80,19 +81,39 @@
...
+                              dry_run=context.obj.dry_run)
+    res = client.request("get", "verification")
+    if res.status_code == HTTPStatus.UNAUTHORIZED:
```
shibuiwilliam, line 95: verify connection and authentication before going into the main processes

launchable/commands/record/tests.py
```diff
@@ -140,14 +161,20 @@ def set_junitxml_parse_func(self, f: JUnitXmlParseFunc):
             def parse(report: str) -> Generator[CaseEventType, None, None]:
                 # To understand JUnit XML format, https://llg.cubic.org/docs/junit/ is helpful
                 # TODO: robustness: what's the best way to deal with broken XML file, if any?
-                xml = JUnitXml.fromfile(report, f)
+                try:
+                    xml = JUnitXml.fromfile(report, f)
+                except Exception as e:
+                    click.echo(click.style("Warning: error reading JUnitXml file {filename}: {error}".format(
```
shibuiwilliam, line 167: catch error on opening junitxml

launchable/commands/record/tests.py
```diff
@@ -380,18 +401,18 @@ def parse_launchable_timeformat(t: str) -> datetime.datetime:
...
     if len(s) != 2:
         raise click.UsageError(
             'Invalid subset id. like `subset/123/slice` but got {}'.format(subsetting_id))
 
-    res = LaunchableClient().request("get", subsetting_id)
+    res = client.request("get", subsetting_id)
     if res.status_code != 200:
         raise click.UsageError(click.style("Unable to get subset information from subset id {}".format(
-            subsetting_id), 'yellow'), err=True)
+            subsetting_id), 'yellow'))
```
shibuiwilliam, line 415: remove `err=True` for `click.UsageError` does not take `err` argument

tests/commands/record/test_tests.py
```diff
@@ -37,7 +37,7 @@ def remove_backslash(input: str) -> str:
 
         # normal.xml
         self.assertIn('open_class_user_test.rb', gzip.decompress(
-            responses.calls[2].request.body).decode())
+            responses.calls[3].request.body).decode())
```
shibuiwilliam, line 40: updating response index for the `record tests` got an initial verification request.
same for the rest

launchable/commands/record/tests.py
```diff
@@ -80,19 +81,39 @@
...
+        raise click.UsageError(click.style(
+            "Could not identify Launchable organization/workspace. "
+            "Please confirm if you set LAUNCHABLE_TOKEN or LAUNCHABLE_ORGANIZATION and LAUNCHABLE_WORKSPACE environment variables",
+            fg="red"))
+
+    client = LaunchableClient(test_runner=context.invoked_subcommand,
+                              dry_run=context.obj.dry_run)
+    res = client.request("get", "verification")
+    if res.status_code == HTTPStatus.UNAUTHORIZED:
+        raise click.UsageError(click.style("Authentication failed. Most likely the value for the LAUNCHABLE_TOKEN "
```
yoshiori, line 96: There is also a tokenless mode, so you should change the error message.
https://docs.launchableinc.com/getting-started/using-the-cli-with-a-public-repository

launchable/commands/record/tests.py
yoshiori: I believe that the shorter the scope of the variable and the time it is used, the better, if possible. What is the reason for using them around?

launchable/commands/record/tests.py
```diff
@@ -80,19 +81,39 @@
...
+
+    if org is None or workspace is None:
+        raise click.UsageError(click.style(
+            "Could not identify Launchable organization/workspace. "
+            "Please confirm if you set LAUNCHABLE_TOKEN or LAUNCHABLE_ORGANIZATION and LAUNCHABLE_WORKSPACE environment variables",
+            fg="red"))
+
+    client = LaunchableClient(test_runner=context.invoked_subcommand,
+                              dry_run=context.obj.dry_run)
+    res = client.request("get", "verification")
```
yoshiori, line 94: Why not call the verify method instead of sending the request directly to make exception handling and messages common? You can leave the checking of org  and work to them too.

yoshiori (approved):
Direction looks good.

launchable/commands/record/tests.py
shibuiwilliam: fixed
https://github.com/launchableinc/cli/pull/360/commits/b00851afcb13d27e8626430e6c5f8a019426ef2a

launchable/commands/record/tests.py
shibuiwilliam: I don't think calling the command line method from another command line method is well-organized way since each command line has its own logic, ends up in change in `verify` command line affecting `record tests` unintentionally. Rather, I propose to make an independent `verify` method outside of command lines to be used generally to ensure the auth and connection are all fine beforehand.

launchable/commands/record/tests.py
shibuiwilliam: used only within a method is short enough, and I don't think taking overhead to initialize a client so often is good choice.

launchable/commands/record/tests.py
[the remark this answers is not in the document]
Konboi: Or, how about creating a validation.py and using it.

subset command want to validate too, I think
https://github.com/launchableinc/cli/pull/360/files#diff-a54ea2cd0c54426ceb96ffb93873062a86aaa7d4d4cb1310ab874ed0eeda2e85R92

Konboi (commented):
And you need to add error handling here too https://github.com/launchableinc/cli/pull/360/files#diff-a54ea2cd0c54426ceb96ffb93873062a86aaa7d4d4cb1310ab874ed0eeda2e85R272 

e.g) when API server returns 5xx error

launchable/commands/record/tests.py
shibuiwilliam: removed requesting verification after discussion with konboi-san for it is too much to do the verification here.
https://github.com/launchableinc/cli/pull/360/commits/b10d4071ee7d78aa8f3c2ba580316a48203391c2

launchable/utils/authentication.py
```diff
@@ -15,6 +17,16 @@ def get_org_workspace():
     return os.getenv(ORGANIZATION_KEY), os.getenv(WORKSPACE_KEY)
 
 
+def ensure_org_workspace() -> Tuple[str, str]:
```
shibuiwilliam, line 20: add `ensure_org_workspace` to assure the parameters are fine, else raise exception

launchable/commands/record/tests.py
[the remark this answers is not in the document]
shibuiwilliam: fixed to ensure org and workspace are set, and do try/except on session control
https://github.com/launchableinc/cli/pull/360/commits/b10d4071ee7d78aa8f3c2ba580316a48203391c2

tests/cli_test_case.py
```diff
@@ -50,6 +50,10 @@ def setUp(self):
                       json={}, status=200)
         responses.add(responses.GET, "{}/intake/organizations/{}/workspaces/{}/builds/{}".format(get_base_url(), self.organization, self.workspace, self.build_name),
                       json={'createdAt': "2020-01-02T03:45:56.123+00:00", 'id': 123}, status=200)
+        responses.add(responses.GET, "{}/intake/organizations/{}/workspaces/{}/verification".format(
+            get_base_url(), self.organization, self.workspace), status=200)
+        responses.add(responses.GET, "{}/intake/organizations/{}/workspaces/{}/builds/dummy".format(get_base_url(),
```
Konboi, line 55: Do these codes still need?

Konboi (approved):
LGTM

But I commented a bit.

tests/cli_test_case.py
shibuiwilliam: good point. removed https://github.com/launchableinc/cli/pull/360/commits/0da7016577e28d0c2f5b4378c045cee7daea6285

PR #361, opened 2021-12-21
rm validation

Konboi:
Before this change, the cli requires to run `launchable record build` before executing `launchable record session`.
But, it is unintended behavior so I removed validation logic.


kohsuke (approved):
I don't have the context of the change but I can vouch that the change does what you describe!

PR #364, opened 2022-01-06
Add --json option to cucumber plugin

Konboi:
I supported a JSON report format file in the cucumber plugin.

e.g) 
```sh
launchable record tests cucumber --json ./report/**/*.json
```

There are some different formats between XML formant and JSON format. So, JSON's report data that cli sends to Launchable is a little different from using XML formats. 
But cucumber will use file names to subsetting so there is no problem.

___ 

If the user set not JSON file, cli shows an error message like this
```
$ launchable record test cucumber --json ./report/TEST-features-foo-bar.xml
...

    raise Exception(exceptions)
Exception: [Exception('Failed to process a report file: ./report/TEST-features-foo-bar.xml', Exception("Can't read JSON format report file ./report/TEST-features-foo-bar.xml. Make sure to confirm report file."))]
```

And if the user set empty report file, cli shows warning message

```sh
$ launchable record test cucumber --json ./empty.json
Can't find test reports from ./empty.json. Make sure to confirm report file.
Looks like tests didn't run? If not, make sure the right files/directories are passed
```

### DIFF
--- a/launchable/test_runners/cucumber.py
+++ b/launchable/test_runners/cucumber.py
@@ -133,13 +133,13 @@ def _record_tests_from_json(report_file: str) -> Generator[CaseEventType, None, 
     for d in data:
         file_name = d.get("uri", "")
         class_name = d.get("name", "")
-        for e in d.get("elements", []):
-            test_case = e.get("name", "")
+        for element in d.get("elements", []):
+            test_case = element.get("name", "")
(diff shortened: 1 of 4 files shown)


yoshiori (approved):
Great Job!

Konboi:
I'm going to write a document after.

PR #366, opened 2022-01-19
Do not run git command when --commits is specified

draftcode:
* If --no-commit-collection is *not* specified, it runs git commands to
  detect HEAD and submodules HEADs. It sends the commit data to the
  server.
* If --no-commit-collection is specified, but --commits is *not*
  specified, it runs git commands to detect HEAD and submodule HEADs. It
  won't send the commit history. This git call is necessary in order to
  get the commits that are associated with the build.
* If --no-commit-collection is specified and --commits is specified, it
  won't run git commands at all.

PR #367, opened 2022-01-19
Check if commits item is a str

draftcode:
Found in https://github.com/launchableinc/cli/pull/366. It should check
if the item is a str instead of commits itself is a str. Added a test,
but it indicates the cli is calling git command even if --commit is
specified, this should be fixed in the forementioned PR.


shibuiwilliam (approved):
lgmt

PR #369, opened 2022-01-21
Added a debug switch to just report test paths.

kohsuke:
I was inspecting how two different test results compare wrt Launchable,
and it turns out a handy way to do so is to have the CLI print out test
paths, and compare them.

This seems like a reasonable diagnostic capability, so I'm keeping it.


Konboi (approved):
LGTM

PR #371, opened 2022-01-24
e2e test

Konboi:
I added e2e tests and the envs for tests.  When merged to master, it only runs. (But I confirmed it runs before merging ref: https://github.com/launchableinc/cli/actions/runs/1743174489)
image

I tried to add Windows environment but there are some problems e.g) different path format, bazel doesn't work well etc.
So I gave up adding Windows... 😢

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -115,4 +115,4 @@ jobs:
         if: always()
         env:
           LAUNCHABLE_TOKEN: ${{ secrets.LAUNCHABLE_TOKEN_GRADLE }}
-        working-directory: examples/go
+        working-directory: examples/gradle
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,118 @@
+name: e2e
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+
+env:
+  LAUNCHABLE_DEBUG: 1
+  LAUNCHABLE_REPORT_ERROR: 1
+
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
+        with:
+          repository: launchableinc/examples
+          path: examples
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,118 @@
...
+        working-directory: examples/gradle
+      - name: "gradle: Run test"
+        run: ./gradlew test $(cat launchable-subset.txt)
+        working-directory: examples/gradle
+      - name: "gradle: Record test results"
+        run: launchable record tests gradle build/test-results/test
+        if: always()
+        env:
+          LAUNCHABLE_TOKEN: ${{ secrets.LAUNCHABLE_TOKEN_GRADLE }}
+        working-directory: examples/go
```
yoshiori, line 118: Please add new liine

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,118 @@
+name: e2e
+
+on:
+  push:
```
takanabe, line 4: Do you expect both push and pull-request events?

Konboi: No, I don't. I expect only pushed the main branch. In short, I want to run e2e tests after merging to the main branch. 

I think we test `python-package` tests so don't need to run e2e test every pull-request.

takanabe: To be clear, this Actions will be triggered when

* We directly push commits to the main branch
* We merge PRs to the main branch

If we only want to run Actions when we merge PRs to the main branch, we need to specify on.pull_request event with if: github.event.pull_request.merged == true 📝

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,118 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .
+      # bazel
+      - name: Install Bazelisk
+        run: |
+          go get github.com/bazelbuild/bazelisk
```
takanabe, line 43: Because we use Go 1.17 go install might be better for the binary installation
```suggestion
          go install github.com/bazelbuild/bazelisk
```

takanabe (approved):
LGTM with nits

.github/workflows/e2e.yml
Konboi: > with if: github.event.pull_request.merged == true

TIL, Thank you!
But if someone pushes to the main branch directly, we should check it. So, I'll keep it.

PR #375, opened 2022-02-02
Fix commit collect error handling

Konboi:
Before

image

After

image

### DIFF
--- a/launchable/commands/record/commit.py
+++ b/launchable/commands/record/commit.py
@@ -3,6 +3,8 @@ import click
 import sys
 from urllib.parse import urlparse
 
+from launchable.utils import subprocess
+
 from ...utils.env_keys import REPORT_ERROR_KEY
 from ...utils.http_client import get_base_url
 from ...utils.java import get_java_command
@@ -40,6 +42,8 @@ def commit(ctx, source, executable, max_days, scrub_pii):
         if os.getenv(REPORT_ERROR_KEY):
             raise e
         else:
+            click.echo(click.style("Can't get commit history from `{}`. Do you run command root of git-controlled directory? If not, please set a directory use by --source option.".format(
+                os.path.abspath(source)), fg='yellow'), err=True)
             print(e)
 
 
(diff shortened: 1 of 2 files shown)


launchable/commands/record/commit.py
```diff
@@ -54,14 +58,14 @@ def exec_jar(source, max_days, dry_run):
...
 
-    os.system(
-        "{} {} -jar \"{}\" ingest:commit -endpoint {} -max-days {} {} {} {} {}"
-        .format(java, proxy_option, jar_file_path,
-                "{}/intake/".format(base_url), max_days,
-                "-audit" if Logger().logger.isEnabledFor(LOG_LEVEL_AUDIT) else "",
-                "-scrub-pii",
-                "-dry-run" if dry_run else "",
-                source))
+    subprocess.check_output("{} {} -jar \"{}\" ingest:commit -endpoint {} -max-days {} {} {} {} {}"
```
Konboi, line 61: os.system method doesn't throw an Exception when the command fails.
So I change to use subprocess.check_output instead of os.system

shibuiwilliam, line 61: [nits]
knowing this is not the aim of this PR, you can write python string formatter to have placement with variable within strings to specify which placement corresponds to which value. This is useful for better readability in case there is a lot of placement than positional placement.

```python
subprocess.check_output("{java} {proxy_option} -jar \"{jar_file_path}\" ingest:commit -endpoint {base_url}/{intake}/ -max-days {max_days} {audit} {scrub_pii} {dry_run} {source}"
    .format(
        java=java, 
        proxy_option=proxy_option, 
        jar_file_path=jar_file_path,
        base_url=base_url,
        intake="intake", 
        max_days=max_days,
        audit="-audit" if Logger().logger.isEnabledFor(LOG_LEVEL_AUDIT) else "",
        scrub_pii="-scrub-pii",
        dry_run="-dry-run" if dry_run else "",
        source=source,
    ), 
    shell=True)
```

shibuiwilliam (approved):
lgtm
one comment, just a nits.

launchable/commands/record/commit.py
Konboi: TIL, I'll fix it

PR #376, opened 2022-02-04
Add --json option to pytest runner

Konboi:
Although, we want to get a test path of an extended class, when we write tests used by a base and an extended class, the path produced by --junit-xml option becomes the path of the base class file.

e.g) filename of test_example become base.py

```python
# base.py
class Base(object):

# extend.py
class Extend(Base):
  def test_example:
    pass
```

So, we recommend using [pytest-dev/pytest-reportlog](https://github.com/pytest-dev/pytest-reportlog). 
the reporter can produce a correct file path used by [pytest-dev/pytest-reportlog](https://github.com/pytest-dev/pytest-reportlog). 

And I added `--json` option to be able to load JSON report files.


launchable/test_runners/pytest.py
```diff
@@ -92,4 +104,74 @@ def _pytest_formatter(test_path):
...
+        for t in glob.iglob(root, recursive=True):
+            match = True
+            if os.path.isdir(t):
+                client.scan(t, "*.{}".format(ext))
+            else:
+                client.report(t)
+
+        if not match:
+            click.echo("No matches found: {}".format(root), err=True)
+            return
```
shibuiwilliam, line 125: [question]
is it right to return, instead of continue, here?
I mean, if glob.iglob has no iteration but source_roots has other root, don't you have to iterate over rest of the source_roots?

launchable/test_runners/pytest.py
```diff
@@ -92,4 +104,74 @@ def _pytest_formatter(test_path):
...
+
+    if json_report:
+        client.parse_func = PytestJSONReportParser(client).parse_func
+
+    client.run()
+
+
+"""
+If you want to use --json option, please install pytest-dev/pytest-reportlog. (https://github.com/pytest-dev/pytest-reportlog)
```
shibuiwilliam, line 134: [nits]
this comment better be written somewhere in README than within code.

launchable/test_runners/pytest.py
Konboi: Yes right, I copied a logic from here and I changed only to check file extension https://github.com/launchableinc/cli/blob/30d6a238c5a7d0b3996a8dda9284eabd6013f251/launchable/test_runners/launchable.py#L117-L124

launchable/test_runners/pytest.py
Konboi: Exactlly true,

Konboi: added: https://github.com/launchableinc/cli/pull/376/commits/89552787fb794806944deeb9ebedbc62d9d8a634

launchable/test_runners/pytest.py
```diff
@@ -92,4 +104,74 @@ def _pytest_formatter(test_path):
...
+```
+"""
+
+
+class PytestJSONReportParser:
+    def __init__(self, client):
+        self.client = client
+
+    def parse_func(self, report_file: str) -> Generator[CaseEventType, None, None]:
+        with open(report_file, 'r') as json_file:
```
shibuiwilliam, line 151: [nits]
you can load json file 

```python
with open(report_file, 'r') as json_file:
    d = json.load(json_file)
```

`d` will be a dict from json file

shibuiwilliam: wait, do you like to validate each line of json file?
https://github.com/launchableinc/cli/pull/376/files#diff-f30bf261afdbc96299251de2735d7eea4b7456235c83a70cd3352140dd66bc5bR154

tests/test_runners/test_pytest.py
```diff
@@ -49,6 +51,24 @@ def test_record_test_pytest(self):
...
+
+        self.assertEqual(result.exit_code, 0)
+        payload = json.loads(gzip.decompress(
+            responses.calls[1].request.body).decode())
+        expected = self.load_json_from_file(self.result_file_path)
+
+        for e in payload["events"]:
+            del e["created_at"]
+        for e in expected["events"]:
+            del e["created_at"]
```
shibuiwilliam, line 68: [nits]
Better to use `e.pop("created_at", "")` than `del e["created_at"]` to remove a key from dict.
If there is no "created_at" in the dict, `del` will raise KeyError, while `pop` with secondary parameter will not raise any Error.

shibuiwilliam (commented):
commented.

launchable/test_runners/pytest.py
shibuiwilliam: great!

launchable/test_runners/pytest.py
shibuiwilliam: So it is the correct status to `return` than `continue`?

Konboi: Yes, I wanted to say I want to keep back compatibility, for now

launchable/test_runners/pytest.py
Konboi: Yes, right. pytest-dev/pytest-reportlog produces this format https://en.wikipedia.org/wiki/JSON_streaming

shibuiwilliam (approved):

awilkes:
Thanks @Konboi! Also, thank you for updating the documentation at the same time 👍

PR #377, opened 2022-02-16
fix subset in dict

shibuiwilliam:
# What
- escape when subset not in summary.

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -294,9 +294,9 @@ def subset(
 
             header = ["", "Candidates",
                       "Estimated duration (%)", "Estimated duration (min)"]
-            if "subset" not in summary.keys():
+            if "subset" not in summary.keys() or "rest" not in summary.keys():
                 click.echo(click.style(
-                    "Error: no subset found.", 'yellow'), err=True)
+                    "Error: no subset or rest found.", 'yellow'), err=True)
                 return
             rows = [
                 [
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -290,8 +290,6 @@ def subset(
                 self.output_handler(output, rests)
 
             if "subset" not in summary.keys() or "rest" not in summary.keys():
-                click.echo(click.style(
-                    "Error: no subset or rest found.", 'yellow'), err=True)
                 return
 
             build_name, test_session_id = parse_session(session_id)
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -289,15 +289,16 @@ def subset(
             else:
                 self.output_handler(output, rests)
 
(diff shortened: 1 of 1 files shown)


draftcode (approved):

launchable/commands/subset.py
```diff
@@ -294,14 +294,32 @@ def run(self):
 
             header = ["", "Candidates",
                       "Estimated duration (%)", "Estimated duration (min)"]
+            if "subset" not in summary.keys() or "rest" not in summary.keys():
+                click.echo(click.style(
+                    "Error: no subset or rest found.", 'yellow'), err=True)
```
Konboi, line 299: I think we don't have to show this message because already print an error message at https://github.com/launchableinc/cli/pull/377/files#diff-4c62e35fb4e3b2b6bff8593107cdddd1fd47eff90293edc9d838c9b2f3cff94aR279

What do you think?

Konboi (approved):
I commented a bit but almost OK

launchable/commands/subset.py
```diff
@@ -294,14 +294,32 @@ def run(self):
 
             header = ["", "Candidates",
                       "Estimated duration (%)", "Estimated duration (min)"]
+            if "subset" not in summary.keys() or "rest" not in summary.keys():
```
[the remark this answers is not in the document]
shibuiwilliam: fixed 
https://github.com/launchableinc/cli/pull/377/commits/e9200f167ec9264d27d47e181ae94b308d60142f

launchable/commands/subset.py
shibuiwilliam: They are different. The exception is checking error for lanchable client, while this if clause is validating keys in exist in summary dictionary.

Konboi: Sorry, I want to confirm just in case.
Is there any case that even when Launchable returns status code 200, then the subset or the rest field doesn't exist?

shibuiwilliam: removed comment.
[eaa4f74](https://github.com/launchableinc/cli/pull/377/commits/eaa4f748ff8868e40e6d3400878391ba6a96708f)

launchable/commands/subset.py
```diff
@@ -289,19 +289,36 @@ def run(self):
             else:
                 self.output_handler(output, rests)
 
+            if "subset" not in summary.keys() or "rest" not in summary.keys():
```
Konboi, line 292:
```suggestion
            # When Launchable returns an error, the cli skips showing summary report 
            if "subset" not in summary.keys() or "rest" not in summary.keys():
```

Konboi: Sorry, just one last thing, I think need context.

Konboi (approved):

PR #378, opened 2022-02-18
fix dict to get key by get

shibuiwilliam:
# What
- use `get` to get key from dict

### DIFF
--- a/launchable/commands/inspect/subset.py
+++ b/launchable/commands/inspect/subset.py
@@ -48,13 +48,22 @@ def subset(subset_id):
     click.echo(tabulate(rows, header, tablefmt="github"))
 
 
-def convert_row(list: List[Dict], order: int, is_subset: bool):
+def convert_row(data_list: List[Dict], order: int, is_subset: bool):
     """
-    list: testPaths or rest in response to a get subset API
+    data_list: testPaths or rest in response to a get subset API
     order: start number of order
     is_subset: in subset or not
     """
-    return [[order + i, "#".join([path.get("type", "") + "=" + path.get("name", "")
-                                  for path in l.get("testPath", [])]), "✔"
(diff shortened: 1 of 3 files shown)


launchable/commands/inspect/subset.py
```diff
@@ -54,7 +54,7 @@ def convert_row(list: List[Dict], order: int, is_subset: bool):
     order: start number of order
     is_subset: in subset or not
     """
-    return [[order + i, "#".join([path["type"] + "=" + path["name"]
-                                  for path in l["testPath"]]), "✔"
-             if is_subset else "", "{:0.4f}".format(l["duration"] / 1000)]
+    return [[order + i, "#".join([path.get("type", "") + "=" + path.get("name", "")
```
takanabe, line 57: Seems this assumes testPath potentially does not contain type and name. Is it acceptable to create a string like `#=` if the testPaths don't have the keys?

shibuiwilliam: thanks.
fixed to skip if the key does not exist
https://github.com/launchableinc/cli/pull/378/commits/733db94932515a75654e6fb4c8ef3ceb3244e6cb

launchable/commands/inspect/subset.py
```diff
@@ -48,13 +48,22 @@ def subset(subset_id):
     click.echo(tabulate(rows, header, tablefmt="github"))
 
 
-def convert_row(list: List[Dict], order: int, is_subset: bool):
+def convert_row(data_list: List[Dict], order: int, is_subset: bool):
     """
-    list: testPaths or rest in response to a get subset API
+    data_list: testPaths or rest in response to a get subset API
```
shibuiwilliam, line 53: `list` is reserved word so do not use `list` as a variable

takanabe (approved):

PR #380, opened 2022-02-21
rescue file::testcase case

okajimad:
In python, we can write functions directly under the file.
In this case, the behavior of parsing TestPath becomes strange.

### DIFF
--- a/launchable/test_runners/pytest.py
+++ b/launchable/test_runners/pytest.py
@@ -72,7 +72,7 @@ def _parse_pytest_nodeid(nodeid: str) -> TestPath:
             {"type": "file", "name": os.path.normpath(file)},
             {"type": "testcase", "name": testcase}
         ]
-    elif len(data)==3:
+    elif len(data) == 3:
         # tests/test_mod.py::TestClass::test_can_print_aaa -> tests.test_mod.TestClass
         return [
             {"type": "file", "name": os.path.normpath(file)},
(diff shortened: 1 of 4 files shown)


Konboi (commented):
Please fix the test and add a test case to test_pytest https://github.com/launchableinc/cli/blob/main/tests/test_runners/test_pytest.py#L81

launchable/test_runners/pytest.py
```diff
@@ -62,16 +62,26 @@ def _add_testpaths(lines: List[str]):
...
-    return [
-        {"type": "file", "name": os.path.normpath(data[0])},
-        {"type": "class", "name": class_name},
-        {"type": "testcase", "name": data[-1]},
-    ]
+        return [
+            {"type": "file", "name": os.path.normpath(data[0])},
+            {"type": "class", "name": class_name},
+            {"type": "testcase", "name": testcase},
+        ]
```
Konboi, line 84: I think the file variable is already defined so we should use it below to avoid confuse.

```diff
 def _parse_pytest_nodeid(nodeid: str) -> TestPath:
     data = nodeid.split("::")
-    file = data[0]
-    class_name = _path_to_class_name(file)
-
+    file = os.path.normpath(data[0])
+    class_name = _path_to_class_name(data[0])
+
     # file name only
     if len(data) == 1:
         return [
-            {"type": "file", "name": os.path.normpath(data[0])},
+            {"type": "file", "name": file},
             {"type": "class", "name": class_name},
         ]
     # file + testcase, or file + class + testcase
     else:
         testcase = data[-1]
-        if len(data)==3:
+        if len(data) == 3:
             class_name += "." + data[1]

         return [
-            {"type": "file", "name": os.path.normpath(data[0])},
+            {"type": "file", "name": file},
             {"type": "class", "name": class_name},
             {"type": "testcase", "name": testcase},
         ]
```

launchable/test_runners/pytest.py
```diff
@@ -84,22 +94,24 @@ def _path_to_class_name(path):
 def _pytest_formatter(test_path):
     for path in test_path:
         t = path['type']
+        n = path['name']
```
Konboi, line 97: Could you change to use `.get()` method like this PR https://github.com/launchableinc/cli/pull/378?

Konboi (commented):
Thank you for your PR

I commented please check it.

Konboi (approved):
LGTM

PR #381, opened 2022-02-22
Add a root exception handler

draftcode:
By using standalone_mode
(https://click.palletsprojects.com/en/8.0.x/exceptions/), we can get
exceptions from Click without it processing them. This allows us to
catch them and exit with an arbitrary exit code.

This changes the default behavior to exit the CLI always with the exit
code 0, which is a breaking change.

---

While this appears to be a safe change on the first sight, this can
silently disrupt some integrations. Imagine the scenario that, an
integration script is written in a way that if launchable command fails,
it falls back to a full run. With this change, that fallback stops
working, and it can end up reading an empty test subset, and not
executing any test in the worst case.

### DIFF
--- a/launchable/__main__.py
+++ b/launchable/__main__.py
@@ -1,6 +1,10 @@
 import importlib
 import importlib.util
 import logging
+import os
+import sys
+import textwrap
+import traceback
 from glob import glob
 from os.path import basename, dirname, join
 
@@ -79,5 +83,27 @@ main.add_command(split_subset)
 main.add_command(verify)
 main.add_command(inspect)
 
+
+def cli_main():
+    if os.environ.get("LAUNCHABLE_RETURNS_ERROR") == '1':
+        main.main()
+    else:
+        try:
+            # Ignore the exit code.
+            main.main(standalone_mode=False)
+        except BaseException:
+            # Click handles ClickException etc. specially. We don't add that
+            # handling because in the production mode those errors should be
+            # handled exceptionally.
+            click.utils.echo(textwrap.dedent("""
+            ====================================================================
+            Exception raised. But since LAUNCHABLE_RETURNS_ERROR is not set,
+            exit with the exit code 0.
+            ====================================================================
+            """), file=sys.stderr)
+            traceback.print_exc()
+            sys.exit(0)
+
+
 if __name__ == '__main__':
-    main()
+    cli_main()
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ install_requires =
 python_requires = >=3.5
 
 [options.entry_points]
-console_scripts = launchable = launchable.__main__:main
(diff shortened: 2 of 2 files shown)


brianjmurrell:
Just to add context... We've had full CI run failures due to something like `unable to post to https://api.mercury.launchableinc.com/intake/organizations/.../test_sessions`.  So while it would be great if we could depend on every service being up 100% of the time, that is not reality.  In such cases though, those kinds of failures should not tank a whole test run.  They should be silent.

And to be perfectly clear... At this point in time, we are not even requesting subsets, etc.  We are simply just posting test results.  So while I could agree that there might be some debate over what my code should do if requesting a subset has some kind of failure such that I don't get a subset (i.e. my code should probably handle that and just do a full test run, perhaps), when all that I am doing is posting test runs, there is no Launchable infrastructure operation that should surface up a failure such that it tanks my test runs (assuming I am not just ignoring all Launchable-call exit codes).  IMHO.

IOW, if one is just posting test runs, and not requesting subsets, there really should be zero failures from Launchable.  Any infrastructure issues (either establishing the session, or closing it out with test results) should simply result in not posting the results of such a test run.  Hopefully, losing a single test run is insignificant, and if the numbers of lost test runs is more than insigificant, then there are bigger fishes to be fried on Launchable's end WRT to infrastructure.

awilkes:
Thanks for the additional color @brianjmurrell.

@draftcode I want to +1 @kohsuke's suggestion, since it limits the blast radius a bit:
> I'd be a little more comfortable if this is a bit more targeted. For example, `record test` command having `except BaseException` seems like a reasonable defense in depth.

(I would probably add `record build` and `record session` as well.)

Konboi:
Thank you for your comment @brianjmurrell , 

I think the cli should not disrupt the user's CI pipeline by the CLI error or Launchable problem too. So, We are fixing not to do it. If it's not enough, please let me know more details.

Also, the cli should inform the users, if the users set miss-configuration or set miss parameters. The cli stopping the pipeline is correct, I think.

It changes original behavior. So, I think this change is a little too far.  
@brianjmurrell Is it difficult to wrap the cli not to disrupt when the cli fails in your configuration?

If we merge this change, we should change to opt-in future.

brianjmurrell:
> I think the cli should not disrupt the user's CI pipeline by the CLI error or Launchable problem too. So, We are fixing not to do it. 

That's great news.

> Also, the cli should inform the users, if the users set miss-configuration or set miss parameters. The cli stopping the pipeline is correct, I think.

I agree.

> @brianjmurrell Is it difficult to wrap the cli not to disrupt when the cli fails in your configuration?

Well, at minimum, it's a distraction from other pressing activities that I just don't have the time to properly engineer and test at the moment.  But also, it does not provide the behaviour you suggest above to not disrupt the pipeline for issues out of my control but to stop the pipeline for configuration issues, etc.

PR #387, opened 2022-03-08
Support multiple XML files

kohsuke:
Some folks invoke `nunit --explore` separately for different DLLs yet we still would like to make one subset call to select meaningful tests across all such DLLs.


Konboi (approved):
LGTM

takanabe (approved):

PR #389, opened 2022-03-09
Fix/vulnerability

shibuiwilliam:
# What
- fix os command vulnerability

PR #393, opened 2022-03-10
fix vulnerability subprocess

shibuiwilliam:
# What
- fix os command vulnerability
- trigger e2e test github actions on PR to main branch

# Ref
- https://github.com/launchableinc/cli/pull/389
- https://github.com/launchableinc/cli/pull/390
- https://github.com/launchableinc/cli/pull/391
- https://github.com/launchableinc/cli/pull/392

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -3,6 +3,8 @@ name: e2e
 on:
   push:
     branches: [main]
+  pull_request:
+    branches: [ main ]
   workflow_dispatch:
 
 
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -3,8 +3,6 @@ name: e2e
 on:
   push:
     branches: [main]
-  pull_request:
-    branches: [ main ]
   workflow_dispatch:
 
 
--- a/launchable/commands/record/commit.py
+++ b/launchable/commands/record/commit.py
@@ -55,6 +55,8 @@ def exec_jar(source, max_days, dry_run):
 
     base_url = get_base_url()
 
+    # using subprocess.check_out with shell=False and a list of command to prevent vulnerability
+    # https://knowledge-base.secureflag.com/vulnerabilities/code_injection/os_command_injection_python.html
     command = [java]
     https_proxy = os.getenv("HTTPS_PROXY")
     proxy_option = _build_proxy_option(https_proxy) if https_proxy else ""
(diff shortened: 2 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -3,6 +3,8 @@ name: e2e
 on:
   push:
     branches: [main]
+  pull_request:
+    branches: [ main ]
```
Konboi, line 7: We usually make PR to main. In short, e2e test will run each branch.
But it's too much, I think.

How about run e2e before publish new version cli ?

My proposal is 
1. e2e core logic moves to common file use by composite feature
  - https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/
2. e2e test and python-publish use it.

launchable/commands/record/commit.py
```diff
@@ -55,21 +55,35 @@ def exec_jar(source, max_days, dry_run):
...
-                                max_days=max_days,
-                                audit="-audit" if Logger().logger.isEnabledFor(LOG_LEVEL_AUDIT) else "",
-                                scrub_pli="-scrub-pii",
-                                dry_run="-dry-run" if dry_run else "",
-                                source=source),
-                            shell=True)
+    if proxy_option != "":
+        command.append(proxy_option)
+
+    command.extend(
```
Konboi, line 64: How about adding a comment that why we use extend and don't use system?

shibuiwilliam: what do you mean by system?

.github/workflows/e2e.yml
shibuiwilliam: removed the change since making a composition looks out of scope for the purpose of the PR.
https://github.com/launchableinc/cli/pull/393/commits/d5ec7704fce3b154eb11bc3bb3a46c09a4cffc92

launchable/commands/record/commit.py
shibuiwilliam: added comment.
https://github.com/launchableinc/cli/pull/393/commits/d5ec7704fce3b154eb11bc3bb3a46c09a4cffc92

Konboi: I wanted to say python method's `extend` and `os. system.`

shibuiwilliam: `extend` is just adding a list to another list, and it is different topic of `os.system`.

E.g.
```python
a_list = [1, 2, 3]
b_list = [4, 5, 6]
a_list.extend(b_list)
print(a_list)
# [1, 2, 3, 4, 5, 6]
```

Konboi: Sounds good
I thought we should comment on why we build a command step by step, why not write one-liner. And your comment meets it. Thank you

Konboi (approved):
LGTM

PR #395, opened 2022-03-23
change pypi publish to API token auth

shibuiwilliam:
# What
- change PyPI authentication to API token
  - use Test PyPI for test publish and PyPI for official one
- update Github actions for PyPI publish
  - github action: https://github.com/marketplace/actions/pypi-publish
  - usage with build and publish: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -29,7 +29,7 @@ jobs:
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         user: __token__
-        password: ${{ secrets.PYPI_API_TOKEN }}
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
         repository_url: https://test.pypi.org/legacy/
     - name: Publish to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -22,13 +22,20 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install setuptools setuptools_scm wheel twine
-    - name: Build and publish
-      env:
-        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
-        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+    - name: Build
       run: |
         python setup.py sdist bdist_wheel
-        twine upload dist/*
+    - name: Test publish to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+    - name: Publish to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_API_TOKEN }}
     - name: Actions for Discord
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -1,3 +1,12 @@
 from setuptools import setup
 
-setup()
+
+def local_scheme(version) -> str:
```
shibuiwilliam, line 4: add this change to avoid versioning like `1.38.11.dev13+g9aabe7c` which is invalid version name for PyPI
https://github.com/pypa/setuptools_scm/issues/342

.github/workflows/test-python-publish.yml
```diff
@@ -0,0 +1,46 @@
+# This workflows will upload a Python Package to Test PyPI
+# For more information see: https://github.com/marketplace/actions/pypi-publish
+
+name: TEST Upload Python Package
```
shibuiwilliam, line 4: add test publish to assure the library can be released to PyPI

.github/workflows/python-publish.yml
```diff
@@ -14,21 +14,41 @@ jobs:
...
-        pip install setuptools setuptools_scm wheel twine
-    - name: Build and publish
-      env:
-        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
-        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-      run: |
-        python setup.py sdist bdist_wheel
-        twine upload dist/*
+    # build and publish package using GitHub Actions workflow
+    # https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
```
shibuiwilliam, line 24: refer for this change
https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

.github/workflows/test-python-publish.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
Konboi, line 10: If there are some branches, some test versions will be released, right?

If so, it's enough when the codes will be merged to main, I think. What do you think?

.github/workflows/python-publish.yml
```diff
@@ -14,21 +14,41 @@ jobs:
...
+      with:
+        user: __token__
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+    # actual publish
+    - name: Publish to PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
Konboi, line 51: I confirmed these secrets key were set.

.github/workflows/test-python-publish.yml
shibuiwilliam: Purpose of this is to test if the package can be built and pushed to test pypi before running publish job.

> If so, it's enough when the codes will be merged to main, I think. What do you think?
Probably you mean running this job only for `push: branches: [main]` is enough, correct?

Konboi: Got it.
I don't have a strong opinion about it and we don't use test package now.
 so it looks problem.

Konboi (approved):
Thank you. Great work!!

PR #398, opened 2022-03-30
If test case name is not specified, we assume execution unit is file.

okajimad:


Konboi:
Can you add test case?

Konboi:
ref: #416

PR #400, opened 2022-04-05
Support ignore new files in subset

Konboi:
In Launchable, it takes time until the execution time of a newly added file is estimated
But since they are local, you may want to ignore those files when you run the test

So I added an option `--ignore-new-tests` to support this


docs/resources/cli-reference.md
```diff
@@ -175,6 +175,7 @@ launchable subset [OPTIONS] TESTRUNNER ...
...
+| `--ignore-new-tests`      | Ignore that were added recently. NOTICE: this option will ignore tests that you added just now, too.                                                                                                                                                                                                                                                                                                                                                                                                                                      | No                                                        |
```
shibuiwilliam, line 178: [grammer]
`Ignore tests that were added recently. NOTICE: this option will ignore tests that you added just now as well.`

launchable/commands/subset.py
```diff
@@ -93,6 +93,12 @@
     """,
     is_flag=True
 )
+@click.option(
+    "--ignore-new-tests",
+    "ignore_new_tests",
+    help='Ignore tests that were added recently.\n\nNOTICE: this option will ignore tests that you added just now, too',
```
shibuiwilliam, line 99: https://github.com/launchableinc/cli/pull/400/files#r842393927

shibuiwilliam (approved):
just a grammatical error

docs/resources/cli-reference.md
Konboi: Ops... Thank you!

draftcode (approved):

PR #404, opened 2022-04-13
For consistency, promote key=value syntax

kohsuke:
While trying to explain a certain CLI use to a customer, I noticed that the `--commit` option shows the odd form of `key:value` syntax when everything else, such as `--source` or `--flavor` shows `key=value` syntax.

Turns out the code supports all of them, so node code change needed, but I thought the doc should promote the consistent syntax as our preferred one.


awilkes (approved):
Good catch

Konboi (approved):

PR #416, opened 2022-06-07
Fix Pytest profile failed with "local variable 'case' referenced before assignment"

ninjinkun:
When the CLI working with `pytest-launchable`, the Pytest profile on the CLI failed with the message "local variable 'case' referenced before assignment". So, I add a few lines to define the variables at the top of the function.

### DIFF
--- a/launchable/test_runners/pytest.py
+++ b/launchable/test_runners/pytest.py
@@ -2,7 +2,7 @@ import glob
 import json
 import pathlib
 from platform import node
-from typing import Generator, List
+from typing import Generator, List, Optional
 from os.path import *
 import subprocess
 import click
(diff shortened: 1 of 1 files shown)


ninjinkun:
Oh, man… 🤦‍♂️
```
launchable/test_runners/pytest.py:96: error: Variable annotation syntax is only supported in Python 3.6 and greater
Found 1 error in 1 file (errors prevented further checking)
```

Konboi (commented):
Is it related PR?
https://github.com/launchableinc/cli/pull/398

Konboi:
Is it difficult to add test codes to confirm fixing it?

PR #422, opened 2022-06-14
Change an option name in the doc

ninjinkun:
At https://github.com/launchableinc/pytest-launchable/pull/15, I renamed the option names to hyphenated. That is why I modified the option name in doc.


ninjinkun:
@awilkes just FYI

PR #423, opened 2022-06-16
Support slack notification

Konboi:
I added a beta feature.
If we want to use feature, we need to add `LAUNCHABLE_SLACK_NOTIFICATION` env value when we execute the cli

### DIFF
--- a/launchable/commands/record/tests.py
+++ b/launchable/commands/record/tests.py
@@ -324,7 +324,10 @@ def tests(
                     send(p)
                     exceptions.extend(es)
 
-                res = client.request("patch", "{}/close".format(session_id))
+                metadata = get_env_values(client)
+
+                res = client.request(
+                    "patch", "{}/close".format(session_id), payload={"metadata": metadata})
                 res.raise_for_status()
 
                 if len(exceptions) > 0:
@@ -442,3 +445,19 @@ def get_session_and_record_start_at_from_subsetting_id(subsetting_id: str, clien
         "session": "builds/{}/test_sessions/{}".format(build_number, test_session_id),
         "start_at": parse_launchable_timeformat(created_at)
     }
+
+
+def get_env_values(client: LaunchableClient):
+    sub_path = "slack/notification/key/list"
+    res = client.request("get", sub_path=sub_path)
+
+    metadata = {}
+    if res.status_code != 200:
+        return metadata
+
+    keys = res.json()["keys"]
+    for key in keys:
+        val = os.environ.get(key, "")
+        metadata[key] = val
+
+    return metadata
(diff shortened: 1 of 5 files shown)


launchable/commands/record/tests.py
```diff
@@ -442,3 +449,19 @@ def get_session_and_record_start_at_from_subsetting_id(subsetting_id: str, clien
...
+    sub_path = "slack/notification/key/list"
+    res = client.request("get", sub_path=sub_path)
+
+    metadata = {}  # type: Dict[str, str]
+    if res.status_code != 200:
+        return metadata
+
+    keys = res.json()["keys"]
+    for key in keys:
+        val = os.environ.get(key, "")
```
shibuiwilliam, line 464: [nits]
You can write `os.getenv(key, "")` here. Just FYI.

launchable/commands/record/tests.py
```diff
@@ -442,3 +449,19 @@ def get_session_and_record_start_at_from_subsetting_id(subsetting_id: str, clien
...
+
+def get_env_values(client: LaunchableClient) -> Dict[str, str]:
+    sub_path = "slack/notification/key/list"
+    res = client.request("get", sub_path=sub_path)
+
+    metadata = {}  # type: Dict[str, str]
+    if res.status_code != 200:
+        return metadata
+
+    keys = res.json()["keys"]
```
shibuiwilliam, line 462: [must]
This would raise error if there is no `keys` key in response json.
Better write like `keys = res.json().get("keys", [])` here.

launchable/commands/record/tests.py
```diff
@@ -442,3 +449,19 @@ def get_session_and_record_start_at_from_subsetting_id(subsetting_id: str, clien
...
         "start_at": parse_launchable_timeformat(created_at)
     }
+
+
+def get_env_values(client: LaunchableClient) -> Dict[str, str]:
+    sub_path = "slack/notification/key/list"
+    res = client.request("get", sub_path=sub_path)
+
+    metadata = {}  # type: Dict[str, str]
+    if res.status_code != 200:
```
shibuiwilliam, line 459: [question]
While you are using `res.raise_for_status()` [here](https://github.com/launchableinc/cli/pull/423/files#diff-a54ea2cd0c54426ceb96ffb93873062a86aaa7d4d4cb1310ab874ed0eeda2e85R335), why are you validating status code != 200 here?

shibuiwilliam:
It seems Code scanning got timed out. Please fix this.

launchable/commands/record/tests.py
Konboi: Good question, I think notifying slack is an optional feature. So I don't want to suspend main feature (record tests & close session) so I don't use raise_for_status when we request to `slack/notification/key/list`

shibuiwilliam: Good.

Konboi:
@shibuiwilliam Please review again when you free

shibuiwilliam (approved):

PR #425, opened 2022-06-22
Honor default test inclusion/exclusion with Maven

kohsuke:
Maven Surefire has a mechanism to not only recursively scan directories to find tests but also include/exclude subset of the files as valid tests.

Emulate this behavior to collectly list up tests that should be selected. Since these are configurable parameters in our CLI as well, but I'm kicking that can down the road until it's needed. Especially given that the long term future of the Maven support would have to be in-process.


Konboi (approved):
LGTM

PR #430, opened 2022-07-04
Add observation mode

Konboi:
Add new option `--observation`.
If the user uses this option, Launchable will show details information for the subset and the record test results.

### DIFF
--- a/launchable/commands/helper.py
+++ b/launchable/commands/helper.py
@@ -3,7 +3,7 @@ from typing import Optional
 from ..utils.session import read_build, read_session
 
 
-def find_or_create_session(context: click.core.Context, session: Optional[str], build_name: Optional[str], flavor=[]) -> Optional[str]:
+def find_or_create_session(context: click.core.Context, session: Optional[str], build_name: Optional[str], flavor=[], evaluation=False) -> Optional[str]:
     """Determine the test session ID to be used.
 
     1. If the user explicitly provides the session id via the `--session` option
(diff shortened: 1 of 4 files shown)


launchable/commands/helper.py
```diff
@@ -3,7 +3,7 @@
 from ..utils.session import read_build, read_session
 
 
-def find_or_create_session(context: click.core.Context, session: Optional[str], build_name: Optional[str], flavor=[]) -> Optional[str]:
+def find_or_create_session(context: click.core.Context, session: Optional[str], build_name: Optional[str], flavor=[], is_observation=False) -> Optional[str]:
```
shibuiwilliam, line 6: [must]
please add type hinting here.
`is_observation:bool=False`

launchable/commands/record/session.py
```diff
@@ -35,8 +35,15 @@
     cls=KeyValueType,
     multiple=True,
 )
+@click.option(
+    "--observation",
+    "is_observation",
+    help='observation',
```
shibuiwilliam, line 41: [question]
Is it obvious for user what `--observation` is used for with help only saying `observation`?

Konboi: Ops, it's still a draft... I'll fix it. Thanks

launchable/commands/record/session.py
```diff
@@ -35,8 +35,15 @@
...
+    is_flag=True,
+    required=False,
```
shibuiwilliam, line 43: [nits]
Since this is a flag that specifying `--observation` means to use observation mode, I think it is redundant to say `required=False`.

launchable/commands/record/session.py
```diff
@@ -35,8 +35,14 @@
     cls=KeyValueType,
     multiple=True,
 )
+@click.option(
+    "--observation",
+    "is_observation",
+    help='enable observation mode',
```
shibuiwilliam, line 41: [nits]
use double quotation instead of single quotation.

shibuiwilliam (approved):
one nits, but not necessary to fix

PR #433, opened 2022-07-12
Fix/quotations

shibuiwilliam:
# What
- Fix single quote to double quote where possible to integrate usage of quotation through codebase.
- This is just refactoring, and no need to hurry.


Konboi:
By the way, where is the mentioned that we should use a double quote instead of a single quote?

Konboi (approved):
LGTM but …
I want to keep the git history without these changes for easy to access past contexts, especially for the `git blame`
Do you have any good ideas?

shibuiwilliam:
> LGTM but … I want to keep the git history without these changes for easy to access past contexts, especially for the `git blame` Do you have any good ideas?

Oh, ok. I close this PR without merging.
I just aimed to organize the code to use the same quotation rather than mixing them.

Konboi:
@shibuiwilliam How about adding `.git-blame-ignore-revs` same time?  It seems to be that we can ignore these commit.

PR #434, opened 2022-07-15
Add extra information for TestNG users

Konboi:
SSIA


kohsuke:
I now propose this change be abandoned in favor of #439. That is, let's make our CLI do the right thing for customers, instead of them using our CLI differently when using TestNG.

PR #435, opened 2022-07-20
Add options for zero input subsetting

Konboi:
I added two options 

- `--output-exclusion-rules`
- `--get-tests-from-previous-full-runs`

### --output-exclusion-rules

When the `--output-exclusion-rules` option enabled, the cli will switch the subset result and the rest result for excluding option

e.g)
```sh
launchable subset --target 30% --output-exclusion-rules src/test/** > exclude.txt
mvn test -Dsurefire.excludesFile=exclude.txt
```

### --get-tests-from-previous-full-runs

The `get-tests-from-previous-full-runs` option is new feature of subset. The user can execute a subset request without subset test lists, after reporting test results sometimes.

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -110,6 +110,12 @@ from tabulate import tabulate
     help="get subset list from previous full tests",
     is_flag=True,
 )
+@click.option(
+    "--output-exclusion-rules",
+    "is_output_exclusion_rules",
+    help="outputs the exclude test list. Switch the subset and rest.",
+    is_flag=True,
+)
 @click.pass_context
 def subset(
     context: click.core.Context,
@@ -126,6 +132,7 @@ def subset(
     ignore_new_tests: bool,
     is_observation: bool,
     is_get_tests_from_previous_full_runs: bool,
+    is_output_exclusion_rules: bool,
 ):
 
     if is_observation and is_get_tests_from_previous_full_runs:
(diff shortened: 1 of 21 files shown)


awilkes:
We should rename `--get-tests-from-previous-full-runs` to `--get-tests-from-previous-sessions`:

1. The original design suggested that customers would mark certain sessions as "full runs" to generate the test list. In contrast, the current design suggests generating the test list by unioning all sessions in a certain timeframe (e.g. the last 2 weeks). Therefore `full` is not necessary.
2. "sessions" is used more consistently than "runs" across the product.

launchable/commands/subset.py
```diff
@@ -120,8 +131,15 @@ def subset(
     no_base_path_inference: bool,
     ignore_new_tests: bool,
     is_observation: bool,
+    is_get_tests_from_previous_full_runs: bool,
+    is_output_exclusion_rules: bool,
 ):
 
+    if is_observation and is_get_tests_from_previous_full_runs:
+        click.echo(click.style(
+            "Can not use --observation and --get-tests-from-previous-full-runs options at the same time", fg="red"), err=True)
```
shibuiwilliam, line 140: [must]
`Can not` -> `Cannot`

shibuiwilliam (approved):
only one comment.

Konboi:
@awilkes I totally agree with your suggestion, I’ll fix it.
Thank you!

PR #436, opened 2022-07-21
Capture job links

ninjinkun:
The CLI captures job links from CI providers in the `test sessions` command to show the links on the web app. It supports the following providers:

* Jenkins
* Github Action
* CircleCI 
  * It doesn't exist in the original plan, but I added it because it's easy to add.

### DIFF
--- a/launchable/commands/record/session.py
+++ b/launchable/commands/record/session.py
@@ -1,7 +1,7 @@
 import click
 import os
 import sys
-from typing import Sequence
+from typing import Dict, Mapping, Optional, Sequence
 from http import HTTPStatus
 from ...utils.http_client import LaunchableClient
 from ...utils.env_keys import REPORT_ERROR_KEY
@@ -10,6 +10,14 @@ from ...utils.click import KeyValueType
 from ...utils.logger import Logger, AUDIT_LOG_FORMAT
 
 LAUNCHABLE_SESSION_DIR_KEY = 'LAUNCHABLE_SESSION_DIR'
+JENKINS_URL_KEY = 'JENKINS_URL'
+JENKINS_BUILD_URL_KEY = 'BUILD_URL'
+GITHUB_ACTION_KEY = 'GITHUB_ACTION'
+GITHUB_SERVER_URL_KEY = 'GITHUB_SERVER_URL'
+GITHUB_REPOSITORY_KEY = 'GITHUB_REPOSITORY'
+GITHUB_RUN_ID_KEY = 'GITHUB_RUN_ID'
+CIRCLECI_KEY = 'CIRCLECI'
+CIRCLECI_BUILD_URL_KEY = 'CIRCLE_BUILD_URL'
 
 
 @click.command()
@@ -67,13 +75,18 @@ def session(ctx: click.core.Context, build_name: str, save_session_file: bool, p
         elif isinstance(f, Sequence):
             flavor_dict[f[0]] = f[1]
 
+    link = _capture_link(os.environ)
+    payload = {
+        "flavors": flavor_dict,
+        "observation": is_observation,
+    }
+    if link:
+        payload["link"] = link
+
     client = LaunchableClient(dry_run=ctx.obj.dry_run)
     try:
         sub_path = "builds/{}/test_sessions".format(build_name)
-        res = client.request("post", sub_path, payload={
-                             "flavors": flavor_dict,
-                             "observation": is_observation,
-                             })
+        res = client.request("post", sub_path, payload=payload)
 
         if res.status_code == HTTPStatus.NOT_FOUND:
             click.echo(click.style(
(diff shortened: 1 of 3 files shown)


launchable/commands/record/session.py
```diff
@@ -11,6 +13,15 @@
 
 LAUNCHABLE_SESSION_DIR_KEY = 'LAUNCHABLE_SESSION_DIR'
 
+JENKINS_URL_KEY = 'JENKINS_URL'
+JENKINS_BUILD_URL_KEY = 'BUILD_URL'
+GITHUB_ACTION_KEY = 'GITHUB_ACTION'
```
shibuiwilliam, line 18: [question]
Is it ok we have `GITHUB ACTION` here and `GITHUB ACTIONS` below?
https://github.com/launchableinc/cli/pull/436/files#diff-83169579c4ecf3f02d6c5587f3d60e8727f3799b09f8d97a1f75db4dd2f688bdR6

Official name is `GitHub Actions`.
https://github.com/features/actions

Konboi: It seems to be that `GITHUB_ACTION` and `GITHUB_ACTIONS` both exist. https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

In our case that only confirms a job is running on Github Actions or not. So I think we should use `GITHUB_ACTIONS`

launchable/commands/record/session.py
```diff
@@ -94,3 +110,14 @@ def session(ctx: click.core.Context, build_name: str, save_session_file: bool, p
             raise e
         else:
             click.echo(e, err=True)
+
+
+def _capture_link(env: Mapping[str, str]) -> Optional[Dict[str, str]]:
+    if env.get(JENKINS_URL_KEY):
+        return {"provider": CIProvider.JENKINS.value, "url": env.get(JENKINS_BUILD_URL_KEY, "")}
+    elif env.get(GITHUB_ACTION_KEY):
+        return {"provider": CIProvider.GITHUB_ACTIONS.value, "url": "{}/{}/actions/runs/{}".format(env.get(GITHUB_SERVER_URL_KEY), env.get(GITHUB_REPOSITORY_KEY), env.get(GITHUB_RUN_ID_KEY))}
```
Konboi, line 119: memo: https://github.com/launchableinc/cli/actions/runs/2710758567

tests/commands/record/test_session.py
```diff
@@ -6,9 +7,17 @@
...
+    This test needs to specify `clear=True` in mocking because the test is run on GithubActions. 
+    Otherwise GithubActions will export $GITHUB_* variables at runs.
+    """
 
     @responses.activate
-    @mock.patch.dict(os.environ, {"LAUNCHABLE_TOKEN": CliTestCase.launchable_token})
+    @mock.patch.dict(os.environ, {
+        "LAUNCHABLE_TOKEN": CliTestCase.launchable_token,
+        # it's needed to run the CLI because the line cleared all exsting environment
+        'LANG': 'C.UTF-8',
```
Konboi, line 19: Q: Why do we need LANG env?

launchable/commands/record/session.py
```diff
@@ -11,6 +13,15 @@
...
 LAUNCHABLE_SESSION_DIR_KEY = 'LAUNCHABLE_SESSION_DIR'
 
+JENKINS_URL_KEY = 'JENKINS_URL'
+JENKINS_BUILD_URL_KEY = 'BUILD_URL'
+GITHUB_ACTION_KEY = 'GITHUB_ACTION'
+GITHUB_SERVER_URL_KEY = 'GITHUB_SERVER_URL'
+GITHUB_REPOSITORY_KEY = 'GITHUB_REPOSITORY'
+GITHUB_RUN_ID_KEY = 'GITHUB_RUN_ID'
+CIRCLECI_KEY = 'CIRCLECI'
+CIRCLECI_BUILD_URL_KEY = 'CIRCLE_BUILD_URL'
```
Konboi, line 23: memo: https://circleci.com/docs/env-vars#built-in-environment-variables

tests/commands/record/test_session.py
```diff
@@ -50,3 +65,53 @@ def test_run_session_with_observation(self):
...
+        self.assertEqual(result.exit_code, 0)
+
+        payload = json.loads(responses.calls[0].request.body.decode())
+        self.assert_json_orderless_equal(
+            {"flavors": {}, "observation": True, "link": {"provider": "jenkins", "url": "https://jenkins.example.com/job/launchableinc/job/example/357/"}}, payload)
+
+    @responses.activate
+    @mock.patch.dict(os.environ, {
+        "LAUNCHABLE_TOKEN": CliTestCase.launchable_token,
+        GITHUB_ACTION_KEY: "1",
```
Konboi, line 88: nits:

`GITHUB_ACTIONS` and `CIRCLECI` will be set `true` as a bool value. How about aligning the value type?

tests/commands/record/test_session.py
ninjinkun: Because `CliRunner().invoke(command)` needs the flag. In other cases, the flag is automatically added by the shell. But in this case, `clear=True` removes it. See the run result that is without `LANG=C.UTF-8` https://github.com/launchableinc/cli/runs/7442868706?check_suite_focus=true#step:10:102 .

ninjinkun:
@Konboi I modified the code based on suggestions. Please review it again.

Konboi (approved):
LGTM

PR #437, opened 2022-07-22
Support observation mode output

Konboi:
At first, we planned to receive a merged subset result (subset + rests) from Launchable. However, if we did so, output summary data will be changed.
So I changed to merge at the cli side.

___

I need to edit around the codes that were changed for zero input subsetting mode.
So, I made a branch from zero-input-subsetting mode and make PR to it.

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -292,6 +292,7 @@ def subset(
             summary = {}
             subset_id = ""
             is_brainless = False
+            is_observation = False
 
             if not session_id:
                 # Session ID in --session is missing. It might be caused by Launchable API errors.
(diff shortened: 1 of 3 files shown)


takanabe:
I could not understand what is this for https://github.com/launchableinc/cli/blob/571b00982cd5cab55f22fda87c2d2e14ee4cf1e0/launchable/commands/split_subset.py#L98 but we need to change[ the outputs of the split subset command](https://github.com/launchableinc/cli/blob/571b00982cd5cab55f22fda87c2d2e14ee4cf1e0/launchable/commands/split_subset.py#L78-L102) as well along with the subset/split-subset API response changes with the observation mode.

launchable/commands/subset.py
```diff
@@ -338,6 +339,8 @@ def run(self):
                 _output, _rests = output, rests
                 if is_output_exclusion_rules:
                     _output, _rests = _rests, _output
+                elif is_observation:
```
shibuiwilliam, line 342: [question]
Is it correct that you can only get change for only one of `is_output_exclusion_rules` and `is_observation`?
I mean, even if `is_output_exclusion_rules == True` and `is_observation == True`, you will not go through `elif is_observation` clause since it is `elif` instead of `if`.

Konboi:
Thank you for your comment. 
It’s for not to suspend the customer's CI pipeline. 
If all test candidates are included in the subset result, then the rest will be empty without this logic. The rest part of tests will happen some errors.
https://github.com/launchableinc/cli/pull/437#issuecomment-1193011826

Anyway, I'll add logic this area too.

launchable/commands/subset.py
Konboi: Oh, nice catch. Yes, the user will be able to use both options in one request. So, I'll fix it

Konboi:
I have to make a branch from main instead of zero-input-subsetting. So, I'll close this PR and make new one.

PR #438, opened 2022-07-28
Implement observation mode output

Konboi:
At first, we planned to receive a merged subset result (subset + rests) from Launchable. However, if we did so, output summary data will be changed.
So I changed to merge at the cli side.

I made another PR https://github.com/launchableinc/cli/pull/437 before, however, we need to release faster than zero input subsetting mode. So, I made a new PR.

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -269,6 +269,7 @@ def subset(
             summary = {}
             subset_id = ""
             is_brainless = False
+            is_observation = False
 
             if not session_id:
                 # Session ID in --session is missing. It might be caused by Launchable API errors.
@@ -293,7 +294,8 @@ def subset(
                     rests = res.json()["rest"]
                     subset_id = res.json()["subsettingId"]
                     summary = res.json()["summary"]
-                    is_brainless = res.json()["isBrainless"]
+                    is_brainless = res.json().get("isBrainless", False)
+                    is_observation = res.json().get("observation", False)
 
                 except Exception as e:
                     if os.getenv(REPORT_ERROR_KEY):
@@ -312,7 +314,11 @@ def subset(
             if split:
                 click.echo("subset/{}".format(subset_id))
             else:
-                self.output_handler(output, rests)
+                _output, _rests = output, rests
+                if is_observation:
+                    _output = _output + _rests
+
+                self.output_handler(_output, _rests)
 
             # When Launchable returns an error, the cli skips showing summary report
             if "subset" not in summary.keys() or "rest" not in summary.keys():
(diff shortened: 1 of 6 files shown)


launchable/commands/subset.py
```diff
@@ -293,7 +294,8 @@ def run(self):
                     rests = res.json()["rest"]
                     subset_id = res.json()["subsettingId"]
                     summary = res.json()["summary"]
-                    is_brainless = res.json()["isBrainless"]
+                    is_brainless = res.json().get("isBrainless", False)
+                    is_observation = res.json().get("observation", False)
```
takanabe, line 298: huh, we use `is` prefix for `isBrainless` to describe a boolean field and now I'm thinking using `isObservation` or `isObservationMode` is better for the consistency.

Konboi: yeah, it's old my bad... 
I prefer `isObservation` to `isObservationMode`. What do you think?

takanabe: agreed. Let's use `isObservation` here and change the payload of the request as well!

launchable/commands/subset.py
```diff
@@ -312,7 +314,11 @@ def run(self):
             if split:
                 click.echo("subset/{}".format(subset_id))
             else:
-                self.output_handler(output, rests)
+                _output, _rests = output, rests
```
shibuiwilliam, line 317: [question]
is it ok to switch output and rest even by default even if it is not `is_observation == true` ?

Konboi: OK, because we need to switch, if the `is_output_exclusion_rules` option is enabled 
https://github.com/launchableinc/cli/pull/435/files#diff-4c62e35fb4e3b2b6bff8593107cdddd1fd47eff90293edc9d838c9b2f3cff94aR339

launchable/commands/subset.py
Konboi: 👍

takanabe: What about `record session` [payload](https://github.com/launchableinc/cli/blob/571b00982cd5cab55f22fda87c2d2e14ee4cf1e0/launchable/commands/record/session.py#L75)? I think this should be `isObservation`.

Konboi: Ah, I agree with you! I'll fix it too.

takanabe (approved):
LGTM

launchable/commands/subset.py
Konboi: done https://github.com/launchableinc/cli/pull/438/commits/d66db7fb09febcabb5b13466bf6471bb2dc4dd6a

PR #439, opened 2022-08-08
Work correctly with Surefire+TestNG

kohsuke:
I just realized that I fundamentally misunderstood the nature of the problem https://github.com/launchableinc/cli/pull/434 is trying to address. The problem it was trying to solve is not to collect `testng-results.xml`, which will get processed if somebody runs `launchable record tests maven target/surefire-reports` like the doc currently points out.

My misunderstanding was that `target/surefire-reports/junitreports` do not get processed by `launchable record tests maven target/surefire-reports` because directories are not searched recursively. I incorrectly thought #434 is about avoiding `junitreports/TEST-*.xml` from getting processed.

Given that, I think this change is a better way to address this that supersedes #434. It makes the CLI "just work", in the sense that whether you are using JUnit or TestNG, the same invocation of `launchable record tests maven` would work.


launchable/test_runners/maven.py
```diff
@@ -70,4 +70,11 @@ def file2test(f: str) -> Optional[List]:
...
 split_subset = launchable.CommonSplitSubsetImpls(__name__).split_subset()
-record_tests = launchable.CommonRecordTestImpls(__name__).report_files()
+# TestNG produces surefire-reports/testng-results.xml in TestNG's native format.
+# Surefire produces TEST-*.xml in JUnit format (see Surefire's StatelessXmlReporter.getReportFile)
+# In addition, TestNG also produces surefire-reports/junitreports/TEST-*.xml (see TestNG's JUnitReportReporter.getFileName)
+# And there are more test reports in play.
+#
+# So to collectly find tests without duplications, we need to find surefire-reports/TEST-*.xml
+# not surefire-reports/**/TEST-*.xml nor surefire-reports/*.xml
+record_tests = launchable.CommonRecordTestImpls(__name__).report_files(file_mask="TEST-*.xml")
```
Konboi, line 80: Ah, Make sense. I agree with your idea 👍

Konboi (approved):
LGTM

PR #440, opened 2022-08-08
Documented the TestNG support in subsetting

kohsuke:
Explaining how to use https://github.com/launchableinc/testng/ for Maven+TestNG subsetting.

### DIFF
--- a/docs/resources/integrations/maven.md
+++ b/docs/resources/integrations/maven.md
@@ -53,8 +53,33 @@ launchable subset \
 * The `--build` should use the same `<BUILD NAME>` value that you used before in `launchable record build`.
 * The `--confidence` option should be a percentage; we suggest `90%` to start. You can also use `--time` or `--target`; see [Subsetting your test runs](../../features/predictive-test-selection/subsetting-your-test-runs.md) for more info.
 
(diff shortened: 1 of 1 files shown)


docs/resources/integrations/maven.md
```diff
@@ -53,8 +53,32 @@ launchable subset \
 * The `--build` should use the same `<BUILD NAME>` value that you used before in `launchable record build`.
 * The `--confidence` option should be a percentage; we suggest `90%` to start. You can also use `--time` or `--target`; see [Subsetting your test runs](../../features/predictive-test-selection/subsetting-your-test-runs.md) for more info.
 
-The `launchable subset` command outputs a file called `launchable-subset.txt` that you can pass into Maven to run Launchable's selected subset. You should **not** use any extra includes/excludes files here; they were already processed during `mvn test-compile`:
+### Maven + JUnit
+The `launchable subset` command outputs a file called `launchable-subset.txt` that you can pass into Maven to run Launchable's selected subset.
```
awilkes, line 57: I suggest duplicating this line to line 77 because line 78 references `launchable-subset.txt` without too much prior context.

awilkes (approved):

docs/resources/integrations/maven.md
kohsuke: Ah. I think a better fix is to move this line 57 above line 56, no? My thinking is that this sentence applies to all branches.

awilkes: That is indeed a better fix!

PR #446, opened 2022-08-24
Handle go-junit-report v2 changes

draftcode:
This absorbs the difference between go-junit-report v1 and v2. Making
the data backward compatible with v1.


launchable/test_runners/go_test.py
```diff
@@ -32,6 +37,37 @@ def subset(client):
...
+            return
+
+    default_path_builder = client.path_builder
+
+    def path_builder(case: TestCase, suite: TestSuite, report_file: str) -> TestPath:
+        tp = default_path_builder(case, suite, report_file)
+        for tpc in tp:
+            if 'type' in tpc and 'name' in tpc and tpc['type'] == 'class':
+                # go-junit-report v2 reports full package name. go-junit-report
+                # v2 reports only the last component of the package name. In
```
Konboi, line 63: > go-junit-report v2 reports only the last component of the package name

I think v1 is correct

launchable/test_runners/go_test.py
```diff
@@ -32,6 +37,37 @@ def subset(client):
...
+                # order to make this backward compatible, we align this to v1.
+                tpc['name'] = tpc['name'].split('/')[-1]
```
Konboi, line 65: make sense

Konboi (approved):
LGTM

launchable/test_runners/go_test.py
draftcode: Done.

PR #451, opened 2022-09-09
Send testRunner in subset command

takanabe:
# Background

We want to configure appropriate altitude mappings for each tenant automatically. To achieve the automatic configurations, we need test runner information for each tenant. 

We are [sending test runner information using UserAgent via Python's HTTP client](https://github.com/launchableinc/cli/blob/d5bd24b9084de85109d9392639651af4df616872/launchable/utils/http_client.py#L92-L106). 

```python
    def _headers(self, compress):
        h = {
            "User-Agent": "Launchable/{} (Python {}, {})".format(__version__, platform.python_version(),
                                                                 platform.platform()),
            "Content-Type": "application/json"
        }

        if compress:
            h["Content-Encoding"] = "gzip"

        if self.test_runner != "":
            h["User-Agent"] = h["User-Agent"] + \
                " TestRunner/{}".format(self.test_runner)

        return {**h, **authentication_headers()}
```

We will extend this capability to `subset` and record tests command payload so we can detect the test runners in our APIs. The original HTTP client implementation expects that received test runners possibly become empty string but `subset` command always receive the test runners. So, `testRunner` is always not empty string.


launchable/commands/subset.py
```diff
@@ -254,9 +254,10 @@ def default_path_builder(file_name):
                     if path:
                         self.test_paths.append(self.to_test_path(path))
 
-        def get_payload(self, session_id, target, duration):
+        def get_payload(self, session_id, target, duration, test_runner):
             payload = {
                 "testPaths": self.test_paths,
+                "testRunner": test_runner,
```
takanabe, line 260: The field name reflects our public document https://docs.launchableinc.com/resources/integrations.

Konboi: Could you add this field to record tests command too?

takanabe: Sure! Seems it's redundant but is this correct place if we think of the current payload structure?

```json
{
  "events": [
    {
      "type": "case",
      "testPath": [
        {
          "type": "file",
          "name": "components/layouts/loading/snapshot.test.tsx"
        },
        { "type": "testcase", "name": "Renders correctly" }
      ],
      "testRunner": "jest",   // this guy!
      "duration": 0.047,
      "status": 1,
      "stdout": "",
      "stderr": "",
      "created_at": "2021-11-09T05:36:19",
      "data": null
    },
```

https://github.com/launchableinc/cli/blob/3baf8d6068d6fe039064a9559e090fffb219d3b8/tests/data/jest/record_test_result.json#L3-L18

Konboi: In my image, like this
```
{
  "events": [],
   "testRunner": "jest"   // this guy!
}
```

What do you think?

takanabe: OK, that's simpler!

takanabe: added with https://github.com/launchableinc/cli/pull/451/commits/9f901a3be1513dd0cb23068ccf63260704b747d4

Konboi (approved):
LGTM Great work!!

awilkes:
I'm wondering if it would be better to define the _altitude_ in the profile and send _that_ value to the backend instead of the profile name, since the altitude for a profile might change, and we don't version profiles.

Konboi:
Exactly..., we should send altitude too

PR #457, opened 2022-09-27
Add a git-log parser

draftcode:
This implements --import-git-log-output option. In some cases, it's hard
to access the git repository from the machine that connects to the
Internet. This makes a workaround that a user can create a file based on
a Git log command (shown in the doc comment), transfer that to the
machine with the Internet connection, and run a commit ingestion.

Tested with one of the demo workspaces.

### DIFF
--- a/launchable/utils/commit_ingester.py
+++ b/launchable/utils/commit_ingester.py
@@ -1,0 +1,57 @@
+from .git_log_parser import GitCommit
+from .http_client import LaunchableClient
+from typing import List, Dict, Optional
+from datetime import tzinfo
+import hashlib
+
+
+def _sha256(s: str) -> str:
+    return hashlib.sha256(s.encode('utf8')).hexdigest()
+
+
+def _format_tzinfo(tz: Optional[tzinfo]) -> int:
+    if not tz:
+        return 0
+    delta = tz.utcoffset(None)
+    if not delta:
+        return 0
+    return round(delta.total_seconds() / 60)
+
+
+def _convert_git_commit(commit: GitCommit) -> Dict:
+    changed_files = []
+    for changed_file in commit.changed_files:
+        cf = dict()
+        cf['linesAdded'] = changed_file.added
+        cf['linesDeleted'] = changed_file.deleted
+        cf['status'] = 'M'
+        cf['path'] = changed_file.path
+        cf['pathTo'] = changed_file.path
+        changed_files.append(cf)
+    parents = dict()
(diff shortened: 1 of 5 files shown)


launchable/utils/git_log_parser.py
```diff
@@ -0,0 +1,50 @@
...
+
+def parse_git_log(fp: TextIO) -> List[GitCommit]:
+    """Parses the output of a git log command.
+
+    This parses the output of `git log --pretty='format:{"commit": "%H",
+    "parents": "%P", "authorEmail": "%ae", "authorTime": "%aI",
+    "committerEmail": "%ce", "committerTime": "%cI"}' --numstat`
+    """
+    ret = []
+    meta = {}  # type: Dict[str, Any]
```
shibuiwilliam, line 23: [imo]
it can be written like
```python
meta: Dict[str, Any] = {}
```

launchable/utils/git_log_parser.py
```diff
@@ -0,0 +1,50 @@
...
+    files = []  # type: List[ChangedFile]
```
shibuiwilliam, line 24: [imo]
same here
```python
files: List[ChangedFile] = []
```

launchable/utils/git_log_parser.py
draftcode: Note that this needs to be compatible with Python 3.5

launchable/utils/git_log_parser.py
draftcode: Same. Not compatible with 3.5

launchable/utils/commit_ingester.py
```diff
@@ -0,0 +1,57 @@
...
+        cf['status'] = 'M'
+        cf['path'] = changed_file.path
+        cf['pathTo'] = changed_file.path
+        changed_files.append(cf)
+    parents = dict()
+    if len(commit.parents) > 0:
+        # We don't know which diff is for which parent. Use the first parent.
+        parents[commit.parents[0]] = changed_files
+        for parent in commit.parents[1:len(commit.parents)]:
+            parents[parent] = []
```
shibuiwilliam, line 36: [question]
is it correct that you are adding empty list here?

draftcode: Yes. All changed files are added to the first parent, and the rest of the commits gets an empty changed file list.

shibuiwilliam (approved):

PR #458, opened 2022-09-28
Post to Slack channel when docs change

kohsuke:
We should be all aware of what's changing the doc


kohsuke:
The change currently goes to the CSM channel.

awilkes:
Great idea!

awilkes:
Hm, docs updates that come from Gitbook don't come through as pull requests. Can we watch for changed files on the main branch instead?

kohsuke:
Good catch. Fixed.

takanabe (approved):

PR #463, opened 2022-10-04
Add a page about the git-log-output import

draftcode:


docs/sending-data-to-launchable/running-under-restricted-networks.md
```diff
@@ -0,0 +1,88 @@
...
+"authorTime": "%aI", "committerEmail": "%ce", "committerTime": "%cI"}'
+--numstat`.
+
+```bash
+# Check out the repository and build artifacts
+git clone https://git.internal.example.com/repo
+cd repo
+make all
+
+# Save git log data to a file. Using --max-count to limit the number of commits.
```
awilkes, line 54: Since it's important, we should include `--max-count` in the CLI reference and on line 43 in this file.

I'd like to provide guidance on the right value to choose, but that seems impossible.

Because of this, perhaps we could recommend using `--since` instead of `--max-count`? e.g. --since="2 days ago"

awilkes (approved):

docs/sending-data-to-launchable/running-under-restricted-networks.md
draftcode: I think the time range is also tricky (consider that the test run after a long weekend). I just gave 10-20 as a ballpark figure. This should be suffice in many cases.

awilkes: OK! We can always adjust later.

PR #464, opened 2022-10-04
Example should show multiple createdFiles.lst

kohsuke:
... which is standard for most real world multi-module setup.

### DIFF
--- a/docs/resources/integrations/maven.md
+++ b/docs/resources/integrations/maven.md
@@ -47,8 +47,8 @@ This process generates a file you can pass into `launchable subset`. This file i
 launchable subset \
   --build <BUILD NAME> \
   --confidence <TARGET> \
-  maven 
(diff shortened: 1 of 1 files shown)


kohsuke:
`<(cmd)` is a bash process substitution: https://tldp.org/LDP/abs/html/process-sub.html
I'm not sure the portability with other shell variants.

Earlier I experimented with `$(find . -path '*/created-file.lst' -printf "--test-compile-created-file=%p\n")` but couldn't make the multi-argument handling right, so I abandoned it.

kohsuke:
Given the history of that, I appreciate independent experimentation to make sure this works.

kohsuke:
`{}` in `find -exec` handles whitespace in the path name as expected, meaning the tokenization is done by the shell that invokes `find`, and `find` simply puts the whole path name as one argument and then spawns a child process.

draftcode (approved):

PR #465, opened 2022-10-05
rename bin variable name to bin_target to avoid using reserved word bin

shibuiwilliam:
# What
- Renamed variable `bin` in Python to `bin_target` since `bin` is a reserved word in Python.
- https://docs.python.org/ja/3/library/functions.html#bin


Konboi (approved):
LGTM 

> bin is a reserved word in Python.

TIL

PR #471, opened 2022-10-12
add `stats test-sessions` subcommand

Songmu:
With this change, we can get the stats of test sessions via the API, which can be obtained with the stats subcommand.

There is also a function to filter by number of days and flavor, which is useful for predicting test times before running tests.

```console
$ curl -H "Content-Type: application/json" -H "Authorization: Bearer $LAUNCHABLE_TOKEN" https://api.mercury.launchableinc.com/intake/organizations/OOO/workspaces/WWW/stats/test-sessions'?flavor=os=ubuntu-latest'
{"averageDurationSeconds":0.007610803807083564,"count":158,"days":7}
```

```console
$ launchable stats test-sessions --flavor=os=windows-latest
{"averageDurationSeconds":0.007610803807083564,"count":158,"days":7}
```

### DIFF
--- a/launchable/commands/stats/test_sessions.py
+++ b/launchable/commands/stats/test_sessions.py
@@ -10,7 +10,7 @@ from ...utils.http_client import LaunchableClient
 @click.option(
     '--days',
     'days',
-    help='subsetting target from 0% to 100%',
+    help='How many days of test sessions in the past to be stat',
     type=int,
     default=7
 )
--- a/launchable/commands/stats/test_sessions.py
+++ b/launchable/commands/stats/test_sessions.py
@@ -1,8 +1,10 @@
 import click
 import os
-from typing import Sequence, List
+from typing import List
+
 from ...utils.click import KeyValueType
 from ...utils.env_keys import REPORT_ERROR_KEY
+from ...utils.flavor import normalize_flavors
 from ...utils.http_client import LaunchableClient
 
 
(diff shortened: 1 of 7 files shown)


launchable/utils/flavor.py
```diff
@@ -0,0 +1,23 @@
+from typing import List, Sequence, Tuple
```
Konboi, line 1: 👍

Konboi (approved):
LGTM but please approve from @shibuiwilliam, just incase

launchable/commands/stats/test_sessions.py
```diff
@@ -0,0 +1,58 @@
...
+    # If we don't write this, `pip run type` will assume it is Dict[str, int], and the check will not pass.
+    params = {'days': days, 'flavor': []}
+    flavors = []
+    for f in normalize_flavors(flavor):
+        flavors.append('%s=%s' % (f[0], f[1]))
+
+    if flavors:
+        params['flavor'] = flavors
+    else:
+        del params['flavor']
```
shibuiwilliam, line 43: [nits]
Just to avoid removing a key from dict does not raise error for the key does not exist (I know obviously it does not happen in this code), I prefer to write `params.pop("flavor", None)` which will remove a key if exists or do nothing if not exists.

Songmu: I see. This is a very good lesson. Thank you very much.

PR #473, opened 2022-10-17
Add warning when the user set --observation option to subset command not record session command

Konboi:
When executing a `launchable subset` command with a session issued by the `launchable record session` command and want to enable observation mode, the user needs to `—observation` option to the `launchable record session` command.

However, sometimes the option is set to the `launchable subset` command, and observation mode isn’t enabled.

So, added a warning message when it happened.

```sh
$ $ launchable record build --name observation-mode
$ launchable record session --build observation-mode > session.txt
$ find tests/**/test_*.py | launchable subset --session $(cat session.txt) --target 30% --observation file > subset.txt
Warning: you set --observation option this command but you need to set it `launchable record session` command # added
Your model is currently in training
Launchable created subset 96 for build observation-mode (test session 69) in workspace launchableinc/mothership

|           |   Candidates |   Estimated duration (%) |   Estimated duration (min) |
|-----------|--------------|--------------------------|----------------------------|
| Subset    |           13 |                  28.8889 |                0.000216667 |
| Remainder |           32 |                  71.1111 |                0.000533333 |
|           |              |                          |                            |
| Total     |           45 |                 100      |                0.00075     |

Run `launchable inspect subset --subset-id 96` to view full subset details
```

### DIFF
--- a/launchable/commands/helper.py
+++ b/launchable/commands/helper.py
@@ -27,7 +27,7 @@ def find_or_create_session(
     from .record.session import session as session_command
 
     if session:
-        check_observation_mode_status(session, is_observation)
+        _check_observation_mode_status(session, is_observation)
         return session
 
     saved_build_name = read_build()
@@ -47,7 +47,7 @@ def find_or_create_session(
 
         session_id = read_session(saved_build_name)
         if session_id:
-            check_observation_mode_status(session_id, is_observation)
+            _check_observation_mode_status(session_id, is_observation)
             return session_id
         else:
             context.invoke(
@@ -61,7 +61,7 @@ def find_or_create_session(
             return read_session(saved_build_name)
 
 
-def check_observation_mode_status(session: str, is_observation: bool):
+def _check_observation_mode_status(session: str, is_observation: bool):
     if is_observation is False:
         return
 
--- a/launchable/commands/helper.py
+++ b/launchable/commands/helper.py
@@ -1,7 +1,7 @@
 from typing import Optional
-
 import click
 
+from ..utils.http_client import LaunchableClient
 from ..utils.session import read_build, read_session
 
 
@@ -27,6 +27,7 @@ def find_or_create_session(
     from .record.session import session as session_command
 
     if session:
+        check_observation_mode_status(session, is_observation)
         return session
 
     saved_build_name = read_build()
@@ -46,6 +47,7 @@ def find_or_create_session(
 
         session_id = read_session(saved_build_name)
         if session_id:
+            check_observation_mode_status(session_id, is_observation)
             return session_id
         else:
             context.invoke(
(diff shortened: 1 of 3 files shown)


launchable/commands/helper.py
```diff
@@ -57,3 +59,19 @@ def find_or_create_session(
...
+
+    client = LaunchableClient()
+    res = client.request("get", session)
+
+    # only check when the status code is 200 not to stop the command
+    if res.status_code == 200:
+        is_observation_in_recorded_session = res.json().get("isObservation", False)
+        if is_observation and is_observation_in_recorded_session is False:
+            click.echo(click.style(
+                "Warning: you set --observation option in this command but you need to set it `launchable record session` command", fg='yellow'),
```
Konboi, line 76: @awilkes Could you check this error message?

launchable/commands/helper.py
```diff
@@ -57,3 +59,19 @@ def find_or_create_session(
...
+    if is_observation is False:
+        return
+
+    client = LaunchableClient()
+    res = client.request("get", session)
+
+    # only check when the status code is 200 not to stop the command
+    if res.status_code == 200:
+        is_observation_in_recorded_session = res.json().get("isObservation", False)
+        if is_observation and is_observation_in_recorded_session is False:
```
shibuiwilliam, line 74: [must]
Better and rather common to write like `if is_observation and not is_observation_in_recorded_session`. 
`is` in Python is used to compare two variables pointing to the same object in memory. I know it works for comparing boolean, since `True` and `False` are singleton in Python, but it is Pythonic to just write like `if a` to evaluate `a==True` or `a==False`.

launchable/commands/helper.py
shibuiwilliam, line 76: [nits]
I feel it is better to point out clearly that `--observation` option does not work if used in `launchable subset` command.

launchable/commands/helper.py
```diff
@@ -57,3 +59,19 @@ def find_or_create_session(
                 is_observation=is_observation,
             )
             return read_session(saved_build_name)
+
+
+def _check_observation_mode_status(session: str, is_observation: bool):
+    if is_observation is False:
```
shibuiwilliam, line 65: same here.
https://github.com/launchableinc/cli/pull/473/files#r996671800

launchable/commands/helper.py
awilkes:
```
WARNING: --observation flag was ignored. Observation mode can only be enabled for a test session during its initial creation. Add `--observation` option to the `launchable record session` command instead.
```

shibuiwilliam (approved):

launchable/commands/helper.py
Konboi: Thanks, I'll change

PR #489, opened 2022-11-14
Implement an experimental GitHub OIDC auth option

draftcode:


launchable/utils/authentication.py
```diff
@@ -36,6 +37,24 @@ def authentication_headers():
     if token:
         return {'Authorization': 'Bearer {}'.format(token)}
 
+    if os.getenv('EXPERIMENTAL_GITHUB_OIDC_TOKEN_AUTH'):
+        req_url = os.environ['ACTIONS_ID_TOKEN_REQUEST_URL']
+        rt_token = os.environ['ACTIONS_ID_TOKEN_REQUEST_TOKEN']
+        if not req_url or not rt_token:
```
shibuiwilliam, line 43: [must]
`os.environ[]` may raise key error if the keys are not present. 
Better to use `os.getenv()` which will return None if the key is absent.

draftcode: Done

launchable/utils/authentication.py
```diff
@@ -36,6 +37,24 @@ def authentication_headers():
...
+                    "Confirm that you have added necessary permissions following "
+                    "https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers#adding-permissions-settings",  # noqa: E501
+                    fg="red"))
+        r = requests.get(req_url,
+                         headers={
+                             'Authorization': 'Bearer {}'.format(rt_token),
+                             'Accept': 'application/json; api-version=2.0',
+                             'Content-Type': 'application/json',
+                         })
+        return {'Authorization': 'Bearer {}'.format(r.json()['value'])}
```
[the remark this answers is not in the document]
draftcode: Done

PR #492, opened 2022-11-15
fix lint and format config

shibuiwilliam:


Songmu (approved):
Excellent. 
This would have been monotonous and tedious work, but someone has to do it. I am very glad you did it.

PR #497, opened 2022-11-16
Support zero input subset for raw profile

Konboi:
Now the raw plugin isn't supported [Zero Input Subsetting](https://docs.launchableinc.com/resources/integrations/gradle#using-zero-input-subsetting).
So, I updated the raw plugin to support Zero Input Subsetting.

e.g)
```
$ launchable subset --target 30% --get-tests-from-previous-sessions -rest rest.txt > subset.txt
$ pytest $(cat rest.txt | grep file= | awk -F# '{print $1}' |  awk -F= '{print $2}' | sort | uniq | xargs -I% echo '--ignore %' | tr '\n' ' ')
```

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -179,6 +179,7 @@ def subset(
             self.output_handler = self._default_output_handler
             self.exclusion_output_handler = self._default_exclusion_output_handler
             self.is_get_tests_from_previous_sessions = is_get_tests_from_previous_sessions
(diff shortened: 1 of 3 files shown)


draftcode:
Same as the other PR. I see some formatting changes, and it seems to me that it's reverting to the previous format.

Konboi:
@draftcode  Sorry, I fixed code format. Please review, again.

draftcode (approved):

PR #502, opened 2022-11-18
Split subset by group

Konboi:
Add `--split-by-groups` option to the `split-subset` command

e.g)

```sh
$ launchable record tests --group e2e ./e2e-test-result.xml
$ find ./tests/**/test_*.py  | launchable subset --split --target 40% file > subset_id.txt
$ launchable split-subset --subset-id $(cat subset-id.txt) --split-by-groups file
$ ls -l 
rest-e2e.txt
rest-nogroup.txt
subset-e2e.txt 
subset-groups.txt
subset-nogroup.txt
```

### DIFF
--- a/launchable/commands/split_subset.py
+++ b/launchable/commands/split_subset.py
@@ -264,7 +264,7 @@ def split_subset(
                           "w+", encoding="utf-8") as f:
                     f.write("\n".join(rest_group_names))
 
-        def split_by_group_namess(self):
+        def split_by_group_names(self):
             try:
                 res = client.request("POST", "{}/split-by-groups".format(subset_id))
                 res.raise_for_status()
@@ -312,7 +312,7 @@ def split_subset(
                     "Missing option '--bin' or '--split-by-groups/--split-by-groups-with-rest'")
 
             if self._is_split_by_groups():
-                self.split_by_group_namess()
+                self.split_by_group_names()
             else:
                 self.split_by_bin()
 
--- a/launchable/commands/split_subset.py
+++ b/launchable/commands/split_subset.py
@@ -100,7 +100,7 @@ def split_subset(
             self.split_by_groups_output_dir = split_by_groups_output_dir
 
         def _default_split_by_groups_output_handler(self, group_name: str, subset: List[TestPath], rests: List[TestPath]):
-            if is_split_by_groups_with_rest and len(rests) > 0:
+            if is_split_by_groups_with_rest:
                 self.write_file("{}/rest-{}.txt".format(split_by_groups_output_dir, group_name), rests)
 
             if len(subset) > 0:
@@ -259,7 +259,7 @@ def split_subset(
                           "w+", encoding="utf-8") as f:
                     f.write("\n".join(subset_group_names))
 
-            if is_split_by_groups_with_rest and len(rest_group_names) > 0:
+            if is_split_by_groups_with_rest:
                 with open("{}/{}".format(split_by_groups_output_dir, SPLIT_BY_GROUP_REST_GROUPS_FILE_NAME),
                           "w+", encoding="utf-8") as f:
                     f.write("\n".join(rest_group_names))
@@ -291,7 +291,7 @@ def split_subset(
                     if len(subset) > 0 and group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
                         subset_group_names.append(group_name)
 
-                    if len(rests) > 0 and group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
+                    if group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
                         rest_group_names.append(group_name)
 
                     if is_output_exclusion_rules:
--- a/launchable/commands/split_subset.py
+++ b/launchable/commands/split_subset.py
@@ -10,6 +10,7 @@ from .test_path_writer import TestPathWriter
 
 SPLIT_BY_GROUPS_NO_GROUP_NAME = "nogroup"
 SPLIT_BY_GROUP_SUBSET_GROUPS_FILE_NAME = "subset-groups.txt"
+SPLIT_BY_GROUP_REST_GROUPS_FILE_NAME = "rest-groups.txt"
 
 
 @click.group(help="Split subsetting tests")
@@ -79,14 +80,13 @@ def split_subset(
 ):
     TestPathWriter.base_path = base_path
 
+    client = LaunchableClient(test_runner=context.invoked_subcommand, dry_run=context.obj.dry_run)
+
     class SplitSubset(TestPathWriter):
         def __init__(self, dry_run: bool = False):
             super(SplitSubset, self).__init__(dry_run=dry_run)
 
-        def run(self):
-            if (not is_split_by_groups and bin_target is None) or (is_split_by_groups and bin_target):
-                raise click.BadOptionUsage("--bin or --split-by-groups", "Missing option '--bin' or '--split-by-groups'")
-
+        def split_by_bin(self):
             index, count = 0, 0
             if not is_split_by_groups:
                 index = bin_target[0]
@@ -117,8 +117,6 @@ def split_subset(
             is_observation = False
 
             try:
-                client = LaunchableClient(test_runner=context.invoked_subcommand, dry_run=context.obj.dry_run)
-
                 payload = {
                     "sliceCount": count,
                     "sliceIndex": index,
@@ -199,43 +197,7 @@ def split_subset(
                 output = res.json().get("testPaths", [])
                 rests = res.json().get("rest", [])
                 is_observation = res.json().get("isObservation", False)
-                split_groups = res.json().get("splitGroups", {})
 
-            except Exception as e:
-                if os.getenv(REPORT_ERROR_KEY):
-                    raise e
-                else:
-                    click.echo(e, err=True)
-                    click.echo(click.style(
-                        "Warning: the service failed to split subset. Falling back to running all tests", fg='yellow'),
-                        err=True)
-                    return
-
-            if is_split_by_groups:
-                group_names = []
-                for group in split_groups:
-                    group_name = group.get("groupName", "")
-                    subset = group.get("subset", [])
-                    rests = group.get("rest", [])
-
-                    if is_observation:
-                        subset, rests = subset + rests, []
-                    if is_output_exclusion_rules:
-                        subset, rests = rests, subset
-
-                    if len(subset) > 0:
-                        self.write_file("{}/subset-{}.txt".format(split_by_groups_output_dir, group_name), subset)
-                        if group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
-                            group_names.append(group_name)
-
-                    if len(rests) > 0:
-                        self.write_file("{}/rest-{}.txt".format(split_by_groups_output_dir, group_name), rests)
-
-                if len(group_names) > 0:
-                    with open("{}/{}".format(split_by_groups_output_dir, SPLIT_BY_GROUP_SUBSET_GROUPS_FILE_NAME),
-                              "w+", encoding="utf-8") as f:
-                        f.write("\n".join(group_names))
-            else:
                 if len(output) == 0:
                     click.echo(click.style(
                         "Error: no tests found in this subset id.", 'yellow'), err=True)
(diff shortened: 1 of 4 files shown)


launchable/commands/split_subset.py
```diff
@@ -60,28 +84,33 @@ def __init__(self, dry_run: bool = False):
...
-            if count < index:
-                click.echo(
-                    click.style(
-                        'Error: invalid bin value. Make sure to set below 1 like `--bin 1/2`, `--bin 2/2` '
-                        'but set `--bin {}`'.format(bin_target),
-                        'yellow'),
-                    err=True,
-                )
-                return
+            if not is_split_by_groups and bin_target is None:
```
shibuiwilliam, line 87: [ask]
Is it ok to specify `is_split_by_groups` and `bin_target` simultaneously?

launchable/commands/split_subset.py
```diff
@@ -180,18 +211,44 @@ def run(self):
...
+                        if group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
+                            group_names.append(group_name)
+
+                    if len(rests) > 0:
+                        self.write_file("{}/rest-{}.txt".format(split_by_groups_output_dir, group_name), rests)
+
+                if len(group_names) > 0:
+                    open(
+                        "{}/{}".format(split_by_groups_output_dir, SPLIT_BY_GROUP_SUBSET_GROUPS_FILE_NAME), "w+",
+                        encoding="utf-8").write("\n".join(group_names))
```
shibuiwilliam, line 237: [must]
Since Python won't close an open file automatically, it is recommended to open in `with` clause to avoid forgetting to close.

```python
with open(
      "{}/{}".format(split_by_groups_output_dir, SPLIT_BY_GROUP_SUBSET_GROUPS_FILE_NAME), "w+",
      encoding="utf-8") as f:
      f.write("\n".join(group_names))
```

launchable/commands/split_subset.py
Konboi: No, it isn't so I fixed it. Thank you!

launchable/commands/split_subset.py
```diff
@@ -60,28 +84,33 @@ def __init__(self, dry_run: bool = False):
...
-                    click.style(
-                        'Error: invalid bin value. Make sure to set below 1 like `--bin 1/2`, `--bin 2/2` '
-                        'but set `--bin {}`'.format(bin_target),
-                        'yellow'),
-                    err=True,
-                )
-                return
+            if (not is_split_by_groups and bin_target is None) or (is_split_by_groups and bin_target):
+                raise click.BadOptionUsage("--bin or --split-by-groups", "Missing option '--bin' or '--split-by-groups'")
+
```
draftcode, line 89: We can create `run_split_by_group` and immediately branch to that method at the beginning. Most of the code below are orthogonal to the binning option, and some of the behaviors do not make sense to this mode (e.g. the fallback code).

launchable/commands/split_subset.py
```diff
@@ -180,18 +211,44 @@ def run(self):
...
+                    if is_output_exclusion_rules:
+                        subset, rests = rests, subset
+
+                    if len(subset) > 0:
+                        self.write_file("{}/subset-{}.txt".format(split_by_groups_output_dir, group_name), subset)
+                        if group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
+                            group_names.append(group_name)
+
+                    if len(rests) > 0:
+                        self.write_file("{}/rest-{}.txt".format(split_by_groups_output_dir, group_name), rests)
```
draftcode, line 232: This is not aligned with subset https://github.com/launchableinc/cli/blob/main/launchable/commands/subset.py#L184-L192

launchable/commands/split_subset.py
```diff
@@ -180,18 +211,44 @@ def run(self):
...
+                        subset, rests = subset + rests, []
+                    if is_output_exclusion_rules:
+                        subset, rests = rests, subset
+
+                    if len(subset) > 0:
+                        self.write_file("{}/subset-{}.txt".format(split_by_groups_output_dir, group_name), subset)
+                        if group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
+                            group_names.append(group_name)
+
+                    if len(rests) > 0:
```
draftcode, line 231: Per spec, this should be produced only when --rest is used.

launchable/commands/split_subset.py
```diff
@@ -180,18 +211,44 @@ def run(self):
...
+                        self.write_file("{}/rest-{}.txt".format(split_by_groups_output_dir, group_name), rests)
+
+                if len(group_names) > 0:
+                    with open("{}/{}".format(split_by_groups_output_dir, SPLIT_BY_GROUP_SUBSET_GROUPS_FILE_NAME),
```
draftcode, line 235: Per spec, rest-groups should be created.

launchable/commands/split_subset.py
```diff
@@ -180,18 +211,44 @@ def run(self):
                         err=True)
                     return
 
-            if len(output) == 0:
-                click.echo(click.style("Error: no tests found in this subset id.", 'yellow'), err=True)
-                return
+            if is_split_by_groups:
+                group_names = []
+                for group in split_groups:
```
draftcode, line 216: Per spec, nogroup file should be created.

draftcode (changes requested):

launchable/commands/split_subset.py
draftcode: Read this again. I read in a wrong way.

launchable/commands/split_subset.py
Konboi: click v8 supports optional value https://click.palletsprojects.com/en/8.1.x/options/. 
However, we use click v7 when the python version is under 3.6 https://github.com/launchableinc/cli/blob/main/setup.cfg#L18-L20

Hum... how to implement the rest option with the split-by-groups option ...

Konboi:
## Changes 

- Dividied methods to `split_by_bin` and `split_by_group_names`
- Added `--split-by-groups-with-rest` option
  - Because needs to pass PATH to `--rest` option, however split by groups won't use the path will be received from `--rest` option.
- Added `split_by_groups_output_handler` and `split_by_groups_exclusion_output_handler` properties to be able to edit output format.
  - Supported only gradle plugin

launchable/commands/split_subset.py
```diff
@@ -52,48 +82,74 @@ def split_subset(
...
 
+    client = LaunchableClient(test_runner=context.invoked_subcommand, dry_run=context.obj.dry_run)
+
     class SplitSubset(TestPathWriter):
         def __init__(self, dry_run: bool = False):
+            self.split_by_groups_output_handler = self._default_split_by_groups_output_handler
+            self.split_by_groups_exclusion_output_handler = self._default_split_by_groups_exclusion_output_handler
+            self.is_split_by_groups_with_rest = is_split_by_groups_with_rest
+            self.split_by_groups_output_dir = split_by_groups_output_dir
             super(SplitSubset, self).__init__(dry_run=dry_run)
```
shibuiwilliam, line 100: [nits]
It's rather common to write `super().__init__()` first to override parent constructors. Even though it does not effect in this code, writing this phrase at last will make parent variables override child variables when the child class is meant override parant, which is nonsense.

launchable/commands/split_subset.py
```diff
@@ -180,18 +250,76 @@ def run(self):
...
+                res = client.request("POST", "{}/split-by-groups".format(subset_id))
+                res.raise_for_status()
+
+                is_observation = res.json().get("isObservation", False)
+                split_groups = res.json().get("splitGroups", {})
+
+                subset_group_names = []
+                rest_group_names = []
+
+                for group in split_groups:
```
shibuiwilliam, line 278: [question]
Is it ok `res.json().get("splitGroups")` returns dict (`{}`)?
If it is dict, the `group` in `for group in split_groups:` would be keys in dict.

launchable/commands/split_subset.py
```diff
@@ -180,18 +250,76 @@ def run(self):
...
+                    group_name = group.get("groupName", "")
+                    subset = group.get("subset", [])
+                    rests = group.get("rest", [])
+
+                    if is_observation:
+                        subset, rests = subset + rests, []
+
+                    if is_output_exclusion_rules:
+                        for g in split_groups:
+                            if g.get("groupName", "") != group_name:
```
shibuiwilliam, line 288: [question]
Is it correct that you do `for g in split_groups` inside of `for group in split_groups`?

launchable/commands/split_subset.py
Konboi: Ops, splitGroups is an array. I'll fix default value

launchable/commands/split_subset.py
Konboi: I want to get all subsets and rests of other groups. Is there any good way to get it without loops?

shibuiwilliam: If that is the case, then this loop is fine.
Another way is to write in list comprehension, which tends to be faster than usual for loop in Python, but not mandatory.

Maybe something like this (I didn't tested):

```python
rests += [g.get("subset", []) + g.get("rest", []) for g in split_groups if g.get("groupName", "") != group_name]
```

launchable/commands/split_subset.py
shibuiwilliam: ok

launchable/commands/split_subset.py
shibuiwilliam: ok

launchable/commands/split_subset.py
shibuiwilliam: Anyway you don't have to fix here if you feel unnecessary.

Konboi: one line is a little difficult to read (in my case), so I'll keep it. Thanks

shibuiwilliam (approved):

launchable/commands/split_subset.py
```diff
@@ -44,6 +49,31 @@
...
 )
+@click.option(
+    "--split-by-groups",
+    'is_split_by_groups',
+    help="split by groups that were set by `launchable record tests --group`",
+    is_flag=True
+)
+@click.option(
+    "--split-by-groups-with-rest",
+    'is_split_by_groups_with_rest',
```
draftcode, line 60: The spec was using --rest instead of adding this. Is there a reason for not going with the spec?

launchable/commands/split_subset.py
```diff
@@ -180,18 +250,76 @@ def run(self):
...
+                    subset = group.get("subset", [])
+                    rests = group.get("rest", [])
+
+                    if is_observation:
+                        subset, rests = subset + rests, []
+
+                    if is_output_exclusion_rules:
+                        for g in split_groups:
+                            if g.get("groupName", "") != group_name:
+                                rests = rests + g.get("subset", []) + g.get("rest", [])
```
draftcode, line 289: This part is not needed. As discussed in the spec, we don't consider this case for now. You can just use the `rests` as is without collecting all other rests.

launchable/commands/split_subset.py
```diff
@@ -180,18 +250,76 @@ def run(self):
...
-            if is_observation:
-                output = output + rests
-            if rest:
-                if len(rests) == 0:
-                    rests.append(output[0])
+                    if len(subset) > 0 and group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
+                        subset_group_names.append(group_name)
 
-                self.write_file(rest, rests)
+                    if len(rests) > 0 and group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
```
draftcode, line 294: `rest_group_names` should be the complement to `subset_group_names`. This creates an overlap.

draftcode (changes requested):

launchable/commands/split_subset.py
Konboi: Ooo, I had recognized it incorrectly...

launchable/commands/split_subset.py
Konboi: yes, originally set rest output **file** path to the `--rest` option. Also, needs to set something value when we use the `—rest` option. e.g) launchable-rest.txt

But in split by group should produce `{OUTPUT_DIR}/rest-${GROUP_NAME}.txt`. In my understanding Set `OUTPUT_DIR` path by `---split-by-groups-output-dir` options

So, if we keep the `—rest` option, the user needs to set a dummy path `—rest hoge` to use the rest option.

draftcode: Got it. Updated the spec with the rationale.

launchable/commands/split_subset.py
```diff
@@ -180,18 +250,71 @@ def run(self):
...
-            if is_observation:
-                output = output + rests
-            if rest:
-                if len(rests) == 0:
-                    rests.append(output[0])
+                    if len(subset) > 0 and group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
+                        subset_group_names.append(group_name)
 
-                self.write_file(rest, rests)
+                    if group_name != SPLIT_BY_GROUPS_NO_GROUP_NAME:
```
draftcode, line 289: `elif`?

Konboi: When the `--output-exclusion-rules` option enable, subset group and rest group will switch. Then, there is rest data, but if it's not in rest_group_names, it's difficult to create subset_group_names.txt
So I added filter logic https://github.com/launchableinc/cli/pull/502/files/8bd7e154a0063632b2a3cf74640024c442cf89c9#diff-36cadddae14cddbb1f42e166c7cb78bb91deeaf3761052e9438363d14dd2c2cbR257-R260

Konboi: I misunderstood so I fixed

Konboi:
@draftcode PTAL 🙇

draftcode (approved):

PR #505, opened 2022-11-22
Teach the raw profile to take a JUnit XML file

draftcode:
Currently the raw profile only takes a JSON file. This change make it
take a JUnit XML file.


takanabe (approved):
LGTM

draftcode:
It seems the failure is not related to this change...

PR #506, opened 2022-11-23
Gradle+testng

kohsuke:
Documenting Gradle+TestNG that uses the Launchable TestNG extension.


awilkes (approved):
We should add a link to the Gradle+TestNG page from the regular Gradle page, but it's not a blocker.

PR #509, opened 2022-11-23
Docs: raw profile JUnit support

awilkes:


launchable/commands/record/tests.py
```diff
@@ -214,9 +214,13 @@ def parse(report: str) -> Generator[CaseEventType, None, None]:
                 else:
                     raise InvalidJUnitXMLException(filename=report)
 
-                for suite in testsuites:
-                    for case in suite:
-                        yield CaseEvent.from_case_and_suite(self.path_builder, case, suite, report)
+                try:
```
draftcode, line 217: (Just rant) This is where Gerrit shines. GitHub cannot handle these PR-on-PR things well.

PR #516, opened 2022-12-07
zero input subset + TestNG support documentation

kohsuke:
Documenting how to use Zero Input Subset with TestNG, which impacts both Maven & Gradle.

I created a separate section in Maven to discuss various complications, as opposed to duplicating this twice. Probably a matter of taste, but I made this change based on my thinking that there will be more complications to be added in the future.


awilkes (approved):
Thanks @kohsuke

PR #523, opened 2022-12-14
Support  --ignore-flaky-tests-above option

Konboi:
Support new option `--ignore-flaky-tests-above`.
This option can ignore flaky tests with higher scores than the value as you set this option.
You can confirm flaky scores on the WebApp.

e.g)
```sh
find tests/**/test_*.py | launchable subset --target 30% --ignore-flaky-tests-above 0.3 file > launchable-subset.txt
```


awilkes (approved):
Good stuff.

draftcode (approved):

PR #530, opened 2022-12-22
reject %25

shibuiwilliam:
# What
- Reject taking `%25` for build name that may be potentially malicious string.

### DIFF
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -69,6 +69,8 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
 
     if "/" in build_name or "%2f" in build_name.lower():
         sys.exit("--name must not contain a slash and an encoded slash")
+    if "%25" in build_name:
+        sys.exit("--name must not contain %25")
     if not no_commit_collection and len(commits) != 0:
         sys.exit("--no-commit-collection must be specified when --commit is used")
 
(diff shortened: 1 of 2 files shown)


launchable/commands/record/session.py
```diff
@@ -72,6 +72,11 @@ def session(
     you should set print_session = False because users don't expect to print session ID to the subset output.
     """
 
+    if "/" in build_name or "%2f" in build_name.lower():
```
Konboi, line 75: I think we need this validation only for the build command.

shibuiwilliam: fixed
https://github.com/launchableinc/cli/pull/530/commits/1ac7aa7512721010d60cc22f5fe185734a7cce92

launchable/commands/record/build.py
```diff
@@ -69,6 +69,8 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
 
     if "/" in build_name or "%2f" in build_name.lower():
         sys.exit("--name must not contain a slash and an encoded slash")
+    if "%25" in build_name:
+        sys.exit("--name must not contain %25")
```
Konboi, line 73: How about `"--name must not contain encoded % (%25)"`?

Konboi (approved):

launchable/commands/record/build.py
shibuiwilliam: fixed
https://github.com/launchableinc/cli/pull/530/commits/5364a285aefafed554a6a77ce36d60198bf16253

PR #532, opened 2023-01-12
[WA-77] format with fixed point notation

kohsuke:
Estimated duration of "3.33333e-05" is silly. Our valid value range is far limited, and the precision at those small numbers are illusions.

Better to use a fixed point precision. Also limit the number of precision digits. There's enough measurement margin errors, so added digits do not add any real value.

Briefly I was looking to print "3" as "3", not "3.00" but I don't think there's any predefined formatter that does it.

See: https://docs.python.org/3/library/string.html#format-specification-mini-language

### DIFF
--- a/launchable/commands/inspect/subset.py
+++ b/launchable/commands/inspect/subset.py
@@ -47,7 +47,7 @@ def subset(subset_id: int):
     rest_row = convert_row(rest, len(subset) + 1, False)
     rows = subset_row + rest_row
 
-    click.echo(tabulate(rows, header, tablefmt="github"))
+    click.echo(tabulate(rows, header, tablefmt="github", floatfmt=".2f"))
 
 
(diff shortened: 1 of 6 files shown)


launchable/commands/inspect/subset.py
```diff
@@ -47,7 +47,7 @@ def subset(subset_id: int):
     rest_row = convert_row(rest, len(subset) + 1, False)
     rows = subset_row + rest_row
 
-    click.echo(tabulate(rows, header, tablefmt="github"))
+    click.echo(tabulate(rows, header, tablefmt="github", floatfmt=".2f"))
```
Konboi, line 50: memo: https://github.com/astanin/python-tabulate#number-formatting

Konboi (approved):

PR #536, opened 2023-02-06
Fix subsetting

shibuiwilliam:
# What
- This PR is to validate subset ID in the split subset request to ensure the value exists, and cannot be null.

### DIFF
--- a/launchable/commands/split_subset.py
+++ b/launchable/commands/split_subset.py
@@ -87,6 +87,13 @@ def split_subset(
         split_by_groups_output_dir: click.Path,
         is_output_exclusion_rules: bool,
 ):
+    if subset_id == "" or not subset_id:
+        click.echo(
+            click.style('Error: subsetting ID cannot be empty', 'yellow'),
+            err=True,
+        )
+        return
+
     TestPathWriter.base_path = base_path
 
     client = LaunchableClient(test_runner=context.invoked_subcommand, dry_run=context.obj.dry_run)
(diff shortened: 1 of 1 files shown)


launchable/commands/split_subset.py
```diff
@@ -87,6 +87,14 @@ def split_subset(
         split_by_groups_output_dir: click.Path,
         is_output_exclusion_rules: bool,
 ):
+    if subset_id == "" or not subset_id:
```
Konboi, line 90: This validation is already applied by the click ref: https://github.com/launchableinc/cli/pull/536/files#diff-36cadddae14cddbb1f42e166c7cb78bb91deeaf3761052e9438363d14dd2c2cbR24

Also, even if set this validation, an error will happen like below.
```sh
 launchable split-subset --subset-id subset/ --bin 1/2 file
500 Server Error:  for url: http://localhost:8080/intake/organizations/ogogog/workspaces/wswsws/subset//slice
```

So, we need to another validation for the `subset_id`

e.g) 
```python
if (len(subset_id.split("/")) != 2):
  click.echo(clickc.style('Error: invalid subset_id format. Make sure to set subset-id option `subset/<subset id>` format))
```
( error message is tentative so need to concern)

shibuiwilliam: fixed
https://github.com/launchableinc/cli/pull/536/commits/00c4839cf809f6a07b69254fd5c74ac481836157

Konboi (approved):

PR #539, opened 2023-02-15
fix duration to be always positive or 0

shibuiwilliam:
# What
- Add if-else statement to make the duration always positive or 0.0 if it happens to be negative.


Konboi (approved):
LGTM

PR #541, opened 2023-02-20
add perl profile

shibuiwilliam:
# What
- Add Perl profile to support Perl test.
- The Perl profile is made to take a Junit report produced by [TAP::Harness::JUnit](https://metacpan.org/pod/TAP::Harness::JUnit) with environment variable of `JUNIT_NAME_MANGLE=none`. Otherwise, Launchable cannot support Perl test since the report produced may not contain test file path or duration.

### DIFF
--- a/launchable/test_runners/perl.py
+++ b/launchable/test_runners/perl.py
@@ -17,15 +17,6 @@ def subset(client):
 @click.argument('reports', required=True, nargs=-1)
 @launchable.record.tests
 def record_tests(client, reports):
-    click.echo(
-        click.style(
-            "Perl profile is made to take Junit report produced by "
-            "TAP::Harness::JUnit (https://metacpan.org/pod/TAP::Harness::JUnit) "
-            "with environment variable of JUNIT_NAME_MANGLE=none. "
-            "If you are not using TAP::Harness::JUnit, "
-            "please change your reporting to TAP::Harness::JUnit.",
-            fg="yellow"),
-        err=False)
 
     def path_builder(case: TestCase, suite: TestSuite,
                      report_file: str) -> TestPath:
@@ -43,7 +34,12 @@ def record_tests(client, reports):
         filepath = find_filename()
         if not filepath:
             raise click.ClickException(
-                "No file name found in %s" % report_file)
+                "No file name found in %s."
+                "Perl profile is made to take Junit report produced by "
+                "TAP::Harness::JUnit (https://metacpan.org/pod/TAP::Harness::JUnit) "
+                "with environment variable of JUNIT_NAME_MANGLE=none. "
+                "If you are not using TAP::Harness::JUnit, "
+                "please change your reporting to TAP::Harness::JUnit." % report_file)
 
         # default test path in `subset` expects to have this file name
         test_path = [client.make_file_path_component(filepath)]
--- a/launchable/test_runners/perl.py
+++ b/launchable/test_runners/perl.py
@@ -17,6 +17,16 @@ def subset(client):
 @click.argument('reports', required=True, nargs=-1)
 @launchable.record.tests
 def record_tests(client, reports):
+    click.echo(
+        click.style(
+            "Perl profile is made to take Junit report produced by "
+            "TAP::Harness::JUnit (https://metacpan.org/pod/TAP::Harness::JUnit) "
+            "with environment variable of JUNIT_NAME_MANGLE=none. "
+            "If you are not using TAP::Harness::JUnit, "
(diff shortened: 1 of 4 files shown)


launchable/test_runners/perl.py
```diff
@@ -0,0 +1,58 @@
+import click
+from junitparser import TestCase, TestSuite  # type: ignore
+
+from ..testpath import TestPath
+from . import launchable
+
+
+@launchable.subset
+def subset(client):
```
Konboi, line 9: Just a Question: Why did you adopt this style and don't adopt the below style? 

```
subset = launchable.CommonSubsetImpls(__name__).scan_files('*.t')
```

tests/data/perl/report.xml
```diff
@@ -0,0 +1,39 @@
...
+    ok 1 - a + dddd
+    ok 2 - bb + ccc
+    1..2
+}
+1..2
+</system-out>
+    <testcase name="a + b = ab" classname="t/easy/01_easy.t" time="0.152199029922485" />
+    <testcase name="str_concat {" classname="t/easy/01_easy.t" time="0.000449180603027344" />
+  </testsuite>
+</testsuites>
```
Konboi, line 39: How about adding a failure case, just in case?

tests/test_runners/test_perl.py
```diff
@@ -0,0 +1,35 @@
...
+from pathlib import Path
+from unittest import mock
+
+import responses  # type: ignore
+
+from launchable.utils.session import read_session, write_build
+from tests.cli_test_case import CliTestCase
+
+
+class GoTestTest(CliTestCase):
```
Konboi, line 13: GoTestTest? PerlTest is right.

launchable/test_runners/perl.py
```diff
@@ -0,0 +1,58 @@
...
+                "Perl profile is made to take Junit report produced by "
+                "TAP::Harness::JUnit (https://metacpan.org/pod/TAP::Harness::JUnit) "
+                "with environment variable of JUNIT_NAME_MANGLE=none. "
+                "If you are not using TAP::Harness::JUnit, "
+                "please change your reporting to TAP::Harness::JUnit." % report_file)
+
+        # default test path in `subset` expects to have this file name
+        test_path = [client.make_file_path_component(filepath)]
+        if suite.name:
+            test_path.append({"type": "testsuite", "name": suite.name})
```
Konboi, line 47: Question: Do we need a testsuite data? I don't think there is any problem in adding it but just in case.

launchable/test_runners/perl.py
shibuiwilliam: I see there are some Perl project using `.test` as extension to test file, while most of them use `.t`. I felt better to be able to support both.

tests/test_runners/test_perl.py
shibuiwilliam: Sorry. Fixed.
https://github.com/launchableinc/cli/pull/541/commits/33a9f9b29701794d88b8900f7dbda964ae415392

tests/data/perl/report.xml
shibuiwilliam: Added.
https://github.com/launchableinc/cli/pull/541/commits/33a9f9b29701794d88b8900f7dbda964ae415392

launchable/test_runners/perl.py
shibuiwilliam: Fixed.
https://github.com/launchableinc/cli/pull/541/commits/33a9f9b29701794d88b8900f7dbda964ae415392

launchable/test_runners/perl.py
Konboi: I see, thanks

tests/data/perl/report.xml
```diff
@@ -1,12 +1,18 @@
 <?xml version='1.0' encoding='utf-8'?>
 <testsuites>
-  <testsuite name="t/00_compile.t" errors="0" failures="0" skipped="0" tests="1" time="0.0876882076263428">
-    <system-out>ok 1 - use Example;
-1..1
+  <testsuite name="t/easy/01_easy.t" errors="0" failures="0" skipped="0" tests="2" time="0.157027006149292">
+    <system-out>ok 1 - a + b = ab
```
Konboi, line 4: Should we send system-out data as stdout to Launchable? What do you think?

shibuiwilliam: Perhaps no, since it is not sent in most of the test profiles currently.

Konboi (approved):
LGTM! We finally support Perl 🐪

PR #543, opened 2023-02-24
rename perl profile to prove

shibuiwilliam:
# What
- Rename Perl profile to Prove profile since `prove` is the command for running Perl test.


Konboi (approved):
👍

PR #546, opened 2023-02-27
Support --session-name option to `record session` and `record tests` commands

Konboi:
I supported `--session-name` option to `record session` and `record tests` commands.
Previously, the user needed to pass a test-session-id 
to other machines. The user doesn’t need to pass it if using a `—session-name` option.

# Usage

```
# create build
$  launchable record build --name with-session-name
Launchable recorded 22 commits from repository /Users/yabuki-ryosuke/src/github.com/launchableinc/cli
Launchable recorded build with-session-name to workspace example/test with commits from 1 repository:

| .      | .      | 14c0b210c606bf9a4744167064d282318142511d |

# create session with name
$ launchable record session --build with-session-name --session-name example-name
builds/with-session-name/test_sessions/1065

# record tests with session name
$  launchable record tests --build with-session-name --session-name example-name file ./test-results/*.xml
Launchable recorded tests for build with-session-name (test session 1065) to workspace example/test from 45 files:

|   Files found |   Tests found |   Tests passed |   Tests failed |   Total duration (min) |
|---------------|---------------|----------------|----------------|------------------------|
|            45 |           143 |            142 |              1 |                   0.13 |

Visit https://app.launchableinc.com/organizations/example/workspaces/test/test-sessions/1065 to view uploaded test results (or run `launchable inspect tests --test-session-id 1065`)
```

However, the session name must be unique. So if the user set the same session name, the command will fail like below.

```
$ launchable record session --build example-build  --session-name same-build-name
Usage: launchable record session [OPTIONS]
Try 'launchable record session --help' for help.

Error: This session name (same-build-name) is already used. Please set another name.
```

### DIFF
--- a/launchable/commands/record/session.py
+++ b/launchable/commands/record/session.py
@@ -207,7 +207,7 @@ def add_session_name(
     payload = {
         "name": session_name
     }
-    res = client.request("put", sub_path, payload=payload)
+    res = client.request("patch", sub_path, payload=payload)
 
     if res.status_code == HTTPStatus.NOT_FOUND:
         click.echo(
--- a/launchable/commands/record/session.py
+++ b/launchable/commands/record/session.py
@@ -1,4 +1,5 @@
 import os
+import re
 import sys
 from http import HTTPStatus
 from typing import List, Optional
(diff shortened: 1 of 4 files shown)


launchable/commands/record/session.py
```diff
@@ -149,3 +181,49 @@ def session(
...
+    sub_path = "builds/{}/test_sessions/{}".format(build_name, session_id)
+    payload = {
+        "name": session_name
+    }
+    res = client.request("patch", sub_path, payload=payload)
+
+    if res.status_code == HTTPStatus.NOT_FOUND:
+        click.echo(
+            click.style(
+                "Test session {} was not found. Record session may be failed.".format(session_id),
```
shibuiwilliam, line 215: `Record session may have failed.` might be better, since the `record session` is done before this request.

launchable/commands/record/session.py
```diff
@@ -149,3 +181,49 @@ def session(
...
+                'yellow'),
+            err=True,
+        )
+        sys.exit(1)
+    if res.status_code == HTTPStatus.BAD_REQUEST:
+        click.echo(
+            click.style(
+                "You cannot use test session name {} since it is already used by other test session in your workspace."
```
shibuiwilliam, line 223: [nits]
At this point, the `record session` should be done correctly, and it is failing on adding name. I think it is better to advise the test session is created.
`"You cannot use test session name {} since it is already used by other test session in your workspace. The record session is completed successfully without session name."`

shibuiwilliam (approved):
only nits on click.echo message.

PR #547, opened 2023-03-17
Link to build from CLI output

awilkes:
```
awilkes@MacBook-Pro-2 cli % python3 -m launchable record build --name foo03172
Launchable recorded 2 commits from repository /Users/awilkes/code/launchable/cli
Launchable recorded build foo03172 to workspace launchableinc/awilkes with commits from 1 repository:

| .      | .      | 6ce5dd86412c9c3010803cedeaad21b248143708 |

Visit https://app.launchableinc.com/organizations/launchableinc/workspaces/awilkes/data/builds/1647218 to view this build and its test sessions
```

### DIFF
--- a/tests/commands/record/test_build.py
+++ b/tests/commands/record/test_build.py
@@ -1,25 +1,14 @@
 import json
 import os
-import shutil
-import tempfile
 from unittest import mock
 
 import responses  # type: ignore
 
-from launchable.utils.session import SESSION_DIR_KEY, clean_session_files, read_build
+from launchable.utils.session import read_build
 from tests.cli_test_case import CliTestCase
 
 
 class BuildTest(CliTestCase):
-    def setUp(self):
-        self.dir = tempfile.mkdtemp()
-        os.environ[SESSION_DIR_KEY] = self.dir
-
-    def tearDown(self):
-        clean_session_files()
-        del os.environ[SESSION_DIR_KEY]
-        shutil.rmtree(self.dir)
-
     # make sure the output of git-submodule is properly parsed
     @responses.activate
     @mock.patch.dict(os.environ, {"LAUNCHABLE_TOKEN": CliTestCase.launchable_token})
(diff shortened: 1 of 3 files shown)


kohsuke:
I was surprised that this requires another round-trip. It feels like an oversight in the API design that the act of creating a build doesn't return the ID.

awilkes:
Oh, yes, I forgot about that! Good catch.

Konboi (commented):
> I was surprised that this requires another round-trip. It feels like an oversight in the API design that the act of creating a build doesn't return the ID.

Ah, you're right

I'll take over this PR and fix API side too.

tests/commands/record/test_build.py
```diff
@@ -1,25 +1,14 @@
...
 
 import responses  # type: ignore
 
-from launchable.utils.session import SESSION_DIR_KEY, clean_session_files, read_build
+from launchable.utils.session import read_build
 from tests.cli_test_case import CliTestCase
 
 
 class BuildTest(CliTestCase):
-    def setUp(self):
```
Konboi, line 14: It's already defined in CliTestCase.

https://github.com/launchableinc/cli/blob/7c61265e748f4551e937a295878c427255931434/tests/cli_test_case.py#L31-L33

tests/commands/record/test_build.py
```diff
@@ -1,25 +1,14 @@
...
-        self.dir = tempfile.mkdtemp()
-        os.environ[SESSION_DIR_KEY] = self.dir
-
-    def tearDown(self):
```
Konboi, line 18: Already defined too.
https://github.com/launchableinc/cli/blob/7c61265e748f4551e937a295878c427255931434/tests/cli_test_case.py#L173-L176

awilkes:
Thanks @Konboi!

shibuiwilliam (approved):

PR #549, opened 2023-03-22
support multiple files

Konboi:
The raw profile can't receive multiple report files, so I fixed supporting multiple report files.

### DIFF
--- a/launchable/test_runners/raw.py
+++ b/launchable/test_runners/raw.py
@@ -48,9 +48,9 @@ def subset(client, test_path_file):
 split_subset = launchable.CommonSplitSubsetImpls(__name__, formatter=unparse_test_path, seperator='\n').split_subset()
 
 
-@click.argument('test_result_file', required=True, type=click.Path(exists=True))
+@click.argument('test_result_files', required=True, type=click.Path(exists=True), nargs=-1)
 @launchable.record.tests
-def record_tests(client, test_result_file):
+def record_tests(client, test_result_files):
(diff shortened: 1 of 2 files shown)


launchable/test_runners/raw.py
```diff
@@ -150,8 +150,12 @@ def parse_json(test_result_file: str) -> Generator[CaseEventType, None, None]:
                 test_path, duration_secs, CaseEvent.STATUS_MAP[status],
                 case['stdout'], case['stderr'], created_at)
 
-    if not test_result_file.endswith('.xml'):
-        client.parse_func = parse_json
+    for test_result_file in test_result_files:
+        if not test_result_file.endswith('.xml'):
+            client.parse_func = parse_json
```
shibuiwilliam, line 155: [question]
Don't we have to determine if the file ends with `.json`?
Or is it filtered beforehand?

Konboi: I found out. It's a historical reason.
We support only JSON format at first. After that, we started to support XML file format.
So, if the report file isn't XML, we set a json format parser.

https://github.com/launchableinc/cli/pull/505

shibuiwilliam (approved):
Only one comment.

launchable/test_runners/raw.py
Konboi: We only support JSON and XML, so we don't need to check an extension is `.json` when it isn't `.xml`.

PR #556, opened 2023-04-26
Add stdout/stderr method to parse messages

Konboi:
We had failed to collect stdout and stderr.
So, I checked the test report formats and fixed the codes.

### DIFF
--- a/launchable/commands/record/case_event.py
+++ b/launchable/commands/record/case_event.py
@@ -6,6 +6,8 @@ from junitparser import Error, Failure, Skipped, TestCase, TestSuite  # type: ig
 
 from ...testpath import FilePathNormalizer, TestPath
 
+POSSIBLE_RESULTS = (Failure, Error, Skipped)
+
 CaseEventType = Dict[str, str]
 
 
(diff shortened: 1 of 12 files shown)


launchable/commands/record/case_event.py
```diff
@@ -76,10 +78,50 @@ def path_canonicalizer(test_path: TestPath) -> TestPath:
...
+
+            """
+            case for:
+                <testcase>
+                <failure message="...>...</failure>
+                </testcase>
+            """
+            stderr = ""
+            for result in case.result:
+                if any(isinstance(result, r) for r in POSSIBLE_RESULTS):
```
shibuiwilliam, line 111: [nits]
This can be written like this:
```python
if type(result) in POSSIBLE_RESULTS:
```

Anyway, either way is fine.

launchable/commands/record/case_event.py
```diff
@@ -76,10 +78,50 @@ def path_canonicalizer(test_path: TestPath) -> TestPath:
...
+                    if result.message and result.text:
+                        stderr = stderr + result.message + "\n" + result.text
+                    elif result.message:
+                        stderr = stderr + result.message
+                    elif result.text:
+                        stderr = stderr + result.text
```
shibuiwilliam, line 117: [question]
Is it correct that this is appending result.message and result.text if they exist?
Let's say you have 3 results, and they all have result.message, like "aa", "bb" and "cc" respectively.
Does the result be like this below?
1st stderr result: aa
2nd stderr result: aabb
3rd stderr result: aabbcc
Or, do you always have line breaks after message that you don't have to take care of separating messages?

shibuiwilliam (commented):
Wrote some comments.

launchable/commands/record/case_event.py
Konboi: > Or, do you always have line breaks after message that you don't have to take care of separating messages?

Yes, I checked some test runner reports, then reports are usually put "\n" at last. 
So, I didn't care it for now. If we faced some problems, I'll fix this then.

shibuiwilliam: If that is the case, why do you need a line break in `result.message + "\n" + result.text` ?

tests/data/pytest/record_test_result.json
```diff
@@ -1,168 +1,168 @@
...
+           "type": "testcase",
+           "name": "test_func1"
+         }
+       ],
+       "duration": 0.001,
+       "status": 1,
+       "stdout": "",
+       "stderr": "",
+       "created_at": "2021-12-09T14:09:43.746900",
+       "data": null
```
shibuiwilliam, line 70: Are these spaces just resulted from formatting?

launchable/commands/record/case_event.py
Konboi: Sorry, I misunderstood that you mention only result.text. 

Test runner which I checked usually produces stderr data like below
```
<testcase>
  <failure message="FAIL_RESULT"> STACK TRACE ... </failure>
</testcase>
```

We can get message data (="FAILE_RESULT") use by result.message method. Also, we can get failure's text(=STACK TRACE...) use by result.text.

Usually, message data doesn't include "\n", so I join message and text with "\n".

After writing the comment, I realized I should change to `stderr = stderr + result.message + "\n"`.
So, I'll change only it.

tests/data/pytest/record_test_result.json
Konboi: yes, except for added stderr

shibuiwilliam: ok

shibuiwilliam (approved):

awilkes:
Thank you @Konboi!

PR #563, opened 2023-05-18
Add split-subset support to dotnet profile

ghost:


launchable/test_runners/dotnet.py
```diff
@@ -53,6 +53,40 @@ def exclusion_output_handler(subset_tests: List[TestPath], rest_tests: List[Test
     client.run()
 
 
+@launchable.split_subset
+def split_subset(client):
```
Konboi, line 57: Thank you, I really forgot to implement it 🤦‍♂️

PR #564, opened 2023-05-18
Better, more consistent option name.

kohsuke:
Kept the historical one for backward compatibility.


Konboi (approved):

kohsuke:
Oops, thanks for the follow up fix!

PR #565, opened 2023-05-18
Add --output-exclusion-rules support to split-subset

awilkes:
* Add `--output-exclusion-rules` support to split-subset
* Allow split-subset to output empty rest files (like regular subset)
* Add split-subset support to dotnet profile


Konboi (approved):
LGTM 💯

PR #570, opened 2023-07-03
Fix way of collecting git branch names

Konboi:
We used `git rev-parse --abbrev-ref HEAD` to collect a git branch name. However, in detached HEAD case, this command couldn't get a collect branch name and got `HEAD`.  e.g) https://stackoverflow.com/questions/6059336/how-to-find-the-current-git-branch-in-detached-head-state

So, we changed to use environment variables that each CI provider provides and `git show-ref | grep '^'$(git rev-parse HEAD)`

### DIFF
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -1,13 +1,13 @@
 import os
 import re
 import sys
-from typing import List
+from typing import List, Optional
 
 import click
 from tabulate import tabulate
 
 from launchable.utils.key_value_type import normalize_key_value_types
-from launchable.utils.link import LinkKind, capture_link
+from launchable.utils.link import CIRCLECI_KEY, GITHUB_ACTIONS_KEY, JENKINS_URL_KEY, LinkKind, capture_link
 
 from ...utils import subprocess
 from ...utils.authentication import get_org_workspace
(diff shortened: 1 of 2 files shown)


launchable/commands/record/build.py
```diff
@@ -248,3 +248,46 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
             ))
 
     write_build(build_name)
+
+
+def _get_branch_name(repo_dist: str) -> Optional[str]:
```
shibuiwilliam, line 253: [question]
When does this return None?
I see the code may return "", but cannot determine None.

shibuiwilliam (commented):
one question.

launchable/commands/record/build.py
Konboi: I know your opinion  but if I changed `Optional[str]` to `str` , these errors will happen

```
$ pipenv run type
launchable/commands/record/build.py:260: error: Incompatible return value type (got "Optional[str]", expected "str")  [return-value]
launchable/commands/record/build.py:262: error: Incompatible return value type (got "Optional[str]", expected "str")  [return-value]
launchable/commands/record/build.py:267: error: Incompatible return value type (got "Optional[str]", expected "str")  [return-value]
launchable/commands/record/build.py:269: error: Incompatible return value type (got "Optional[str]", expected "str")  [return-value]
launchable/commands/record/build.py:274: error: Incompatible return value type (got "Optional[str]", expected "str")  [return-value]
```

Konboi: Which is better 1. Keep Optional[str] 2. Change to `return os.environ.get(JENKINS_GIT_BRANCH_KEY, "")`

shibuiwilliam: Since you are checking the env var existence in the code, then I think you can just return `os.environ["KEY"]`.
This should return `str` for value of env var dict.

I suppose it is better to avoid None (or nil in Go, Null in Java) than accept None if possible.

Konboi: Thank you! I fixed it https://github.com/launchableinc/cli/pull/570/commits/e448da1b6ae1a4b6287ca9eb951c2f774ca50c1d

shibuiwilliam (approved):

basil:
Causes #581.

PR #572, opened 2023-07-11
fix prove test report

shibuiwilliam:
# What
- This is to remove unnecessary prefix like `1 - ` that happens to be added to the test names produced by [TAP-Formatter-JUnit](https://github.com/bleargh45/TAP-Formatter-JUnit) for Perl Prove.

### DIFF
--- a/launchable/commands/record/case_event.py
+++ b/launchable/commands/record/case_event.py
@@ -119,10 +119,14 @@ class CaseEvent:
             return stderr
 
         return CaseEvent.create(
-            path_canonicalizer(path_builder(case, suite, report_file)), case.time, status,
+            path_canonicalizer(path_builder(case, suite, report_file)),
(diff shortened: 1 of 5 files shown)


launchable/commands/record/case_event.py
```diff
@@ -119,10 +119,14 @@ def stderr(case: TestCase) -> str:
...
         return CaseEvent.create(
-            path_canonicalizer(path_builder(case, suite, report_file)), case.time, status,
+            path_canonicalizer(path_builder(case, suite, report_file)),
+            case.time,
+            status,
             stdout(case),
             stderr(case),
-            suite.timestamp, data)
+            suite.timestamp,
+            data,
```
Konboi, line 128: note for me: just formatted

Konboi (approved):

PR #574, opened 2023-07-12
Validate if the repository uses shallow clones

ono-max:
Currently, we send one commit which has no parents when the repository is a shallow clone. However, we can't collect all commits data in this case. Thus, we need to ask users to use full clone or disable commit collection by --no-commit-collection option. I added this feature in this PR.

## Example

```shell
$ /Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/bin/launchable --log-level=audit record build --name "testbuild" --max-days 90
Can't collect all commits from /Users/ono-max/workspace/example since it is the shallow repository. Please use full clone or disable commit collection by --no-commit-collection option.
```

### DIFF
--- a/launchable/commands/record/commit.py
+++ b/launchable/commands/record/commit.py
@@ -62,7 +62,7 @@ def commit(ctx, source: str, executable: bool, max_days: int, scrub_pii: bool, i
             universal_newlines=True).strip()
         if is_shallow == "true":
             click.echo(click.style(
-                "Can't collect all commits from {} since it is the shallow repository. "
+                "Can't collect commit history from {} since it is the shallow repository. "
                 "Please use full clone or disable commit collection by --no-commit-collection option."
                 .format(cwd),
                 fg='yellow'),
--- a/launchable/commands/record/commit.py
+++ b/launchable/commands/record/commit.py
@@ -53,8 +53,22 @@ def commit(ctx, source: str, executable: bool, max_days: int, scrub_pii: bool, i
         _import_git_log(import_git_log_output, ctx.obj.dry_run)
         return
 
+    cwd = os.path.abspath(source)
     try:
-        exec_jar(os.path.abspath(source), max_days, ctx.obj.dry_run)
+        is_shallow = subprocess.check_output(
+            ['git', 'rev-parse', '--is-shallow-repository'],
+            stderr=subprocess.DEVNULL,
+            cwd=cwd,
+            universal_newlines=True).strip()
+        if is_shallow == "true":
+            click.echo(click.style(
+                "Can't collect all commits from {} since it is the shallow repository. "
+                "Please use full clone or disable commit collection by --no-commit-collection option."
+                .format(cwd),
+                fg='yellow'),
+                err=True)
+            sys.exit(1)
+        exec_jar(cwd, max_days, ctx.obj.dry_run)
     except Exception as e:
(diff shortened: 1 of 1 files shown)


launchable/commands/record/commit.py
```diff
@@ -53,16 +53,30 @@ def commit(ctx, source: str, executable: bool, max_days: int, scrub_pii: bool, i
...
+            ['git', 'rev-parse', '--is-shallow-repository'],
+            stderr=subprocess.DEVNULL,
+            cwd=cwd,
+            universal_newlines=True).strip()
+        if is_shallow == "true":
+            click.echo(click.style(
+                "Can't collect all commits from {} since it is the shallow repository. "
+                "Please use full clone or disable commit collection by --no-commit-collection option."
+                .format(cwd),
+                fg='yellow'),
```
Konboi, line 68: We'll exit as code=1 so, should use error color of `fg='red'` than warning color of `fg='yellow'`

launchable/commands/record/commit.py
```diff
@@ -53,16 +53,30 @@ def commit(ctx, source: str, executable: bool, max_days: int, scrub_pii: bool, i
...
     try:
-        exec_jar(os.path.abspath(source), max_days, ctx.obj.dry_run)
+        is_shallow = subprocess.check_output(
+            ['git', 'rev-parse', '--is-shallow-repository'],
+            stderr=subprocess.DEVNULL,
+            cwd=cwd,
+            universal_newlines=True).strip()
+        if is_shallow == "true":
+            click.echo(click.style(
+                "Can't collect all commits from {} since it is the shallow repository. "
```
Konboi, line 65: nits: we don't collect all commits.  I think `Can't collect commit history from ...` is better. WDYT???

launchable/commands/record/commit.py
```diff
@@ -53,16 +53,30 @@ def commit(ctx, source: str, executable: bool, max_days: int, scrub_pii: bool, i
...
+                "Please use full clone or disable commit collection by --no-commit-collection option."
```
Konboi, line 66: `use full clone` is a little strange for me.  `do full clone`...? but I'm not familiar with English...

Konboi (approved):
LGTM and I could check this validation!
https://github.com/launchableinc/cli/actions/runs/5537799118/jobs/10107032456
image

But I made a few nits comments, please check it

launchable/commands/record/commit.py
ono-max: I referred use full clone from https://docs.aws.amazon.com/codepipeline/latest/userguide/tutorials-github-gitclone.html. What do you think?

Konboi: Nice, so you don't need to update it!

launchable/commands/record/commit.py
ono-max: [Update error messages](https://github.com/launchableinc/cli/pull/574/commits/67a499f6784af7076450df95ef792ca541f32aa6)

launchable/commands/record/commit.py
ono-max: [Update message color](https://github.com/launchableinc/cli/pull/574/commits/b90aed90e6d82f6f0e38f7b17d4e229c21333f65)

Konboi (approved):

PR #576, opened 2023-07-12
Specify the version of importlib-metadata

ono-max:
Currently, python lint fails on the main branch such as https://github.com/launchableinc/cli/actions/runs/5528890710/jobs/10086219577. This PR fixes it

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -15,7 +15,7 @@ wheel = "*"
 # 2.7.0, < 2.8.0". The latest autopep8 specifies "pycodestyle >= 2.8.0". This
 # conflict cannot be resolved. Pin the version to resolve this.
 autopep8 = "<=1.5.7"
-importlib-metadata = "*"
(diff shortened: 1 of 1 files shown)


Pipfile
```diff
@@ -15,7 +15,7 @@ wheel = "*"
 # 2.7.0, < 2.8.0". The latest autopep8 specifies "pycodestyle >= 2.8.0". This
 # conflict cannot be resolved. Pin the version to resolve this.
 autopep8 = "<=1.5.7"
-importlib-metadata = "*"
+importlib-metadata = "<5.0"
```
ono-max, line 18: Looks like there is a breaking change in https://importlib-metadata.readthedocs.io/en/latest/history.html#v5-0-0.

Konboi:
@shibuiwilliam Could you double check, just in case?

shibuiwilliam (approved):
Let's make a try.

PR #585, opened 2023-07-14
Run CI daily

ono-max:
The CI may be broken someday. We need to check the status of CI by running it daily.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -10,8 +10,8 @@ on:
   pull_request:
     branches: [ main ]
   schedule:
-    # This job runs at 00:00 JST on Friday.
-    - cron: '0 0 * * *'
+    # This job runs at 00:00 JST every day.
+    - cron: '0 9 * * *'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/python-package.yml
```diff
@@ -9,6 +9,9 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+  schedule:
+    # This job runs at 00:00 JST on Friday.
+    - cron: '0 0 * * *'
```
shibuiwilliam, line 14: [must]
I believe Github Actions timezone is UTC, and you need to add 9 hours to make it JST.
Something like `0 9 * * 5`.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

ono-max: Oh, my comment is wrong. I copied and pasted it from another one I wrote before. Let me modify it. Thank you!

shibuiwilliam (approved):

PR #587, opened 2023-07-18
Handle exceptions in requests package

ono-max:
Basically, we should not stop users' CI pipeline. Thus, we should use exit code 0 unless it is an intentional notification. This PR fixes it by catching exception in client.request.

### DIFF
--- a/launchable/commands/record/session.py
+++ b/launchable/commands/record/session.py
@@ -142,7 +142,6 @@ def session(
                 raise e
             else:
                 click.echo(e, err=True)
-                return
 
     flavor_dict = {}
     for f in normalize_key_value_types(flavor):
--- a/launchable/commands/record/session.py
+++ b/launchable/commands/record/session.py
@@ -127,11 +127,22 @@ def session(
 
     if session_name:
         sub_path = "builds/{}/test_session_names/{}".format(build_name, session_name)
-        res = client.request("get", sub_path)
+        try:
+            res = client.request("get", sub_path)
 
-        if res.status_code != 404:
-            raise click.UsageError(
-                'This session name ({}) is already used. Please set another name.'.format(session_name))
+            if res.status_code != 404:
+                click.echo(click.style(
+                    "This session name ({}) is already used. Please set another name."
+                    .format(session_name),
+                    fg='red'),
+                    err=True)
+                sys.exit(2)
+        except Exception as e:
+            if os.getenv(REPORT_ERROR_KEY):
+                raise e
+            else:
+                click.echo(e, err=True)
+                return
 
     flavor_dict = {}
     for f in normalize_key_value_types(flavor):
--- a/launchable/commands/record/session.py
+++ b/launchable/commands/record/session.py
@@ -9,7 +9,7 @@ import click
 from launchable.utils.key_value_type import normalize_key_value_types
 from launchable.utils.link import LinkKind, capture_link
 
-from ...utils.click import KeyValueType
+from ...utils.click import KeyValueType, ignorable_error
 from ...utils.env_keys import REPORT_ERROR_KEY
 from ...utils.http_client import LaunchableClient
 from ...utils.no_build import NO_BUILD_BUILD_NAME
@@ -141,7 +141,7 @@ def session(
             if os.getenv(REPORT_ERROR_KEY):
                 raise e
             else:
-                click.echo(e, err=True)
+                click.echo(ignorable_error(e))
 
     flavor_dict = {}
     for f in normalize_key_value_types(flavor):
--- a/launchable/utils/click.py
+++ b/launchable/utils/click.py
@@ -169,5 +169,6 @@ def convert_to_seconds(s: str):
 
     return float(duration)
 
+
 def ignorable_error(e: Exception):
     return "An error occurred on Launchable CLI. You can ignore this message since the process will continue. Error: {}".format(e)
--- a/launchable/utils/click.py
+++ b/launchable/utils/click.py
@@ -168,3 +168,6 @@ def convert_to_seconds(s: str):
         duration += int(val) * u
 
     return float(duration)
+
+def ignorable_error(e: Exception):
+    return "An error occurred on Launchable CLI. You can ignore this message since the process will continue. Error: {}".format(e)
--- a/tests/commands/test_api_error.py
+++ b/tests/commands/test_api_error.py
@@ -290,7 +290,7 @@ class APIErrorTest(CliTestCase):
                 build=self.build_name,
                 session_id=self.session_id),
             body=ReadTimeout("error"))
-        with tempfile.NamedTemporaryFile() as rest_file:
+        with tempfile.NamedTemporaryFile(delete=False) as rest_file:
             self.assert_exit_code(["subset",
                                    "--target",
                                    "30%",
--- a/tests/commands/test_api_error.py
+++ b/tests/commands/test_api_error.py
@@ -197,7 +197,7 @@ class APIErrorTest(CliTestCase):
 
         with tempfile.NamedTemporaryFile(delete=False) as rest_file:
             self.assert_exit_code(["subset", "--target", "30%", "--session", self.session, "--rest", rest_file.name,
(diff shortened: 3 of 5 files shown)


launchable/commands/record/session.py
```diff
@@ -118,11 +118,21 @@ def session(
...
+                    "This session name ({}) is already used. Please set another name."
+                    .format(session_name),
+                    fg='red'),
+                    err=True)
+                sys.exit(2)
+        except Exception as e:
+            if os.getenv(REPORT_ERROR_KEY):
+                raise e
+            else:
+                click.echo(e, err=True)
```
Konboi, line 135: When an error happens and just an error is printed, the user will be troubled. 

Also, customers will ask us for an error then, it’s difficult to detect the cause with only an error message, I think.

So I think we should tell that why this message is printed.
e.g) `An error occured on Launchable side, verify request failed. However, the process will continue. Error: {}`

What do you think??

ono-max: ## Before

```shell
$ /Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/bin/launchable record session --build test --session-name hoge
unable to post to hogehoge/intake/organizations/launchableinc/workspaces/mothership/builds/test/test_session_names/hoge
unable to post to hogehoge/intake/organizations/launchableinc/workspaces/mothership/builds/test/test_sessions
cannot access local variable 'session_id' where it is not associated with a value
```

## After

```shell
$ /Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/bin/launchable record session --build test --session-name hoge
An error occurred on Launchable CLI. You can ignore this message since the process will continue. Error: unable to post to hogehoge/intake/organizations/launchableinc/workspaces/mothership/builds/test/test_session_names/hoge
unable to post to hogehoge/intake/organizations/launchableinc/workspaces/mothership/builds/test/test_sessions
cannot access local variable 'session_id' where it is not associated with a value
```

ono-max:
@Konboi 

Updated. Thank you!

launchable/utils/click.py
```diff
@@ -168,3 +168,7 @@ def convert_to_seconds(s: str):
         duration += int(val) * u
 
     return float(duration)
+
+
+def ignorable_error(e: Exception):
```
Konboi, line 173: 👍

tests/commands/test_api_error.py
Konboi, line 1: Could you add a usual use case test?
Now, we're testing per endpoint. But our cli needs to work a whole process without exist when Launchable is down.
So, I think we should check an actual use case. 
e.g.)launchable verify -> launchable record build -> launchable -> subset -> launchable record tests

ono-max: Sorry, I can't get what you're saying well.
Do you mean https://github.com/launchableinc/cli/blob/main/.github/workflows/e2e.yml?

Konboi: I wanted to say add a test like below

```py
def test_example_usecase(self):
        # verify
        responses.add(
            responses.GET,
            "{base}/intake/organizations/{org}/workspaces/{ws}/verification".format(
                base=get_base_url(),
                org=self.organization,
                ws=self.workspace),
            body=ReadTimeout("error"))

        build = "internal_server_error"
        responses.add(
            responses.POST,
            "{base}/intake/organizations/{org}/workspaces/{ws}/builds/{build}/test_sessions".format(
                base=get_base_url(),
                org=self.organization,
                ws=self.workspace,
                build=build),
            status=500)

       # record build
        success_server = HTTPServer(("", 0), SuccessCommitHandlerMock)
        thread = threading.Thread(None, success_server.serve_forever)
        thread.start()
        host, port = success_server.server_address
        endpoint = "http://{}:{}".format(host, port)

        responses.add(responses.POST, "{base}/intake/organizations/{org}/workspaces/{ws}/builds".format(
            base=get_base_url(), org=self.organization, ws=self.workspace), status=500)
        result = self.cli("record", "build", "--name", "example")

        # subset
        responses.replace(
            responses.POST,
            "{base}/intake/organizations/{org}/workspaces/{ws}/subset".format(
                base=get_base_url(),
                org=self.organization,
                ws=self.workspace),
            status=500)


        result = self.cli("verify")

        result = self.cli("launchable record build --name example")
        self.assertEqual(result.exception, None)
        success_server.shutdown()
        thread.join()

       result = self.cli("launchable subset --target 50%")
```

Konboi: Can you understand??

tests/commands/test_api_error.py
```diff
@@ -249,3 +249,60 @@ def test_record_tests(self):
 
         result = self.cli("record", "tests", "--session", self.session, "minitest", str(self.test_files_dir) + "/")
         self.assertEqual(result.exit_code, 0)
+
+    @responses.activate
+    @mock.patch.dict(os.environ, {"LAUNCHABLE_TOKEN": CliTestCase.launchable_token})
+    def test_all_workflow_when_server_down(self):
+        responses.add(
```
Konboi, line 256: Question: Don't we set a status code?

tests/commands/test_api_error.py
```diff
@@ -249,3 +249,60 @@ def test_record_tests(self):
...
+        responses.replace(
+            responses.GET,
+            "{base}/intake/organizations/{org}/workspaces/{ws}/builds/{build}/test_sessions/{session_id}".format(
+                base=get_base_url(),
+                org=self.organization,
+                ws=self.workspace,
+                build=self.build_name,
+                session_id=self.session_id),
+            body=ReadTimeout("error"))
+        with tempfile.NamedTemporaryFile(delete=False) as rest_file:
```
Konboi, line 292: Why don't you delete temp file??

tests/commands/test_api_error.py
```diff
@@ -249,3 +249,60 @@ def test_record_tests(self):
 
         result = self.cli("record", "tests", "--session", self.session, "minitest", str(self.test_files_dir) + "/")
         self.assertEqual(result.exit_code, 0)
+
+    @responses.activate
+    @mock.patch.dict(os.environ, {"LAUNCHABLE_TOKEN": CliTestCase.launchable_token})
```
Konboi, line 254: nits: If possible, can the test part `self.cli(...)` and `self.assertEqual(...)` be summarized below?

Konboi (commented):
Thank you! 
Almost OK, but I made some comments. Please check it.

tests/commands/test_api_error.py
ono-max: We don't need it since the exception(ReadTimeout) is occurred in `client.request` method. Thus, we can't get a status code in this case even though we set it.

tests/commands/test_api_error.py
ono-max: Good catch. Thank you.

ono-max: Well, it looks like the error occurs if we try to temp file. https://app.launchableinc.com/organizations/launchableinc/workspaces/cli/data/test-sessions/2068736

Other test cases also set `delete=False` in test_api_error.py when creating temp files.

ono-max: I added the comment.

tests/commands/test_api_error.py
Konboi: Sorry I wanted to say that I want to divide into setup section and test section

```
# set up section
response.add(...) # verify
response.add(...) # record build
response.add(...) # subset
response.add(..) # subset

# test section

verify = self.cli("verify")
self.assertEqual(verify.exit_code, 0)

record_build = self.cli("record build --name xxx")
self.assertEqual(record_build.exit_code, 0)
...
```

tests/commands/test_api_error.py
Konboi: OK, I could confirm it locally 👍

Konboi (commented):
Sorry, I commented one point

ono-max:
Rebased to resolve conflicts

ono-max:
@Konboi 

I've updated 👍 Thank you.

Konboi (approved):

PR #589, opened 2023-07-20
Support --branch name option when --no-commit-collection option is enabled

Konboi:
The cli collects branch names from git repository information. However, if the `--no-commit-collection` is enabled, cannot collect a branch name.
So, we add a new option `--branch` name.

e.g.)

```sh
$ launchable record build --name <BUILD NAME> \
--no-commit-collection \
--commit a-repo=abc123 \ 
--commit b-repo=56cde \
--branch a-repo=v1.5.2 \
--branch b-repo=v.2.2.4
```

### DIFF
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -82,9 +82,17 @@ CODE_BUILD_WEBHOOK_HEAD_REF_KEY = "CODEBUILD_WEBHOOK_HEAD_REF"
     default=[],
     cls=KeyValueType,
 )
+@click.option(
+    '--branch',
+    'branches',
+    help="Set repository name and branch name when you use --no-commit-collection option. Please use the same repository name with a commit option",  # noqa: E501
+    multiple=True,
+    default=[],
+    cls=KeyValueType,
+)
 @click.pass_context
 def build(ctx: click.core.Context, build_name: str, source: List[str], max_days: int, no_submodules: bool,
-          no_commit_collection: bool, scrub_pii: bool, commits: List[str], links: List[str]):
+          no_commit_collection: bool, scrub_pii: bool, commits: List[str], links: List[str], branches: List[str]):
 
     if "/" in build_name or "%2f" in build_name.lower():
         sys.exit("--name must not contain a slash and an encoded slash")
@@ -177,6 +185,9 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
             submodules.append((repo_name, "", hash))
         if invalid:
             sys.exit(1)
+
+    if no_commit_collection and len(branches) != 0:
+        branch_name_map = dict(normalize_key_value_types(branches))
 
     # Note: currently becomes unique command args and submodules by the hash.
     # But they can be conflict between repositories.
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -188,8 +188,8 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
 
     # Note: currently becomes unique command args and submodules by the hash.
     # But they can be conflict between repositories.
-    uniq_submodules = {commit_hash: (name, repo_dist, commit_hash)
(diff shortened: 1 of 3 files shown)


launchable/commands/record/build.py
```diff
@@ -183,6 +191,17 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
     uniq_submodules = {commit_hash: (name, repo_dist, commit_hash)
                        for name, repo_dist, commit_hash, in sources + submodules}.values()
 
+    if no_commit_collection and len(branches) != 0:
+        branch_name_map = dict(normalize_key_value_types(branches))
```
shibuiwilliam, line 195: [must]
If you want to use `normalize_key_value_types` here, please fix the comment in the method that is saying it is for flavor.
https://github.com/launchableinc/cli/blob/4344bd0463fdc4fc5a12bfa3e1d86232be2f4d9c/launchable/utils/key_value_type.py#L4

launchable/commands/record/build.py
```diff
@@ -183,6 +191,17 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
     uniq_submodules = {commit_hash: (name, repo_dist, commit_hash)
                        for name, repo_dist, commit_hash, in sources + submodules}.values()
```
shibuiwilliam, line 192: [question]
Do you have a specific reason you are making dict and then dict_values?
Can you initialize as list without going through dict, for readability?

[nits]
make this list instead of dict_values.
```python
uniq_submodules = list({commit_hash: (name, repo_dist, commit_hash)
                       for name, repo_dist, commit_hash, in sources + submodules}.values())
```

launchable/commands/record/build.py
```diff
@@ -183,6 +191,17 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
...
 
+    if no_commit_collection and len(branches) != 0:
+        branch_name_map = dict(normalize_key_value_types(branches))
+        repo_names = [m[0] for m in uniq_submodules]
+        not_match_branch_repos = set(branch_name_map.keys()) - set(repo_names)
+        if len(not_match_branch_repos) > 0:
+            click.echo(
+                click.style(
+                    "--branch option has been set for {} but not set in the --commit option".format(not_match_branch_repos),
```
shibuiwilliam, line 201: [must]
`--branch option is set for {} but was not set for the --commit option`

shibuiwilliam, line 201: [must]
`--branch option is set for {} but is not set for the --commit option`

shibuiwilliam (commented):
commented

launchable/commands/record/build.py
Konboi: > Do you have a specific reason you are making dict and then dict_values?

To be honest, I'm not original author of it, so, I don't know the reason.

> Can you initialize as list without going through dict, for readability?

In my opinion, if we passed the test and don't change the request payload format, we can change it.

launchable/utils/key_value_type.py
```diff
@@ -2,10 +2,10 @@
 
 
 def normalize_key_value_types(kv_types_raw: List[str]) -> List[Tuple[str, str]]:
-    """Normalize flavor list, because the type of the flavor list specified in the command line flags differs depending
+    """Normalize key/value list, because the type of the key/value list specified in the command line flags differs depending
     on the version of the click.
 
-    TODO: handle extraction of flavor tuple to dict in better way for >=click8.0 that returns tuple of tuples as tuple
+    TODO: handle extraction of fkey-value tuple to dict in better way for >=click8.0 that returns tuple of tuples as tuple
```
shibuiwilliam, line 8: [typo]
x fkey-value
o key-value

shibuiwilliam (commented):
only fix one typo and I will approve.

launchable/utils/key_value_type.py
Konboi: fixe

shibuiwilliam (approved):

PR #592, opened 2023-07-28
Work around a bug in NUnitXML.Logger mishandling nested type

kohsuke:
* Added nested type to a test case based on NUnit console reporter to make sure it's producing report XML that we expect.
* Added a captured broken test report from launchableinc/nunit that shows how NUnitXML.Logger mishandles nested type
* Improved the nunit profile & the dotnet profile to handle this broken XML in the least invasive way possible
* Added test case to prove the handling of this case.


Konboi (approved):
LGTM

PR #595, opened 2023-08-02
Automate creating tag in releasing process

ono-max:
Currently, we create a tag manually when releasing new changes. We can automate the process by using https://github.com/Songmu/tagpr.

When merging PR, the generated PR by tagpr is created, e.g. https://github.com/ono-max/vscode-launchable/pull/2. We can release new changes only by merging it in the new releasing process.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -5,11 +5,26 @@ name: Upload Python Package
 
 on:
   workflow_dispatch:
-  release:
-    types: [published]
+  push:
+    branches:
+    - main
 
 jobs:
+  tagpr:
+    permissions:
+      actions: write
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@96f53100ba2a5449eb71d2e6604bbcd94b9449b5
+    - uses: Songmu/tagpr@43d52e123cf8d55db9d602601f115f530588e2f8
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
   deploy:
+    needs: tagpr
+    if: needs.tagpr.outputs.tag != ''
 
     runs-on: ubuntu-20.04
 
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -17,8 +17,8 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@96f53100ba2a5449eb71d2e6604bbcd94b9449b5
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -5,11 +5,26 @@ name: Upload Python Package
...
+    - uses: actions/checkout@96f53100ba2a5449eb71d2e6604bbcd94b9449b5
+    - uses: Songmu/tagpr@43d52e123cf8d55db9d602601f115f530588e2f8
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
   deploy:
+    needs: tagpr
+    if: needs.tagpr.outputs.tag != ''
 
     runs-on: ubuntu-20.04
```
shibuiwilliam, line 29: [question]
Just curious why the tagpr runs on ubuntu-latest while deploy is using ubuntu-20.04?

ono-max: I don't have a reason. Since the environment of tagpr is different from the environment for releasing CLI, I thought that it's no problem.

ono-max: Should I match the version of them?

shibuiwilliam: ok

.github/workflows/python-publish.yml
```diff
@@ -5,11 +5,26 @@ name: Upload Python Package
...
 jobs:
+  tagpr:
+    permissions:
+      actions: write
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@96f53100ba2a5449eb71d2e6604bbcd94b9449b5
+    - uses: Songmu/tagpr@43d52e123cf8d55db9d602601f115f530588e2f8
```
Konboi, line 21: Could you comment these actions version?

.github/workflows/python-publish.yml
```diff
@@ -5,11 +5,26 @@ name: Upload Python Package
 
 on:
   workflow_dispatch:
-  release:
-    types: [published]
```
Konboi, line 9: Don't we need to separate workflow files? just in case.

ono-max: I think that it's not possible to separate workflow into files.

.github/workflows/python-publish.yml
ono-max: [08bcc0c](https://github.com/launchableinc/cli/pull/595/commits/08bcc0c9b7641ce9330eaff01dca1c688e8b858c)

.github/workflows/python-publish.yml
Konboi: OK

Konboi (approved):

PR #599, opened 2023-08-08
Add the ability to publish CLI with workflow_dispatch

ono-max:
I confirmed this works in https://github.com/ono-max/vscode-launchable/commit/37519472ad111fc22e8eeb12527e24bca612c19a.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -24,7 +24,7 @@ jobs:
 
   deploy:
     needs: tagpr
-    if: needs.tagpr.outputs.tag != ''
+    if: needs.tagpr.outputs.tag != '' || github.event_name == 'workflow_dispatch'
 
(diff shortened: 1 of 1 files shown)


Konboi:
In my understanding `'workflow_dispatch'` is for manual execution.
Why do you need this if statement??

ono-max:
There are two reasons.

1. Since setting for tagpr is wrong, I need to release https://github.com/launchableinc/cli/releases/tag/v1.71.0 manually now.
2. tagpr might be broken if Github is not stable. We'll need to publish it manually in this case.

Konboi (approved):
> Since setting for tagpr is wrong, I need to release https://github.com/launchableinc/cli/releases/tag/v1.71.0 manually now.

I got it

ono-max:
Thank you!

PR #600, opened 2023-08-08
Define outputs for retrieving tagpr's output

ono-max:
Although we can get output between steps, it's not possible to get output directory between jobs. Thus, we need to use `jobs.<job_id>.outputs`.

Ref: https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

I confirmed that this works well by releasing my extension with https://github.com/ono-max/vscode-launchable/commit/e3f27aa202cc6788912355dfef3734694463ddc5.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -16,9 +16,12 @@ jobs:
       contents: write
       pull-requests: write
     runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.run-tagpr.outputs.tag }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -16,9 +16,12 @@ jobs:
       contents: write
       pull-requests: write
     runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.run-tagpr.outputs.tag }}
```
Konboi, line 20: memo:
```yaml
    needs: tagpr
    if: needs.tagpr.outputs.tag != ''
```

OK

Konboi (approved):

PR #605, opened 2023-08-08
Update discord messages

ono-max:
Currently discord workflow is failed in https://github.com/launchableinc/cli/actions/runs/5793863689/job/15702394002#step:9:15 since the triggered event name for releasing is changed in https://github.com/launchableinc/cli/pull/595. Thus, we need to update the message in discord workflow.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -66,4 +66,4 @@ jobs:
       env:
         DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
       with:
-        args: 'Launchable CLI {{ EVENT_PAYLOAD.release.tag_name }} is released! {{ EVENT_PAYLOAD.release.html_url }}'
+        args: 'Launchable CLI is released!'
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -66,4 +66,4 @@ jobs:
       env:
         DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
       with:
-        args: 'Launchable CLI is released!'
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -66,4 +66,4 @@ jobs:
       env:
         DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
       with:
-        args: 'Launchable CLI {{ EVENT_PAYLOAD.release.tag_name }} is released! {{ EVENT_PAYLOAD.release.html_url }}'
```
Konboi, line 69: Can't we use `needs.tagpr.outputs.tag`?

Only 'Launchable CLI is released!` is no meaning...

e.g.) current message format
image

ono-max: I checked `needs.tagpr.outputs.tag` works 👍 

Ref:
https://github.com/ono-max/vscode-launchable/blob/main/.github/workflows/release.yaml#L34
https://github.com/ono-max/vscode-launchable/actions/runs/5794230906/job/15703421203#step:4:5

.github/workflows/python-publish.yml
```diff
@@ -66,4 +66,4 @@ jobs:
...
+        args: 'Launchable CLI {{ needs.tagpr.outputs.tag }} is released!'
```
Konboi, line 69: I think we can replace from `{{ EVENT_PAYLOAD.release.html_url }}` to `https://github.com/launchableinc/cli/releases/tag/{{ needs.tagpr.outputs.tag }}`.
WDYT??

ono-max: [0feb57a](https://github.com/launchableinc/cli/pull/605/commits/0feb57ad7f045d13a008fe449bb6799df68f0919)

Konboi (approved):

PR #607, opened 2023-08-09
Fix the notification message in Github Actions for Discord

ono-max:
`needs.tagpr.outputs.tag` cannot be referred from Ilshidur/action-discord (https://github.com/launchableinc/cli/actions/runs/5803613994/job/15731973019#step:9:15). Thus, we need to use enviroment variables instead.

> Environment variables can be interpolated in the message using brackets ({{ and }}) :
https://github.com/Ilshidur/action-discord#arguments


Konboi:
What do you think my change? https://github.com/launchableinc/cli/pull/608

PR #608, opened 2023-08-09
Pass to output values as string

Konboi:
To fix https://github.com/launchableinc/cli/actions/runs/5803613994/job/15731973019#step:9:15

We pass these output values as an object(or a function) now. (Sorry, I don't know how should I say)
But we need to pass these output values as strings. So I fixed it.

> When you use expressions in an if conditional, you may omit the ${{ }} expression syntax because GitHub Actions automatically evaluates the if conditional as an expression. Using the ${{ }} expression syntax turns the contents into a string, and strings are truth.

https://docs.github.com/en/actions/learn-github-actions/expressions#about-expressions

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -66,4 +66,4 @@ jobs:
       env:
         DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
       with:
-        args: 'Launchable CLI {{ needs.tagpr.outputs.tag }} is released! https://github.com/launchableinc/cli/releases/tag/{{ needs.tagpr.outputs.tag }}'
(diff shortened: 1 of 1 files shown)


ono-max:
Yes, that's correct in Github Actions, but I'm not sure if this expression can be accepted in Ilshidur/action-discord.
See https://github.com/Ilshidur/action-discord#arguments.

ono-max (approved):
In https://qiita.com/kyohah/items/4a2895230d0e525501a2#githubworkflowsdeployyml-%E3%82%92%E4%BD%9C%E6%88%90, github expression is used as follows. Let's give it a try in this way!

> with:
          args: "${{ github.event.comment.body }} command done"

PR #610, opened 2023-08-09
fix api values

Konboi:
The api pf HTTP error event value was fixed as "verification". So, I fixed to use an url path


ono-max (approved):
👍

ono-max (approved):
Thank you!

PR #613, opened 2023-08-10
Add validation when both option --no-build and --session are set

Konboi:
When the user set both the `--session` option and the `--no-build` option are set, the `--no-build` option, the `--no-build` option will be prioritized now. 
However, if the user set a session value, the CLI should prioritize it, so I fixed it.

### DIFF
--- a/launchable/commands/record/tests.py
+++ b/launchable/commands/record/tests.py
@@ -186,7 +186,7 @@ def tests(
     if is_no_build and session:
         click.echo(
             click.style(
-                "WARNING: Both options `--session` and `--no-build` are set.\nUsing --session option value ({}) and ignore `--no-build` option".format(session),  # noqa: E501
+                "WARNING: `--session` and `--no-build` are set.\nUsing --session option value ({}) and ignoring `--no-build` option".format(session),  # noqa: E501
                 fg='yellow'),
             err=True)
         is_no_build = False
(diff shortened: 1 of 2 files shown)


launchable/commands/record/tests.py
```diff
@@ -183,6 +183,14 @@ def tests(
         raise click.UsageError(
             'The cli already created `.launchable` file. If you want to use `--no-build` option, please remove `.launchable` file before executing.')  # noqa: E501
 
+    if is_no_build and session:
+        click.echo(
+            click.style(
+                "WARNING: Both options `--session` and `--no-build` are set.\nUsing --session option value ({}) and ignore `--no-build` option".format(session),  # noqa: E501
```
shibuiwilliam, line 189: I think '`--session` and `--no-build` options are set.' is natural.

launchable/commands/subset.py
```diff
@@ -198,6 +198,14 @@ def subset(
             )
             sys.exit(1)
 
+    if is_no_build and session:
+        click.echo(
+            click.style(
+                "WARNING: Both options `--session` and `--no-build` are set.\nUsing --session option value ({}) and ignore `--no-build` option".format(session),  # noqa: E501
```
shibuiwilliam, line 204: https://github.com/launchableinc/cli/pull/613/files#r1289407049

shibuiwilliam (approved):

PR #614, opened 2023-08-10
Fixed a truncation bug in the NUnit profile

kohsuke:
Turns out there's another odd behaviour with NUnitXML.Logger that we run into with a customer, which involves the use of TestCase annotation with parameter.

Improved the logic, and added a test case.

### DIFF
--- a/launchable/test_runners/nunit.py
+++ b/launchable/test_runners/nunit.py
@@ -44,7 +44,8 @@ def build_path(e: Element):
         # work around a bug in NUnitXML.Logger.
         # see nunit-reporter-bug-with-nested-type.xml test case
         methodname = e.attrs['methodname']
-        idx = methodname.rfind(".")
+        bra = methodname.find("(")
+        idx = methodname.rfind(".",0,bra)
         if idx >= 0:
             # when things are going well, method name cannot contain '.' since it's not a valid character in a symbol.
             # but when NUnitXML.Logger messes up, it ends up putting the class name and the method name, like
--- a/launchable/test_runners/nunit.py
+++ b/launchable/test_runners/nunit.py
@@ -45,7 +45,7 @@ def build_path(e: Element):
         # see nunit-reporter-bug-with-nested-type.xml test case
         methodname = e.attrs['methodname']
         bra = methodname.find("(")
-        idx = methodname.rfind(".",0,bra)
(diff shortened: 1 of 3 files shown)


launchable/test_runners/nunit.py
```diff
@@ -44,7 +44,8 @@ def build_path(e: Element):
         # work around a bug in NUnitXML.Logger.
         # see nunit-reporter-bug-with-nested-type.xml test case
         methodname = e.attrs['methodname']
-        idx = methodname.rfind(".")
+        bra = methodname.find("(")
+        idx = methodname.rfind(".",0,bra)
```
Konboi, line 48: FYI: https://github.com/launchableinc/cli/actions/runs/5815538212/job/15767255587?pr=614#step:9:21

tests/data/nunit/nunit-reporter-bug-with-nested-type.json
```diff
@@ -16,7 +16,25 @@
...
+    },
+    {
+      "type": "case",
+      "testPath": [
+        { "type": "Assembly",    "name": "Launchable.NUnit.Test.dll" },
+        { "type": "TestSuite",   "name": "Launchable" },
+        { "type": "TestSuite",   "name": "NUnit" },
+        { "type": "TestSuite",   "name": "Test" },
+        { "type": "TestSuite",   "name": "InlineParameterTests" },
+        { "type": "TestCase",    "name": "StrTest(\"dot.in.it\")" }
```
Konboi, line 28: note:  `methodname="StrTest(&quot;dot.in.it&quot;)"`

Konboi (approved):
LGTM

Please merge after fixing lint errors.

kohsuke:
Oops, I somehow always forget that the linter is in use here. It looks like I'm supposed to use pre-commit so let me fix that as well...

tests/data/nunit/nunit-reporter-bug-with-nested-type.json
kohsuke: Are you asking why this is the correct assertion?

Konboi: No, just for my confirmation. Sorry if I made you confused.

PR #618, opened 2023-08-28
Add validation for scenarios when both --observation and --output-exclusion-rules options are used

ono-max:
When using both --observation and --output-exclusion-rules options, no output is generated in the standard error. It would be beneficial to inform users about this behaviour.

## Example
```shell
$ /Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/bin/launchable subset --output-exclusion-rules --observation maven src/test/java
WARNING: Both options --observation and --output-exclusion-rules are set. No output will be generated in this case.
Your model is currently in training
Launchable created subset 52 for build sample (test session 46) in workspace launchableinc/mothership
(This test session is under observation mode)

|           |   Candidates |   Estimated duration (%) |   Estimated duration (min) |
|-----------|--------------|--------------------------|----------------------------|
| Subset    |            4 |                   100.00 |                       0.00 |
| Remainder |            0 |                     0.00 |                       0.00 |
|           |              |                          |                            |
| Total     |            4 |                   100.00 |                       0.00 |

Run `launchable inspect subset --subset-id 52` to view full subset details
```

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -200,8 +200,8 @@ def subset(
 
     if is_observation and is_output_exclusion_rules:
         msg = (
-            "WARNING: Both options --observation and --output-exclusion-rules are set. "
-            "No output will be generated in this case."
+            "WARNING: --observation and --output-exclusion-rules are set. "
+            "No output will be generated."
         )
         click.echo(
             click.style(
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -210,11 +210,11 @@ def subset(
             err=True,
         )
         org, workspace = get_org_workspace()
-        tracking_client.send_error_event(
+        tracking_client.send_event(
             event_name=Tracking.Event.WARNING,
-            stack_trace=msg,
             organization=org or "",
             workspace=workspace or "",
+            metadata={"warningMessage": msg}
         )
 
     if prioritize_tests_failed_within_hours is not None and prioritize_tests_failed_within_hours > 0:
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -187,6 +187,15 @@ def subset(
         )
         sys.exit(1)
 
+    if is_observation and is_output_exclusion_rules:
+        click.echo(
+            click.style(
+                "WARNING: Both options --observation and --output-exclusion-rules are set. "
+                "No output will be generated in this case.",
+                fg="yellow"),
+            err=True,
+        )
+
     if prioritize_tests_failed_within_hours is not None and prioritize_tests_failed_within_hours > 0:
         if ignore_new_tests or (ignore_flaky_tests_above is not None and ignore_flaky_tests_above > 0):
             click.echo(
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -10,6 +10,7 @@ from tabulate import tabulate
 
 from launchable.utils.authentication import get_org_workspace
 from launchable.utils.session import parse_session
+from launchable.utils.tracking import Tracking, TrackingClient
 
 from ..testpath import FilePathNormalizer, TestPath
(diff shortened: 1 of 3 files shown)


launchable/commands/subset.py
```diff
@@ -187,6 +187,15 @@ def subset(
         )
         sys.exit(1)
 
+    if is_observation and is_output_exclusion_rules:
```
ono-max, line 190: Users might receive confusing messages like the following when a session has already been created. However, I believe that this is a rare occurrence, so it should not be a significant issue.
```
- WARNING: Both options --observation and --output-exclusion-rules are set. No output will be generated in this case.
- WARNING: --observation flag was ignored. Observation mode can only be enabled for a test session during its initial creation. Add `--observation` option to the `launchable record session` command instead.
```

Konboi:
```
WARNING: `--observation` and `--outpu-exclusion-rules` are set. No output will be generated"
```

ref: https://github.com/launchableinc/cli/pull/613

Konboi: And for now, let's allow two warning messages will be displayed and see how it goes

launchable/commands/subset.py
```diff
@@ -187,6 +187,15 @@ def subset(
...
+        click.echo(
+            click.style(
+                "WARNING: Both options --observation and --output-exclusion-rules are set. "
+                "No output will be generated in this case.",
+                fg="yellow"),
+            err=True,
+        )
```
Konboi, line 197: Don't we need to send it as telemetry like this warning case?

launchable/commands/subset.py
ono-max: Sure, here is the log generated from running the CLI command that you'd like to review.

```
$ /Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/bin/launchable subset --target 70% --session builds/hogehoge/test_sessions/48 --no-build --rest foo.txt --output-exclusion-rules --observation maven src/test/java

WARNING: Both options --observation and --output-exclusion-rules are set. No output will be generated in this case.
WARNING: `--session` and `--no-build` are set.
Using --session option value (builds/hogehoge/test_sessions/48) and ignoring `--no-build` option
WARNING: --observation flag was ignored. Observation mode can only be enabled for a test session during its initial creation. Add `--observation` option to the `launchable record session` command instead.
```

launchable/commands/subset.py
ono-max: I added WARNING event 👍

launchable/commands/subset.py
Konboi: Sorry, this is my suggestion of warning message https://github.com/launchableinc/cli/pull/618#discussion_r1306999738

ono-max: [64751ae](https://github.com/launchableinc/cli/pull/618/commits/64751aee006630f7a13e8fcc2f6f292dd92511fe)

launchable/commands/subset.py
```diff
@@ -178,24 +179,63 @@ def subset(
...
+        msg = "Cannot use --observation and --get-tests-from-previous-sessions options at the same time"
         click.echo(
             click.style(
-                "Cannot use --observation and --get-tests-from-previous-sessions options at the same time",
+                msg,
                 fg="red"),
             err=True,
         )
+        org, workspace = get_org_workspace()
+        tracking_client.send_error_event(
```
Konboi, line 193: it’s ok in another PR but how about getting and setting org/workspace value in send_error_event?

Konboi (approved):

ono-max:
@Konboi 

Sorry, I've realized that there is a mistake in my previous change. I've noticed that WARNING event should be sent in the `send_event` method, as WARNING event is not an error event. Could you review this again?

launchable/commands/subset.py
ono-max: Sure, I'll address it in the next PR.

Konboi (approved):

PR #625, opened 2023-09-26
Add telemetry when the exit code is 2

ono-max:
I've forgotten to add telemetry when the exist code is 2. This PR addresses it.

### DIFF
--- a/tests/commands/record/test_session.py
+++ b/tests/commands/record/test_session.py
@@ -107,7 +107,7 @@ class SessionTest(CliTestCase):
         result = self.cli("record", "session", "--build", self.build_name, "--session-name", self.session_name)
         self.assertEqual(result.exit_code, 0)
 
-        payload = json.loads(responses.calls[2].request.body.decode())
+        payload = json.loads(responses.calls[3].request.body.decode())
         self.assert_json_orderless_equal(
             {"flavors": {},
              "isObservation": False, "links": [],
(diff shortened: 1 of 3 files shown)


tests/commands/record/test_session.py
```diff
@@ -107,7 +107,7 @@ def test_run_session_with_session_name(self):
         result = self.cli("record", "session", "--build", self.build_name, "--session-name", self.session_name)
         self.assertEqual(result.exit_code, 0)
 
-        payload = json.loads(responses.calls[2].request.body.decode())
+        payload = json.loads(responses.calls[3].request.body.decode())
```
takanabe, line 110: Why does this become from `2` to `3` even we haven't changed http request for `record session`? Do we count `tracking_client.send_error_event` here?

ono-max: > Do we count tracking_client.send_error_event here?

I think so, too. The http request has been added in this change. When I added `print(responses.calls[1].request.body.decode())`, I got the following output:

```
{"command": "RECORD_SESSION", "eventName": "USER_ERROR", "cliVersion": "1.73.1.dev10+g754709d", "metadata": {"stackTrace": "This session name (test_session_name) is already used. Please set another name.", "organization": "launchableinc", "workspace": "mothership", "api": ""}}
```

takanabe (approved):

PR #633, opened 2023-10-13
`record build` refactoring

kohsuke:
I was trying to add a new capability here but felt the logic was too complex at this point. A single big function that has tons of local variables, controlling flow of executions in a non-trivial way.

The main idea of this refactoring is:
- Use a single typed `Workspace` object with properties that we are trying to compute, whereas previously this was done by multiple maps keyed by repository names
- Use well-defined sub-functions to split the work into smaller bite-sized chunks, leaving the main logic (L346-356) higher altitude.

In retrospect, this was probably too big to take on without some prior consensus building. So it's OK if we decided not to do this.

If we do choose to move forward, please advise me how best to roll this out a bit slowly & carefully. There appears to be a good number of test cases to cover this, but this is a pretty critical functionality. If I break this, it could be a major outage.

### DIFF
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -197,7 +197,7 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
         if no_submodules:
             return workspaces
 
-        r = [] + workspaces
+        r = workspaces.copy()
         for w in workspaces:
             submodule_pattern = re.compile(r"^[\+\-U ](?P<hash>[a-f0-9]{40}) (?P<name>\S+)")
 
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -1,7 +1,7 @@
 import os
 import re
 import sys
-from typing import Dict, List, Tuple
+from typing import Dict, List, Optional
 
 import click
 from tabulate import tabulate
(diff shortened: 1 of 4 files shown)


kohsuke:
In retrospect, this was probably too big to take on without some prior consensus building. So it's OK if we decided not to do this.

There appears to be a good number of test cases to cover this, but this is a pretty critical functionality!

launchable/commands/record/build.py
```diff
@@ -169,100 +212,127 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
...
-        _commits = normalize_key_value_types(commits)
+    # Rely on --commit and --branch to create a list of workspaces, even when there's no local Git workspaces
+    def synthesize_workspaces() -> List[Workspace]:
+        # workspace by its name
+        ws = {}  # type: Dict[str,Workspace]
 
         commit_pattern = re.compile("[0-9A-Fa-f]{5,40}$")
-        for repo_name, hash in _commits:
+
+        branch_name_map = dict(normalize_key_value_types(branches))
```
kohsuke, line 250: I'm scratching my head with `normalize_key_value_types`. Shouldn't our key/value style option handling be consistent across the board? Our intended mechanism to do so is `KeyValueType` in `utils/click.py`, so I'd love to have better comment in the code to explain the motivation behind `normalize_key_value_types`, and why that should only apply to options in this command, but not similar options elsewhere.

launchable/commands/record/build.py
```diff
@@ -169,100 +212,127 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
...
-                invalid = True
-            submodules.append((repo_name, "", hash))
-        if invalid:
-            sys.exit(1)
+                sys.exit(1)
 
-    # Note: currently becomes unique command args and submodules by the hash.
-    # But they can be conflict between repositories.
-    uniq_submodules = list({commit_hash: (name, repo_dist, commit_hash)
-                            for name, repo_dist, commit_hash, in sources + submodules}.values())
```
kohsuke, line 193: I also didn't understand this code. I did the code spelunking and traced this all the way back to the original code in 649c1f82e6de9f2f7ce2c32917c5a02690b3439b but it still doesn't make sense.

What it does is that if some of the repositories we have happen to be checking out the same commit hash (so I'm imagining two git submodules pointing to the same commit), then we drop one and only keep one. But that means in one build we can see two sub-modules, the next build we might only see one sub-module, then the next build we might be back to two. That kind of discontinuity would create all sorts of confusion on the server side when we compute CUT.

So I chose not to preserve this behaviour, and instead send all the repositories, even if they point to the same commit. This must be a pretty rare case anyway, so it should be fine.

launchable/commands/record/build.py
```diff
@@ -169,100 +212,127 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
...
-            "buildNumber": build_name,
-            "commitHashes": commitHashes,
-        }
-
-        _links = capture_link(os.environ)
-        if len(links) != 0:
-            for link in normalize_key_value_types(links):
+                    err=True)
+                # TODO: is there any reason this is not an error? for now erring on caution
+                # sys.exit(1)
```
kohsuke, line 271: Calling an attention to here.

The original code was only treating this as a warning and not an error, and there's even a test case for that, so I assume it was important to behave that way, but I can't think of why.

This seems to be pretty squarely fall on the side of incorrect integration that should be an error, not a warning-&-recovery.

tests/cli_test_case.py
```diff
@@ -187,6 +187,9 @@ def cli(self, *args, **kwargs) -> click.testing.Result:
         """
         return CliRunner(mix_stderr=mix_stderr).invoke(main, args, catch_exceptions=False, **kwargs)
 
+    def assert_success(self, result: click.testing.Result):
+        self.assertEqual(result.exit_code, 0, result.stdout)
```
kohsuke, line 191: Makes it easier to diagnose a problem when the command unexpectedly fails. I'll come back and retrofit existing assertions with this one later.

launchable/commands/record/build.py
```diff
@@ -169,100 +212,127 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
...
+                        "If not, please set a directory use by --source option.".format(w.dir),
+                        fg='yellow'),
+                    err=True)
+                print(e, file=sys.stderr)
+                sys.exit(1)
+            w.calc_branch_name()
+
+        # Our historical behaviour is to ignore --branch options. In that case, we should reject if that option is present
+        # I also think it's OK to allow this option to explicitly let the branch name set, since the inference
+        #  of the branch name from the workspace is inherently a little fragile
```
kohsuke, line 241: Calling an attention to this one.

My understanding is that `--branch` option is originally developed as a companion add-on to `--commit`, where the idea is that we have to operate without any local git workspace.

Prior to refactoring, the `--branch` option appears to simply get ignored unless `--commit` option is also in use. But that is always a risky behaviour; Allowing user to set something, and simply ignoring it.

I'd like to suggest we allow the `--branch` option without `--commit` option. The idea is that, normally we infer the branch name automatically from the workspace, but that relies on some heuristics (i.e. picking up the first ref from `git show-ref`, so if multiple refs point to the same commit, only the luck would tell which one we get), so I can see the value of letting the user be precise.

launchable/commands/record/build.py
```diff
@@ -169,100 +212,127 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
...
 
-    if len(commits) != 0:
-        invalid = False
-        _commits = normalize_key_value_types(commits)
+    # Rely on --commit and --branch to create a list of workspaces, even when there's no local Git workspaces
+    def synthesize_workspaces() -> List[Workspace]:
+        # workspace by its name
+        ws = {}  # type: Dict[str,Workspace]
```
Konboi, line 246: https://github.com/launchableinc/cli/pull/636/commits/df475febdc0e3172cdb13baf97676482e371698c

Konboi (approved):
LGTM

__

I made some comments but it's nits

launchable/commands/record/build.py
```diff
@@ -104,63 +104,106 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
...
+                    self.branch = refs[0].split("/")[-1]
+            except Exception:
+                # cannot get branch name by git command
+                pass
+
+    # The first order of business is to ascertain what Git repositories we have in the workspace
+    def list_sources() -> List[Workspace]:
+        # This command accepts REPO_NAME=REPO_DIR as well as just REPO_DIR
+        pattern = re.compile(r'[^=]+=[^=]+')
+        ws = []  # type: List[Workspace]
```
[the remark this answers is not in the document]
kohsuke: Great, TIL. Let's incorporate that PR as well!

kohsuke:
Given the risk of the change, I'm feeling I shouldn't merge this while I'm on the road.

PR #641, opened 2023-10-27
Rewrite nunit plugin with ElementTree api

ono-max:
CLI uses the SAX API instead of the ElementTree API for parsing NUnit, while we commonly use the case ElementTree API for XML parsing. However, it's a little bit hard to extract stdout/stderr if we try to use SAX API. Thus, I've rewritten the parsing logic with ElementTree API.

I'll add the logic to extract stdout/stderr in other PR.

### DIFF
--- a/launchable/test_runners/nunit.py
+++ b/launchable/test_runners/nunit.py
@@ -6,8 +6,7 @@ from xml.etree import ElementTree as ET
 import click
 
 from launchable.commands.record.case_event import CaseEvent
-from launchable.testpath import TestPath
-from ..testpath import unparse_test_path, parse_test_path
+from launchable.testpath import TestPath, unparse_test_path, parse_test_path
 
 from . import launchable
 
--- a/launchable/test_runners/nunit.py
+++ b/launchable/test_runners/nunit.py
@@ -37,6 +37,8 @@ def build_path(e: ET.Element, parent: ET.Element):
     pp: TestPath = []
     parent_start_time = parent.attrib.get('start-time')
     if parent_start_time is not None and e.attrib.get('start-time') is None:
+        # the 'start-time' attribute is normally on <test-case> but apparently not always,
+        # so we try to use the nearest ancestor as an approximate
         e.set('start-time', parent_start_time)
     parent_path = parent.attrib.get('path')
     if parent_path is not None:
@@ -108,8 +110,6 @@ def nunit_parse_func(report: str):
                 status,
                 timestamp=str(e.attrib.get('start-time'))))  # timestamp is already iso-8601 formatted
 
-    # the 'start-time' attribute is normally on <test-case> but apparently not always,
-    # so we try to use the nearest ancestor as an approximate
     _parse_dfs_element(report=report, on_element=on_element)
     # return the obtained events as a generator
     return (x for x in events)
--- a/launchable/test_runners/nunit.py
+++ b/launchable/test_runners/nunit.py
@@ -1,12 +1,13 @@
 import glob
 import os
-from typing import Dict, List
+from typing import Callable, Dict, List
+from xml.etree import ElementTree as ET
 
 import click
 
 from launchable.commands.record.case_event import CaseEvent
 from launchable.testpath import TestPath
-from launchable.utils.sax import Element, SaxParser, TagMatcher
+from ..testpath import unparse_test_path, parse_test_path
 
 from . import launchable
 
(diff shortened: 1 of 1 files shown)


launchable/test_runners/nunit.py
```diff
@@ -162,9 +166,21 @@ def record_tests(client, report_xml):
 """
 
 
-def _replace_fixture_to_suite(paths) -> List[Dict[str, str]]:
+def _replace_fixture_to_suite(tp_str: str) -> List[Dict[str, str]]:
```
Konboi, line 169: What is tp? test_path?

launchable/test_runners/nunit.py
```diff
@@ -33,17 +34,21 @@ class Zot {
...
 
 
-def build_path(e: Element):
+def build_path(e: ET.Element, parent: ET.Element):
     pp: TestPath = []
-    if e.parent:
-        pp = e.parent.tags.get('path') or []  # type: ignore
-    if e.name == "test-suite":
+    parent_start_time = parent.attrib.get('start-time')
+    if parent_start_time is not None and e.attrib.get('start-time') is None:
```
Konboi, line 40: Question: What is this situation??

launchable/test_runners/nunit.py
```diff
@@ -119,13 +123,13 @@ def subset(client, report_xmls):
...
-        if e.name == "test-case":
-            client.test_path(_replace_fixture_to_suite(e.tags['path']))
+    def on_element(e: ET.Element, parent: ET.Element):
+        build_path(e, parent)
+        if e.tag == "test-case":
+            client.test_path(_replace_fixture_to_suite(e.attrib['path']))
 
     for report_xml in report_xmls:
-        SaxParser([], on_element).parse(report_xml)
+        _parse_dfs_element(report=report_xml, on_element=on_element)
```
Konboi, line 132: Q: What does `dfs` mean?

launchable/test_runners/nunit.py
```diff
@@ -119,13 +123,13 @@ def subset(client, report_xmls):
     Parse an XML file produced from NUnit --explore option to list up all the viable test cases
     """
 
-    def on_element(e: Element):
-        build_path(e)
-        if e.name == "test-case":
-            client.test_path(_replace_fixture_to_suite(e.tags['path']))
+    def on_element(e: ET.Element, parent: ET.Element):
```
Konboi, line 126: Why does this method need `parent? Doesn't `e` have a `parent` attribute?

launchable/test_runners/nunit.py
```diff
@@ -1,12 +1,13 @@
...
-from typing import Dict, List
+from typing import Callable, Dict, List
+from xml.etree import ElementTree as ET
 
 import click
 
 from launchable.commands.record.case_event import CaseEvent
 from launchable.testpath import TestPath
-from launchable.utils.sax import Element, SaxParser, TagMatcher
+from ..testpath import unparse_test_path, parse_test_path
```
Konboi, line 10: What is the difference between `launchable.testpath` and `..testpath`?
it seems the same to me...

launchable/test_runners/nunit.py
```diff
@@ -76,38 +81,37 @@ def split_filepath(path: str) -> List[str]:
...
             events.append(CaseEvent.create(
-                _replace_fixture_to_suite(e.tags['path']),  # type: ignore
-                float(e.attrs['duration']),
+                _replace_fixture_to_suite(e.attrib['path']),  # type: ignore
+                float(e.attrib['duration']),
                 status,
-                timestamp=str(e.tags['startTime'])))  # timestamp is already iso-8601 formatted
+                timestamp=str(e.attrib.get('start-time'))))  # timestamp is already iso-8601 formatted
 
     # the 'start-time' attribute is normally on <test-case> but apparently not always,
```
Konboi, line 112: Do we need this comment here?

Konboi (commented):
We didn't change the test results. It means we might be able to keep the back compatibility. 👍 

But I made some comments

launchable/test_runners/nunit.py
ono-max: Depth first search

ono-max: https://en.wikipedia.org/wiki/Depth-first_search

launchable/test_runners/nunit.py
ono-max: [be00068](https://github.com/launchableinc/cli/pull/641/commits/be000687bc284ed8f7440dc18855c9b6e0d39163)

launchable/test_runners/nunit.py
ono-max: In elementTree API, each element does not have parent attribute.

https://docs.python.org/ja/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element

launchable/test_runners/nunit.py
ono-max: [150d812](https://github.com/launchableinc/cli/pull/641/commits/150d8122d2f9e1797453a2c8b266566b42245747)

launchable/test_runners/nunit.py
ono-max: [ec20f06](https://github.com/launchableinc/cli/pull/641/commits/ec20f06440ca75dfb58bfc441e15ea4a130eaa00)

Konboi (approved):
LGTM

PR #645, opened 2023-10-31
Do not override test path when the keyword is same in cucumber

ono-max:


launchable/test_runners/cucumber.py
```diff
@@ -253,8 +253,8 @@ def parse_func(self, report_file: str) -> Generator[CaseEventType, None, None]:
                     {"type": "testcase", "name": test_case},
                 ]
 
-                for k in test_case_info.steps:
-                    test_path.append({"type": k, "name": test_case_info.steps[k]})
+                for step in test_case_info.steps:
+                    test_path.append({"type": step[0], "name": step[1]})
```
Konboi, line 257: Don't we have to check the step size?

ono-max: For validation?

Konboi: Just a question. I think that there isn't any case that step size is under 2.
But if it happens, this code will throw an error.

ono-max: Ok, let me check the size just in case. Thanks.

ono-max: [0dc0d23](https://github.com/launchableinc/cli/pull/645/commits/0dc0d238251c24bd189d1974068e6a2c4cc6ee28)

Konboi (approved):

PR #683, opened 2023-12-25
Enable tokenless auth with Github App in CI

ono-max:


.github/workflows/python-package.yml
```diff
@@ -16,8 +16,8 @@ on:
 env:
   LAUNCHABLE_ORGANIZATION: "launchableinc"
   LAUNCHABLE_WORKSPACE: "cli"
-  EXPERIMENTAL_GITHUB_OIDC_TOKEN_AUTH: 1
```
ono-max, line 19: We need to remove this env because of https://github.com/launchableinc/cli/blob/main/launchable/utils/authentication.py#L40.

Konboi (approved):

PR #687, opened 2023-12-25
Don't have to execute python test when exe.jar related files changed

Konboi:
SSIA

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -7,8 +7,14 @@ on:
   workflow_dispatch:
   push:
     branches: [ main ]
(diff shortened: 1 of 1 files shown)


.github/workflows/python-package.yml
```diff
@@ -7,8 +7,14 @@ on:
   workflow_dispatch:
   push:
     branches: [ main ]
+    paths-ignore:
```
Konboi, line 10: about: paths-ignore https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

ono-max (approved):

PR #707, opened 2024-01-03
Misc update

kohsuke:
* Removed duplicate code that appear to be a result of a refactoring error or merge failure
* Fixed a broken link in error message

### DIFF
--- a/launchable/commands/helper.py
+++ b/launchable/commands/helper.py
@@ -114,32 +114,18 @@ def find_or_create_session(
         _check_observation_mode_status(session_id, is_observation, tracking_client=tracking_client)
         return session_id
 
-    if build_name and saved_build_name != build_name:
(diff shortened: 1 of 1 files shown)


launchable/commands/helper.py
```diff
@@ -106,40 +108,27 @@ def find_or_create_session(
...
-            build_name=saved_build_name,
-            save_session_file=True,
-            print_session=False,
-            flavor=flavor,
-            is_observation=is_observation,
-            links=links,
-            is_no_build=is_no_build,
-            lineage=lineage,
-        )
-        return read_session(saved_build_name)
```
kohsuke, line 142: This code block exactly identical as the code right above, which makes me think this is a copy/paste error or merge error.

Konboi (approved):
Thanks

PR #716, opened 2024-01-10
Refactoring to deal with global configs + SSL cert check bypass

kohsuke:
For unblocking a customer, added a switch to disable SSL cert check.

To deal with growing number of global options, promoted the use of the global 'Application' object (formerly called 'AppBase', but to me, that name implied there's something that extends from this base, which is not the case).

### DIFF
--- a/launchable/__main__.py
+++ b/launchable/__main__.py
@@ -6,6 +6,8 @@ from os.path import basename, dirname, join
 
 import click
 
+from launchable.app import Application
+
 from .commands.inspect import inspect
 from .commands.record import record
 from .commands.split_subset import split_subset
@@ -14,11 +16,6 @@ from .commands.subset import subset
 from .commands.verify import verify
 from .utils import logger
 from .version import __version__
-
-
-class AppBase(object):
-    def __init__(self, dry_run: bool):
-        self.dry_run = dry_run
 
 
 @click.group()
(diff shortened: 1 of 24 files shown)


launchable/__main__.py
```diff
@@ -16,11 +18,6 @@
 from .version import __version__
 
 
-class AppBase(object):
```
kohsuke, line 19: Moved over to `app.py`; I felt a little uneasy importing from `launchable.__main__` for some reasons.

launchable/commands/test_path_writer.py
```diff
@@ -3,17 +3,18 @@
...
     base_path: Optional[str] = None
 
-    def __init__(self, dry_run=False):
-        self._formatter = TestPathWriter.default_formatter
+    def __init__(self, app: Application):
+        self._formatter = TestPathWriter.default_formatter  # type: Callable[[TestPath], str]
         self._separator = "\n"
-        self._same_bin_formatter = None
-        self.dry_run = dry_run
+        self._same_bin_formatter = None  # type: Optional[Callable[[str], Dict[str, str]]]
```
kohsuke, line 16: When I made my change, type checker started complaining, which was strange. More context at https://github.com/launchableinc/cli/pull/716/commits/9435ec06207af3f7c4dcddafbdfdce237d6cc9cf if you care.

src/test/java/com/launchableinc/ingest/commits/BUILD
```diff
@@ -1,21 +1,11 @@
...
-java_test(
-    name = "CommitIngesterTest",
-    srcs = ["CommitIngesterTest.java"],
-    deps = [
-        "//src/main/java/com/launchableinc/ingest/commits",
-        "@maven//:junit_junit",
-        "@maven//:org_eclipse_jgit_org_eclipse_jgit",
         "@maven//:org_mock_server_mockserver_junit_rule_no_dependencies",
     ],
 )
```
kohsuke, line 11: Consolidated test targets instead of adding one more

launchable/commands/test_path_writer.py
```diff
@@ -3,17 +3,18 @@
...
 from ..testpath import TestPath
 
 
 class TestPathWriter(object):
     base_path: Optional[str] = None
 
-    def __init__(self, dry_run=False):
-        self._formatter = TestPathWriter.default_formatter
+    def __init__(self, app: Application):
+        self._formatter = TestPathWriter.default_formatter  # type: Callable[[TestPath], str]
```
Konboi, line 14: nits
```suggestion
        self._formatter: Callable[[TestPath], str] = TestPathWriter.default_formatter
```

launchable/commands/test_path_writer.py
Konboi, line 16: ditto
```suggestion
        self._same_bin_formatter: Optional[Callable[[str], Dict[str, str]]] = None
```

Konboi (approved):

PR #720, opened 2024-01-11
Refactor parse test result logics

Konboi:
Current parse logic is a little difficult to read and understand.
So, renamed and defined some methods for easy to read and understand

I didn't edit any test codes, so the result might not changed.

### DIFF
--- a/launchable/test_runners/cucumber.py
+++ b/launchable/test_runners/cucumber.py
@@ -8,7 +8,6 @@ from typing import Dict, Generator, List, Optional
 from xml.etree import ElementTree as ET
 
 import click
-from typing_extensions import Self
 
 from launchable.testpath import FilePathNormalizer
 
@@ -381,9 +380,10 @@ class TestCaseInfo(Result):
     def to_hook(self) -> TestCaseHookInfo:
         return TestCaseHookInfo(duration_nano_sec=self.duration_nano(), statuses=self.statuses(), stderr=self.stderr())
 
-    # Need to merge Background steps to main test scenario to calculate correct test duration,
-    # then, we don't need step information of Background. So append it as hooks
-    def append_background_results(self, other: Self) -> None:
+    # Type of other TestCaseInfo (Self)..  Python3.6 cannot support `Self`` type even if used typing_extensions module
+    def append_background_results(self, other) -> None:
+        # Need to merge Background steps to main test scenario to calculate correct test duration,
+        # then, we don't need step information of Background. So append it as hooks
         self.append_hook_info(other.to_hook())
 
 
--- a/launchable/test_runners/cucumber.py
+++ b/launchable/test_runners/cucumber.py
@@ -219,28 +219,27 @@ class JSONReportParser:
         for d in data:
             file_name = d.get("uri", "")
             class_name = d.get("name", "")
-            # cucumber can define multiple Background steps
-            unified_backgrounds: Optional[TestCaseInfo] = None
+
+            # Cucumber can define repeating the same `Given` steps as a `Background`
+            # https://cucumber.io/docs/gherkin/reference/#background
+            background: Optional[TestCaseInfo] = None
 
             for element in d.get("elements", []):
                 test_case = element.get("name", "")
                 # Scenario hooks run for every scenario.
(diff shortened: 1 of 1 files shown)


Konboi:
CI passed https://github.com/launchableinc/cli/actions/runs/7485531973

Konboi:
@ono-max Not rush, What do you think my change??

launchable/test_runners/cucumber.py
```diff
@@ -256,16 +257,16 @@ def parse_func(self, report_file: str) -> Generator[CaseEventType, None, None]:
                     {"type": "testcase", "name": test_case},
                 ]
 
-                for step in test_case_info.steps:
+                for step in test_case_info.steps():
```
Konboi, line 260: I felt we can defined this logic as a method

launchable/test_runners/cucumber.py
```diff
@@ -216,36 +216,37 @@ def parse_func(self, report_file: str) -> Generator[CaseEventType, None, None]:
             click.echo("Can't find test reports from {}. Make sure to confirm report file.".format(
                 report_file), err=True)
 
-        background_test_case_info = None
         for d in data:
             file_name = d.get("uri", "")
             class_name = d.get("name", "")
+            # cucumber can define multiple Background steps
```
ono-max, line 222: Are you sure? I don't think we can have multiple Background.

> You can only have one set of Background steps per Feature or Rule. If you need different Background steps for different scenarios, consider breaking up your set of scenarios into more Rules or more Features.

https://cucumber.io/docs/gherkin/reference/#background

Konboi: Ah, sorry, I have to write “multiple steps of a Background “

ono-max: Sorry, I still don't understand what you said. In which case do we need to unify Background durations?

ono-max: `unified_backgrounds` variable is assigned `None` at https://github.com/launchableinc/cli/pull/720/files#diff-959e6ab7da948bb1bcd21d094e6921c6b5d9027f75db90e13901c16753f689d5R243.

ono-max: Thus, I believe that there is no case that we need to unify Background durations. Am I missing something?

launchable/test_runners/cucumber.py
```diff
@@ -304,51 +305,95 @@ def _create_file_candidate_list(file: str) -> List[str]:
...
 
 
-def _extract_test_case_info_from_hook(data):
-    duration = 0  # nano sec
+class TestCaseHookInfo(Result):
+    def __init__(self, duration_nano_sec: int, statuses: List[str], stderr: List[str]) -> None:
+        super().__init__(statuses=statuses, duration_nano_sec=duration_nano_sec, error_message=stderr)
+
+
+def _parse_hook_from_element(element: Dict[str, List]) -> TestCaseHookInfo:
```
ono-max, line 320: This might be too magical, so it's up to you. However, we can write like this.
```py
def _parse_hook_from_element(element: Dict[str, List]) -> TestCaseHookInfo:
    duration_nano_sec = 0
    statuses = []
    stderr = []
    def _parse_step_result(step: Dict[str, Dict]):
        result = step.get("result", None)
        if result:
            nonlocal duration_nano_sec
            duration_nano_sec += result.get("duration", 0)
            statuses.append(result.get("status"))
            if result.get("error_message", None):
                stderr.append(result["error_message"])
    for step in element.get("before", []):
        _parse_step_result(step)
    for step in element.get("after", []):
        _parse_step_result(step)

    return TestCaseHookInfo(
        duration_nano_sec=duration_nano_sec,
        statuses=statuses,
        stderr=stderr
    )
```

launchable/test_runners/cucumber.py
Konboi: I understood what you wanted to say. I'll update it, please wait my fixing

launchable/test_runners/cucumber.py
Konboi: Thanks I apply it

ono-max (approved):
Excellent refactoring! Thanks!

PR #737, opened 2024-01-18
dependency updates

kohsuke:
Built on #735 to ensure Java8 compatibility


src/test/java/com/launchableinc/ingest/commits/java8-compat.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -ex
+# Make sure the java bits we ship are Java8 compatible
+
+# report the full dump to assist investigation
+# the first number printed is the Java class file version. We want to support Java8 = 52.0, so no newer version allowed
```
Konboi, line 5: > # the first number printed is the Java class file version. We want to support Java8 = 52.0, so no newer version allowed

TIL

Konboi (approved):
Thanks

PR #744, opened 2024-01-29
Capture the display name of the URL.

kohsuke:
When you have a long list of "View in Jenkins" it's not useful, but the concrete names that they are linking to would allow users to select the right one much more easily.


Konboi (approved):
Nice

PR #748, opened 2024-01-31
Fetch the source jar + workflow fix

kohsuke:
- The workflow definition had a few issues that need tweaking
- Source jars are crucial for editing with the IDE, and this also conveniently tests the previous change


kohsuke:
And the workflow verified to work as expected!

ono-max (approved):

PR #752, opened 2024-02-02
Explicitly bring in a provider

kohsuke:
Yabuki-san reported that as of the current tip of the main branch (HEAD^), slf4j is issuing a warning message saying there's no provider.

I can only assume some of the recent dependency changes caused this change. Previously some provider must have been brought in via an indirect dependency that's gone now.

So add the provider directly.


Konboi (approved):
Confirmed. LGTM

PR #756, opened 2024-02-05
Make sure tests are triggered with updated dependencies

kohsuke:
The key motivation of this workflow is to make sure tests are run with the updated dependencies, but I noticed that wasn't happening.

I tracked it down to the URL referenced in the comment. Also see https://github.com/EndBug/add-and-commit?tab=readme-ov-file#the-commit-from-the-action-is-not-triggering-ci The issue described there is not specific to that workflow action, it equally applies to our inline definition of git-push.

This should do the trick.


kohsuke:
Reopening to see if that gets tests running again.

jshinn704 (approved):

PR #761, opened 2024-02-07
Handle build/{buildId} correctly

ono-max:
When [adding the mock response which throws timeout error](https://github.com/launchableinc/cli/pull/761/files#diff-d1c38c43b1cef30c3db605e17c4ade6a177835334931b8d03aa0f9783bec0c34R303) in `builds/{build}`, I got the following error. This might stop client's pipeline. Thus, this PR addresses the problem.

```
ERROR: test_record_tests (tests.commands.test_api_error.APIErrorTest.test_record_tests)
  File "/Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/lib/python3.11/site-packages/responses/__init__.py", line 229, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/mock.py", line 1828, in _inner
    return f(*args, **kw)
           ^^^^^^^^^^^^^^
  File "/Users/ono-max/workspace/cli/tests/commands/test_api_error.py", line 326, in test_record_tests
    result = self.cli("record", "tests", "--session", self.session, "minitest", str(self.test_files_dir) + "/")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ono-max/workspace/cli/tests/cli_test_case.py", line 189, in cli
    return CliRunner(mix_stderr=mix_stderr).invoke(main, args, catch_exceptions=False, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/lib/python3.11/site-packages/click/testing.py", line 408, in invoke
    return_value = cli.main(args=args or (), prog_name=prog_name, **extra)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ono-max/.local/share/virtualenvs/cli-o7XEfK99/lib/python3.11/site-packages/click/decorators.py", line 45, in new_func
    return f(get_current_context().obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ono-max/workspace/cli/launchable/test_runners/launchable.py", line 117, in record_tests
    client.scan(t, file_mask)
    ^^^^^^^^^^^
AttributeError: 'Application' object has no attribute 'scan'

````

### DIFF
--- a/launchable/commands/record/tests.py
+++ b/launchable/commands/record/tests.py
@@ -217,7 +217,7 @@ def tests(
             res.raise_for_status()
 
             session_id = "builds/{}/test_sessions/{}".format(build_name, res.json().get("id"))
-            record_start_at = get_record_start_at(session_id, client)
+            record_start_at = get_record_start_at(session=session_id, client=client, tracking_client=tracking_client)
         else:
             # The session_id must be back, so cast to str
             session_id = str(find_or_create_session(
@@ -229,7 +229,7 @@ def tests(
                 lineage=lineage,
                 tracking_client=tracking_client))
             build_name = read_build()
-            record_start_at = get_record_start_at(session_id, client)
+            record_start_at = get_record_start_at(session=session_id, client=client, tracking_client=tracking_client)
 
         build_name, test_session_id = parse_session(session_id)
     except Exception as e:
@@ -588,7 +588,7 @@ def tests(
 INVALID_TIMESTAMP = datetime.datetime.fromtimestamp(0)
 
 
-def get_record_start_at(session: Optional[str], client: LaunchableClient):
+def get_record_start_at(session: Optional[str], client: LaunchableClient, tracking_client: TrackingClient):
     """
     Determine the baseline timestamp to be used for up-to-date checks of report files.
     Only files newer than this timestamp will be collected.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: launchable/commands/record/tests.py [hunk 1]
```diff
@@ -604,26 +604,32 @@ def get_record_start_at(session: Optional[str], client: LaunchableClient):
...
+        created_at = res.json()["createdAt"]
+        Logger().debug("Build {} timestamp = {}".format(build_name, created_at))
+        t = parse_launchable_timeformat(created_at)
+        return t
+    except Exception as e:
+        tracking_client.send_error_event(
+            event_name=Tracking.ErrorEvent.INTERNAL_CLI_ERROR,
+            stack_trace=str(e),
+        )
         return INVALID_TIMESTAMP
```
kohsuke, line 631: What I worry is that, if we introduce any mistakes within the try block, say a simple coding mistake, we'd never notice because this silently recovers with `INVALID_TIMESTAMP`.

I get the value that, during the production customer use, the up to date check is not valuable enough to stop & hold the delivery pipeline, so we should recover from that error.

Can we meet both of those goals instead of sacrificing one over the other?

ono-max:
> Don't we need to solve the underlying problem?

I believe that we can solve the underlying problem if we catch the exception in the callee layer like https://github.com/launchableinc/cli/pull/763. In the current design, the caller side like https://github.com/launchableinc/cli/blob/main/launchable/commands/inspect/tests.py#L23 has a responsibility for handling the exception. If we change the design that the callee layer has a responsibility for handling the exception, there are some pros and cons.

## pros

* We don't need to worry about these case.

## cons

* The caller layer does not know if the exception is occurred. Sometimes, there might be the case that the caller side wanna know if it happens.

ono-max:
@kohsuke 

Does that make sense? Or did I misunderstand your opinion?

kohsuke (approved):
Just to make sure we are talking about the same thing, the part that I can't get over is that I have real difficulty imagining any situations in which we accept `AttributeError: 'Application' object has no attribute 'scan'` on a legitimate code path. That's what I'm referring to as a programming error. From that perspective, and if you accept that premise, then hopefully you agree with the logic that such problem shouldn't be swallowed.

Anyway, I think I'm dragging this out for too long, and it's not important enough to block the orderly flow of engineering execution, so I'll trust your judgement, if you consider my feedback and still feels the current proposed change is the right way to go.

I note a seemingly unrelated test failure from CI but I trust you'll look at it, too.

PR #764, opened 2024-02-14
Export test path in raw profile as a structure

ono-max:
We're going to receive the following JSON in raw profile.

```
{
  "testCases": [
    {
      "testPathComponents": [
        {
          "type": "file",
          "name": "login/test_a.py"
        },
        {
          "type": "class",
          "name": "class1"
        },
        {
          "type": "testcase",
          "name": "testcase#899"
        }
      ],
      "duration": 42,
      "status": "TEST_PASSED",
      "stdout": "This is stdout",
      "stderr": "This is stderr",
      "createdAt": "2021-10-05T12:34:00"
    }
  ]
}
```

### DIFF
--- a/launchable/test_runners/raw.py
+++ b/launchable/test_runners/raw.py
@@ -161,12 +161,13 @@ def record_tests(client, test_result_files):
         default_created_at = datetime.datetime.now(datetime.timezone.utc).isoformat()
         for case in doc['testCases']:
             test_path_components: TestPath = case.get('testPathComponents', None)
-            if test_path_components is None:
-                test_path: str = case.get('testPath', None)
-                if test_path:
-                    test_path_components = parse_test_path(test_path)
-                else:
-                    raise ValueError("Missing testPath or testPathComponents field in the test case.")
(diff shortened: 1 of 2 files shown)


launchable/test_runners/raw.py
```diff
@@ -133,21 +160,27 @@ def parse_json(test_result_file: str) -> Generator[CaseEventType, None, None]:
             doc = json.load(f)
         default_created_at = datetime.datetime.now(datetime.timezone.utc).isoformat()
         for case in doc['testCases']:
-            test_path = parse_test_path(case['testPath'])
+            test_path_components: TestPath = case.get('testPathComponents', None)
```
Konboi, line 163: Don't we need to check `oneOf`? 
It seems `testPathComponents` will take over `testPath` when both are set.

ono-max: [65fa86c](https://github.com/launchableinc/cli/pull/764/commits/65fa86c19329c375cbd5c7649b601e334ba1dea3)

Konboi (approved):

PR #765, opened 2024-02-16
Get branch name correctly

ono-max:
In the push event, GITHUB_HEAD_REF and GITHUB_BASE_REF are empty. In such cases, the git show-ref command is utilized. A complication arises when configuring a branch that contains a slash in the push event. For instance, if the branch name is test/test-branch, the expected value is "test/test-branch," but the current CLI displays "test-branch" instead. This PR resolves this discrepancy.

```
$ python                                   
Python 3.11.4 (v3.11.4:d2340ef257, Jun  6 2023, 19:15:51) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> m = re.search('[a-f0-9]{40} refs/heads/(.*)', '65fa86c19329c375cbd5c7649b601e334ba1dea3 refs/heads/foobar/foobar2')
>>> m.group(1)
'foobar/foobar2'
>>> m = re.search('[a-f0-9]{40} refs/heads/(.*)', '65fa86c19329c375cbd5c7649b601e334ba1dea3 refs/heads/foobar')
>>> m.group(1)
'foobar'
>>> 
```

### DIFF
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -135,6 +135,12 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
 
             # Github Actions
             # ref: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
+            # These environment variables cannot be retrieved when a `push` event is emitted.
+            # Here is a note regarding the output of `git show-ref`:
+            # - Git tag is pushed during a `push` event
+            #   => ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/tags/tag-name
+            # - Git commit is pushed during a `push` event
+            #   => ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/heads/branch/branch-name
             if os.environ.get(GITHUB_ACTIONS_KEY):
                 self.branch = os.environ.get(GITHUB_ACTIONS_GITHUB_HEAD_REF_KEY) or \
                     os.environ.get(GITHUB_ACTIONS_GITHUB_BASE_REF_KEY)
@@ -159,10 +165,12 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
                 refs = [ref for ref in show_ref.split("\n") if self.commit_hash in ref]
 
                 if len(refs) > 0:
-                    # e.g) ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/heads/branch-name
-                    match = re.search('[a-f0-9]{40} refs/heads/(.*)', refs[0])
+                    # We assume the following values:
+                    # * ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/heads/branch/branch-name
+                    # * ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/remotes/origin/branch-name
+                    match = re.search('[a-f0-9]{40} refs/(heads|remotes/origin)/(.*)', refs[0])
                     if match:
-                        self.branch = match.group(1)
+                        self.branch = match.group(2)
                     else:
                         self.branch = refs[0].split("/")[-1]
             except Exception:
--- a/launchable/commands/record/build.py
+++ b/launchable/commands/record/build.py
@@ -160,7 +160,11 @@ def build(ctx: click.core.Context, build_name: str, source: List[str], max_days:
 
                 if len(refs) > 0:
                     # e.g) ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/heads/branch-name
-                    self.branch = refs[0].split("/")[-1]
+                    match = re.search('[a-f0-9]{40} refs/heads/(.*)', refs[0])
+                    if match:
+                        self.branch = match.group(1)
+                    else:
+                        self.branch = refs[0].split("/")[-1]
             except Exception:
                 # cannot get branch name by git command
                 pass


launchable/commands/record/build.py
```diff
@@ -160,7 +160,9 @@ def calc_branch_name(self):
 
                 if len(refs) > 0:
                     # e.g) ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/heads/branch-name
-                    self.branch = refs[0].split("/")[-1]
+                    match = re.search('[a-f0-9]{40} refs/heads/(.*)', refs[0])
```
ono-max, line 163: Do we have a case that commit SHA is not something like `ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/heads/branch-name`?

ono-max: Just in case, I've added like this https://github.com/launchableinc/cli/pull/765/files#diff-dbb014559b2adb369474bcb0dde7d795e50930c4b604a3d94a6a53961159d845R167.

Konboi (commented):
1. Did you check other cases and judge support only `refs/heads` case e.g.) `refs/remote/origin/branch/name`, `refs/tags/alpha/v1.1`, or etc
2. How about changing the branch name that includes `/` to check this change?

ono-max:
> How about changing the branch name that includes / to check this change?

I've confirmed my change works in https://app.launchableinc.com/organizations/launchableinc/workspaces/cli/data/builds/2515737

ono-max:
> Did you check other cases and judge support only refs/heads case e.g.) refs/remote/origin/branch/name, refs/tags/alpha/v1.1, or etc

No, I didn't check them.

Konboi:
>> 2. How about changing the branch name that includes / to check this change?
> I've confirmed my change works in https://app.launchableinc.com/organizations/launchableinc/workspaces/cli/data/builds/2515737

Sorry, my suggestion isn't good. 
Because this repository CI works on GitHub Actions, so the cli got a branch name from an environment value. see: [code](https://github.com/launchableinc/cli/blob/main/launchable/commands/record/build.py#L139-L140). It means this repo won't be affected this change.

I think we should check this change on the other project before merge it, what do you think?

ono-max:
Oh, I checked the branch name in https://github.com/launchableinc/cli/pull/766.

ono-max:
@Konboi 

I've checked my implementation works in https://github.com/launchableinc/cli/pull/771. CLI gets a branch name correctly at https://app.launchableinc.com/organizations/launchableinc/workspaces/cli/data/test-sessions/2630878.

Konboi:
> I've checked my implementation works in https://github.com/launchableinc/cli/pull/771. CLI gets a branch name correctly at https://app.launchableinc.com/organizations/launchableinc/workspaces/cli/data/test-sessions/2630878.

NOTE: We can get `GITHUB_HEAD_REF` value only pull_request events. 

> This is only set when the event that triggers a workflow run is either pull_request or pull_request_target

ref: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables


This actions worked by `push` trigger. But Launchable could collect branch name correctly 
image

image

Konboi:
>> Did you check other cases and judge support only refs/heads case e.g.) refs/remote/origin/branch/name, refs/tags/alpha/v1.1, or etc
> No, I didn't check them.

@ono-max  Don't we have to check this case, right?

ono-max:
Thank you for pointing it out. I believe that we do not have to check refs/remote/origin/branch/name because the target branch is checked out in https://github.com/actions/checkout. However, I'm not sure about refs/tags/alpha/v1.1. Let me confirm it.

ono-max:
@Konboi 

I've noticed that refs/remote/origin/branch/name is used when pushing a tag. Thus, I've fixed it. Can you review it?

# Before

https://github.com/launchableinc/cli/tree/test/test-tag

https://app.launchableinc.com/organizations/launchableinc/workspaces/cli/data/test-sessions/2631169

# After

https://github.com/launchableinc/cli/tree/test/test-tag-after-fix

https://app.launchableinc.com/organizations/launchableinc/workspaces/cli/data/test-sessions/2631278

launchable/commands/record/build.py
```diff
@@ -135,6 +135,12 @@ def calc_branch_name(self):
 
             # Github Actions
             # ref: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
+            # These environment variables cannot be retrieved when a `push` event is emitted.
```
Konboi, line 138: Nice comment 👍

launchable/commands/record/build.py
```diff
@@ -159,8 +165,14 @@ def calc_branch_name(self):
...
 
                 if len(refs) > 0:
-                    # e.g) ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/heads/branch-name
-                    self.branch = refs[0].split("/")[-1]
+                    # We assume the following values:
+                    # * ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/heads/branch/branch-name
+                    # * ed6de84bde58d51deebe90e01ddfa5fa78899b1c refs/remotes/origin/branch-name
+                    match = re.search('[a-f0-9]{40} refs/(heads|remotes/origin)/(.*)', refs[0])
+                    if match:
+                        self.branch = match.group(2)
```
Konboi, line 173: Note: https://docs.python.org/3/library/re.html#re.Match.group
if refs[0] was `refs/heads/branch/branch-name`

match.group(0) is `refs/heads/branch/branch-name`
match.group(1) is `heads`
match.group(2) is `branch/branch-name`

Konboi (approved):
Great work! Thank you!!

PR #790, opened 2024-02-28
Do not stop pipeline in record test command

ono-max:
Alternative solution for https://github.com/launchableinc/cli/pull/761.

### DIFF
--- a/launchable/test_runners/launchable.py
+++ b/launchable/test_runners/launchable.py
@@ -4,6 +4,7 @@ import sys
 import types
 
 import click
+from launchable.app import Application
 
 from launchable.commands.record.tests import tests as record_tests_cmd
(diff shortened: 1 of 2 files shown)


ono-max:
@kohsuke 

In the previous meeting you suggest that moving https://github.com/launchableinc/cli/blob/main/launchable/commands/record/tests.py#L202-L244 into https://github.com/launchableinc/cli/blob/main/launchable/commands/record/tests.py#L406. However, it does not work because https://github.com/launchableinc/cli/blob/main/launchable/test_runners/launchable.py#L117 is called before calling `run()` method.

Thus, I came up with other solution to solve the problem.

kohsuke (approved):

kohsuke:
Thanks! I suggest leaving some comments abouot what that check represents.

PR #811, opened 2024-03-15
[UX-171] error message reporting improvement

kohsuke:
If the server reports an error, print the payload to assist the troubleshooting. When our service responds, there's JSON payload that contains a quick message. Other HTTP intermediaries may insert other failures, including intermediaries on the customer side.

@Konboi Let me know when is the best time to merge something like this. I'd imagine you might want to push this out past the scheduled maintenance.

### DIFF
--- a/launchable/commands/inspect/subset.py
+++ b/launchable/commands/inspect/subset.py
@@ -1,14 +1,12 @@
 import json
-import os
 import sys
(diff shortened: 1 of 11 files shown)


Konboi:
Sorry for late, I'm going to review this PR

Konboi (approved):
LGTM & Great Work!!

PR #817, opened 2024-03-22
Fix add metadata

Konboi:
When comparing a NoneType object and another type, we encountered an error e.g.) `TypeError: '<' not supported between instances of 'NoneType' and 'list'`  and we couldn't use the `assert_json_orderless_equal(self, a, b)` method

So, I fixed it

https://github.com/launchableinc/cli/actions/runs/8390491593?pr=817

### DIFF
--- a/tests/cli_test_case.py
+++ b/tests/cli_test_case.py
@@ -223,7 +223,10 @@ class CliTestCase(unittest.TestCase):
         """
         def tree_sorted(obj):
             if isinstance(obj, dict):
-                return sorted((k, tree_sorted(v)) for k, v in obj.items())
+                # Convert the dictionary items into a list of tuples,
+                # where each tuple contains the key and the recursively sorted value.
+                # this sort option `key=lambda item: (item[1] is None, item)`` is for when the value is None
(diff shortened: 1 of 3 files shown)


tests/cli_test_case.py
```diff
@@ -223,7 +223,10 @@ def assert_json_orderless_equal(self, a, b):
         """
         def tree_sorted(obj):
             if isinstance(obj, dict):
-                return sorted((k, tree_sorted(v)) for k, v in obj.items())
+                # Convert the dictionary items into a list of tuples,
+                # where each tuple contains the key and the recursively sorted value.
+                # this sort option `key=lambda item: (item[1] is None, item)`` is for when the value is None
+                return sorted(list((k, tree_sorted(v)) for k, v in obj.items()), key=lambda item: (item[1] is None, item))
```
ono-max, line 229: nit
```suggestion
                return sorted(
                    [(k, tree_sorted(v)) for k, v in obj.items()],
                    key=lambda item: (item[1] is None, item)
                )
```

ono-max (approved):
Thanks!

PR #821, opened 2024-03-26
No more close

kohsuke:
Eliminated the notion of closing a test session, which at this point was thoroughly meaningless anyway.

The one last remaining value of closing a test session is the metadata getting sent to the server. We move that data into the test data upload.

DO NOT MERGE yet. This change is coordinated with the corresponding server side change.

### DIFF
--- a/launchable/commands/record/tests.py
+++ b/launchable/commands/record/tests.py
@@ -438,7 +438,7 @@ def tests(
 
             # generator that creates the payload incrementally
             def payload(cases: Generator[TestCase, None, None],
-                        test_runner, group: str) -> Tuple[Dict[str, Union[str, List, bool]], List[Exception]]:
+                        test_runner, group: str) -> Tuple[Dict[str, Union[str, List, dict, bool]], List[Exception]]:
                 nonlocal count
                 cs = []
                 exs = []
--- a/launchable/commands/record/tests.py
+++ b/launchable/commands/record/tests.py
@@ -415,7 +415,6 @@ def tests(
 
         def run(self):
             count = 0  # count number of test cases sent
-            is_observation = False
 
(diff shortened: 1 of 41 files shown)


launchable/test_runners/launchable.py
```diff
@@ -5,7 +5,6 @@
 
 import click
 
-from launchable.app import Application
```
kohsuke, line 8: @Konboi your last change in this field left this dangling import statement. I wonder if your pre-commit is set up correctly? It should have caught this.

launchable/utils/launchable_client.py
```diff
@@ -39,7 +40,7 @@ def request(
         timeout: Tuple[int, int] = (5, 60),
         compress: bool = False,
         additional_headers: Optional[Dict] = None,
-    ):
+    ) -> requests.Response:
```
kohsuke, line 43: Technically this is not correct as we can sometimes return DryRunResponse, but this type signature makes auto-completion better. What's the right way to do this?

kohsuke:
Need to merge #822 first

launchable/test_runners/launchable.py
Konboi: Oh, really? Sorry, I'll check my pre-commit config.

launchable/commands/record/tests.py
```diff
@@ -510,11 +514,6 @@ def recorded_result() -> Tuple[int, int, int, float]:
                     send(p)
                     exceptions.extend(es)
 
-                res = client.request("patch", "{}/close".format(self.session),
```
Konboi, line 513: Goodbye close 👋

Konboi (approved):

PR #822, opened 2024-03-26
Pulling back dependencies

kohsuke:
Compatibility test that's supposed to catch a problem didn't, resulting in the breakage in the mainline CI. This change fixes the test, and pulls back some of the Java11 libraries.

I think we need some telemetry support to collect the Java runtime version our users are using, so that we can shore up the required version in an orderly manner. Java 8 is pretty old at this point.


Konboi (approved):
LGTM

> I think we need some telemetry support to collect the Java runtime version our users are using, so that we can shore up the required version in an orderly manner. Java 8 is pretty old at this point.

I totally agree with you

kohsuke:
Incorporated reversing #825 to roll forward.

PR #831, opened 2024-03-28
Ignore environmental LAUNCHABLE_BASE_URL

kohsuke:
When testing CLI with a locally run server, we use this environment variable. This affects the behaviour of unit tests, which is undesirable


ono-max:
Looks like CI failed. Can you please check it?

PR #835, opened 2024-04-08
Supports playwright json format report

Konboi:
We already provide a "playwright" plugin and support [JUnit reporter](https://playwright.dev/docs/test-reporters#junit-reporter) report. 

However, the JUnit reporter produces the final result of the test even if the test is retried. But [JSON reporter](https://playwright.dev/docs/test-reporters#json-reporter) includes all results when the test're executed multiple times.

So updated to support JSON reporter reports and add the `--json` option.

### DIFF
--- a/launchable/test_runners/playwright.py
+++ b/launchable/test_runners/playwright.py
@@ -117,7 +117,7 @@ class JSONReportParser:
                         status=status,
                         stdout=stdout,
                         stderr=stderr,
-                        data={"lineNumber": line_no}
+                        data={"lineNumber": line_no} if line_no else {}
                     ))
 
         return events
--- a/launchable/test_runners/playwright.py
+++ b/launchable/test_runners/playwright.py
@@ -1,17 +1,24 @@
 #
-# The the test runner to support playwright junit report format.
+# The the test runner to support playwright junit and JSON report format.
 # https://playwright.dev/
 #
+import json
+from typing import Dict, Generator, List
+
 import click
 from junitparser import TestCase, TestSuite  # type: ignore
 
+from ..commands.record.case_event import CaseEvent
 from ..testpath import TestPath
 from . import launchable
 
+TEST_CASE_DELIMITER = " › "
 
+
+@click.option('--json', 'json_format', help="use JSON report format", is_flag=True)
 @click.argument('reports', required=True, nargs=-1)
 @launchable.record.tests
-def record_tests(client, reports):
+def record_tests(client, reports, json_format):
     def path_builder(case: TestCase, suite: TestSuite,
                      report_file: str) -> TestPath:
         """
@@ -32,10 +39,15 @@ def record_tests(client, reports):
         if case.name:
             test_path.append({"type": "testcase", "name": case.name})
         return test_path
-    client.path_builder = path_builder
+
+    if json_format:
+        client.parse_func = JSONReportParser(client).parse_func
+    else:
+        client.path_builder = path_builder
 
     for r in reports:
         client.report(r)
+
     client.run()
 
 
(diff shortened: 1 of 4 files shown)


launchable/test_runners/playwright.py
```diff
@@ -1,17 +1,24 @@
...
+from typing import Dict, Generator, List
+
 import click
 from junitparser import TestCase, TestSuite  # type: ignore
 
+from ..commands.record.case_event import CaseEvent
 from ..testpath import TestPath
 from . import launchable
 
+TEST_CASE_DELIMITER = " › "
```
Konboi, line 15: To align JUnit format

launchable/test_runners/playwright.py
```diff
@@ -49,3 +61,85 @@ def subset(client):
...
 
 split_subset = launchable.CommonSplitSubsetImpls(__name__).split_subset()
+
+
+class JSONReportParser:
+    def __init__(self, client):
+        self.client = client
+
+    def parse_func(self, report_file: str) -> Generator[CaseEvent, None, None]:
+        data: Dict[str, Dict]
```
ono-max, line 71: Can we write example JSON file as a comment like https://github.com/launchableinc/cli/blob/main/launchable/test_runners/cucumber.py#L75?

Konboi: https://github.com/launchableinc/cli/pull/835/commits/033c6b5ae915fc6cc386f2c8aeb066bc062f1559

launchable/test_runners/playwright.py
```diff
@@ -49,3 +61,184 @@ def subset(client):
...
+        events = []
+
+        # In some cases, suites are nested.
+        for s in suite.get("suites", []):
+            for e in self._parse_suites(test_file, s, test_case_names + [s.get("title")]):
+                events.append(e)
+
+        for spec in suite.get("specs", []):
+            spec_name = spec.get("title", "")
+            line_no = spec.get("line", None)
```
ono-max, line 200: Sometimes, line number starts from 0 like pytest. Do you know that they start from 1 in playwright?

launchable/test_runners/playwright.py
```diff
@@ -49,3 +61,184 @@ def subset(client):
...
+                for result in test.get("results", []):
+                    test_path: TestPath = [
+                        self.client.make_file_path_component(test_file),
+                        {"type": "testcase", "name": TEST_CASE_DELIMITER.join(test_case_names + [spec_name])}
+                    ]
+
+                    duration_msec = result.get("duration", 0)
+                    status = self._case_event_status_from_str(result.get("status", ""))
+                    stdout = self._parse_stdout(result.get("stdout", []))
+                    stderr = self._parse_stderr(result.get("errors", []))
```
ono-max, line 211: nit

```suggestion
                    stdout = self._parse_stdout(result.get("stdout", ""))
                    stderr = self._parse_stderr(result.get("errors", ""))
```

launchable/test_runners/playwright.py
ono-max: Ok, thanks!

launchable/test_runners/playwright.py
ono-max, line 211: Can we assign an array to stdout and stderr?
Optional[str] is defined as a type in these variables.
https://github.com/launchableinc/cli/blob/main/launchable/commands/record/case_event.py#L152

Konboi: Please see `_parse_xxx` methods. These methods convert from array to string. 
Does it make sense?

tests/test_runners/test_playwright.py
```diff
@@ -10,9 +10,20 @@ class PlaywrightTest(CliTestCase):
...
         result = self.cli('record', 'tests', '--session', self.session,
                           'playwright', str(self.test_files_dir.joinpath("report.xml")))
 
         self.assert_success(result)
         self.assert_record_tests_payload('record_test_result.json')
+
+    @responses.activate
+    @mock.patch.dict(os.environ,
+                     {"LAUNCHABLE_TOKEN": CliTestCase.launchable_token})
+    def test_record_test_with_json_option(self):
```
ono-max, line 23: Could you please add the failed test case in tests/data/playwright/report.json?

Konboi: I wanted to do that and tried it... However, on the Windows environment, some characters didn't match correctly
I gave up  and ignored error case

ono-max: Oh, I see.

ono-max: Can we do something like this: https://github.com/launchableinc/cli/blob/main/tests/test_testpath.py#L94-L99?

Konboi: Can we skip Windows? If ok, I'll update it

ono-max: It should work.

https://docs.python.org/3/library/unittest.html#unittest.skipIf

ono-max (approved):

ono-max:
Thank you!

PR #846, opened 2024-05-02
minitest: Fetch name attribute as a classname

ono-max:
https://github.com/CircleCI-Public/minitest-ci is used for creating test report files. From the following code, it seems to be that a class name is set as a name attribute in testsuite tag.

* https://github.com/CircleCI-Public/minitest-ci/blob/v3.4.0/lib/minitest/ci_plugin.rb#L86-L87
* https://github.com/CircleCI-Public/minitest-ci/blob/v3.4.0/lib/minitest/ci_plugin.rb#L132

Here is a code snippet in minitest. 

```
require 'minitest/autorun'
require "mul"
class TestMul
  class Hoge < Minitest::Test
    def setup
      @mul = Mul.new
    end
    def test_calc_one
      assert_equal @mul.calc(1, 1), 1
    end
    def test_calc_two
      assert_equal @mul.calc(1, 2), 2
    end
    def test_calc_three
      assert_equal @mul.calc(1, 3), 3
    end
  end
end
```
After executing the above script with minitest-ci, I got the following results. As you can see, a class name is set as a name attribute in testsuite tag.

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuite time='0.000020' skipped='0' failures='0' errors='0' name="TestMul::Hoge" assertions='3' tests='3' timestamp="2024-05-02T14:41:57+09:00">
  <testcase time='0.000013' file="test/test_mul.rb" name="test_calc_one" assertions='1'>
  </testcase>
  <testcase time='0.000004' file="test/test_mul.rb" name="test_calc_two" assertions='1'>
  </testcase>
  <testcase time='0.000003' file="test/test_mul.rb" name="test_calc_three" assertions='1'>
  </testcase>
</testsuite>
```

### DIFF
--- a/launchable/test_runners/launchable.py
+++ b/launchable/test_runners/launchable.py
@@ -105,7 +105,6 @@ class CommonRecordTestImpls:
         @click.argument('source_roots', required=True, nargs=-1)
         def record_tests(client, source_roots):
             CommonRecordTestImpls.load_report_files(client=client, source_roots=source_roots, file_mask=file_mask)
-            client.run()
 
         return wrap(record_tests, record_tests_cmd, self.cmdname)
 
@@ -133,6 +132,8 @@ class CommonRecordTestImpls:
                 click.echo("No matches found: {}".format(root), err=True)
                 # intentionally exiting with zero
                 return
+
+        client.run()
 
 
 class CommonSplitSubsetImpls:
--- a/launchable/test_runners/launchable.py
+++ b/launchable/test_runners/launchable.py
@@ -104,13 +104,13 @@ class CommonRecordTestImpls:
 
         @click.argument('source_roots', required=True, nargs=-1)
         def record_tests(client, source_roots):
-            CommonRecordTestImpls.find_files(client=client, source_roots=source_roots, file_mask=file_mask)
+            CommonRecordTestImpls.load_report_files(client=client, source_roots=source_roots, file_mask=file_mask)
             client.run()
 
         return wrap(record_tests, record_tests_cmd, self.cmdname)
 
     @classmethod
-    def find_files(cls, client, source_roots, file_mask="*.xml"):
+    def load_report_files(cls, client, source_roots, file_mask="*.xml"):
         # client type: RecordTests in def launchable.commands.record.tests.tests
         # Accept both file names and GLOB patterns
         # Simple globs like '*.xml' can be dealt with by shell, but
--- a/launchable/test_runners/launchable.py
+++ b/launchable/test_runners/launchable.py
@@ -104,32 +104,35 @@ class CommonRecordTestImpls:
 
         @click.argument('source_roots', required=True, nargs=-1)
         def record_tests(client, source_roots):
-            # client type: RecordTests in def launchable.commands.record.tests.tests
-            # Accept both file names and GLOB patterns
-            # Simple globs like '*.xml' can be dealt with by shell, but
-            # not all shells consistently deal with advanced GLOBS like '**'
-            # so it's worth supporting it here.
-            for root in source_roots:
-                match = False
-                for t in glob.iglob(root, recursive=True):
-                    match = True
-                    if os.path.isdir(t):
-                        client.scan(t, file_mask)
-                    else:
-                        client.report(t)
-
-                if not match:
-                    # By following the shell convention, if the file doesn't exist or GLOB doesn't match anything,
-                    # raise it as an error. Note this can happen for reasons other than a configuration error.
-                    # For example, if a build catastrophically failed and no
-                    # tests got run.
-                    click.echo("No matches found: {}".format(root), err=True)
-                    # intentionally exiting with zero
-                    return
-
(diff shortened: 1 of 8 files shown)


launchable/test_runners/launchable.py
```diff
@@ -104,33 +104,36 @@ def report_files(self, file_mask="*.xml"):
...
-                    # intentionally exiting with zero
-                    return
-
+            CommonRecordTestImpls.find_files(client=client, source_roots=source_roots, file_mask=file_mask)
             client.run()
 
         return wrap(record_tests, record_tests_cmd, self.cmdname)
 
+    @classmethod
+    def find_files(cls, client, source_roots, file_mask="*.xml"):
```
Konboi, line 113: nits & not strong opinion: I prefer `load_report_files` to `find_files`

tests/data/minitest/record_test_result.json
```diff
@@ -7,6 +7,10 @@
         {
           "type": "testcase",
           "name": "test_should_save_user_with_name"
+        },
+        {
+          "type": "class",
+          "name": "UserCopyTest"
```
Konboi, line 13: This components will be `file=test/models/user_copy_test.rb#testcase=test_should_save_user_with_name#class=UserCopyTest` as test path

but I think `file=test/models/user_copy_test.rb#class=UserCopyTest#testcase=test_should_save_user_with_name` is correct.

Could you check it?

tests/test_runners/test_googletest.py
```diff
@@ -43,5 +43,5 @@ def test_record_empty_dir(self):
         path = 'latest/gtest_*_results.xml'
         result = self.cli('record', 'tests', '--session', self.session,
                           'googletest', path)
-        self.assertEqual(result.output.rstrip('\n'), "No matches found: {}".format(path))
+        self.assertIn("No matches found: {}".format(path), result.output.rstrip('\n'))
```
Konboi, line 46: Q: why do we need to change this line?

tests/data/minitest/record_test_result.json
ono-max: Good catch. Let me fix it. Thanks!

launchable/test_runners/launchable.py
ono-max: [ed09dfa](https://github.com/launchableinc/cli/pull/846/commits/ed09dfa20fcecdc4b60226f8eceeb7746e898725)

tests/data/minitest/record_test_result.json
ono-max: [7bd4ddc](https://github.com/launchableinc/cli/pull/846/commits/7bd4ddc3d85994f4920bbb649796a9210ad6318d)

tests/test_runners/test_googletest.py
ono-max: Thank you for asking this question. It's because I called client.run() in `record_tests method`, not `load_report_files`. I've fixed it.

[074d915](https://github.com/launchableinc/cli/pull/846/commits/074d915c1cb79cdbfacad92c7e307e4f46025e94)

tests/data/minitest/record_test_result.json
```diff
@@ -22,6 +26,10 @@
         {
           "type": "testcase",
           "name": "test_should_not_save_user_without_name"
+        },
+        {
+          "type": "class",
+          "name": "UserCopyTest"
```
Konboi, line 32: it seems result json order is still file -> testcase -> class

Konboi: Is it correct??

ono-max: Yes, it is.

ono-max: It's because `assert_json_orderless_equal` is used at  https://github.com/launchableinc/cli/blob/main/tests/cli_test_case.py#L239.

ono-max: To test the order, I've added https://github.com/launchableinc/cli/pull/846/files#diff-c85217de42a62611db47dc8b7f9c6a9efc83f793379b372b3b1c65dd772a2f68R23-R43.

Konboi: I see but I think it's my bad https://github.com/launchableinc/cli/pull/817

Because the test paths should be ordered.
we (Launchable) cannot allow miss ordering e.g.) `file -> testcase -> class`.

You don't have to fix `assert_json_orderless_equal` now. But could you update this result JSON to the correct order to prevent misunderstanding?

ono-max: Let me fix it in another PR.

Konboi: `assert_json_orderless_equal` -> another PR
chunk_json files -> this PR

OK?

ono-max: I plan to fix both changes in another PR since we can check if it works.

Konboi: OK

Konboi (approved):

PR #853, opened 2024-05-08
[IB-395] capture the mechanism in which a workspace information was obtained

kohsuke:
This would allow us to guide the server side behaviour later.

@ono-max can you double check your pre-commit config on this workspace? I had to create 6e5997792783c00a3ba8c720c57494013e36a01f  because my local flake8 run detected unused import statements, which I assume was caused by your recent edits to those files. I expected that when you committed those changes, your pre-commit would have run flake8 that should have fixed them automatically for you.


ono-max (approved):
Thank you for mentioning that. Yes, my pre-commit setting is broken, and I leave it broken 😅 It's good chance to fix my configuration.

PR #856, opened 2024-05-13
Add assertion to check if the order of test paths is correct

ono-max:
Currently, we do not verify the order of JSON at https://github.com/launchableinc/cli/blob/main/tests/cli_test_case.py#L221 because we evaluate JSON in an order-agnostic manner. However, we aim to ensure the order of test paths is accurate. `assert_json_orderless_equal` disregards the order of items intentionally. Consequently, I have introduced the `assert_test_path_orderly_equal` method to address this requirement, instead of modifying `assert_json_orderless_equal`.

@kohsuke 

I think you designed these methods in https://github.com/launchableinc/cli/pull/91. Can you take a look at this PR?

### DIFF
--- a/tests/cli_test_case.py
+++ b/tests/cli_test_case.py
@@ -7,6 +7,7 @@ import tempfile
 import types
 import unittest
 from pathlib import Path
+from typing import Any, Dict, List
 
 import click  # type: ignore
 import responses  # type: ignore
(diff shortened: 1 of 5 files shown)


tests/cli_test_case.py
```diff
@@ -294,3 +295,15 @@ def tree_sorted(obj):
                 return obj
 
         self.assertEqual(tree_sorted(a), tree_sorted(b))
+
+    def assert_test_path_orderly_equal(self, a, b):
```
kohsuke, line 299: Let's define the contract of this method. What exactly does this method assert?

I ask because I was expecting `a` and `b` are symmetric, but L308-309 doesn't work that way. For exampe, if `a` is empty list, this assertion passes no matter what `b` is.

I'm almost tempted to just test `a['events']['testPath']==b['events']['testPath']`; Is that not what we want to do?

tests/test_runners/test_minitest.py
```diff
@@ -20,27 +20,6 @@ def test_record_test_minitest(self):
         self.assert_success(result)
         self.assert_record_tests_payload('record_test_result.json')
 
-    # For testing https://github.com/launchableinc/cli/pull/846/files#r1591707246.
```
kohsuke, line 23: Is this removal intentional?

tests/cli_test_case.py
ono-max: > I'm almost tempted to just test a['events']['testPath']==b['events']['testPath']; Is that not what we want to do?

I initially planed to do, but it's not possible because the value associated with the 'events' key is in random order.

kohsuke:
Thanks! There's some loose ends when duplicates are present but now the intent is clar, so that's OK for now.

PR #861, opened 2024-06-04
Add --test-suite option to record session

takanabe:
# Background

When we introduced a workspace, we expect that customers use a workspace per test suite. However, we  noticed that some customers use a single workspace with multiple test suites. Mixing test suites in a workspace make many features less valuable. For example, trends, test issue triages, test session list, and unhealthy tests are affected. It’s natural to promote “test suites” as a first class data model on Launchable so our customers can sort, filter, and analyze data based on test suites in a workspace.

# Change

- Add a new option `--test-suite` to specify a name of test suite executed in a test session.

### DIFF
--- a/launchable/commands/record/session.py
+++ b/launchable/commands/record/session.py
@@ -90,6 +90,14 @@ def _validate_session_name(ctx, param, value):
     type=str,
     metavar='LINEAGE',
 )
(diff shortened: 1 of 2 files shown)


Konboi (approved):
LGTM

PR #868, opened 2024-06-13
Support markers in pytest

ono-max:
This PR supports https://docs.pytest.org/en/stable/how-to/mark.html in Launchable CLI. Here is an example of an XML file when using marker object.

```
<?xml version="1.0" encoding="utf-8"?>
<testsuites>
    <testsuite name="pytest" errors="0" failures="4" skipped="0" tests="12" time="0.035"
        timestamp="2024-06-13T12:33:48.703447" hostname="ip-192-168-0-28.us-west-2.compute.internal">
        <testcase classname="tests.fooo.filenameonly_test" name="test_x"
            file="tests/fooo/filenameonly_test.py" line="0" time="0.000" />
        <testcase classname="tests.test_funcs1" name="test_func1" file="tests/test_funcs1.py"
            line="2" time="0.000">
            <properties>
                <property name="name" value="foo" />
                <property name="args" value="()" />
                <property name="kwargs" value="{}" />
            </properties>
        </testcase>
...
```

### DIFF
--- a/launchable/test_runners/pytest.py
+++ b/launchable/test_runners/pytest.py
@@ -3,7 +3,7 @@ import json
 import os
 import pathlib
 import subprocess
-from typing import Dict, Generator, List
+from typing import Generator, List
 
 import click
 from junitparser import Properties, TestCase  # type: ignore
@@ -148,13 +148,8 @@ def record_tests(client, json_report, source_roots):
                     </properties>
                 ```
             """
-            markers = []
-            marker: Dict[str, str] = {}
-            for prop in props:
-                markers.append({"name": prop.name, "value": prop.value})
-            if len(marker) > 0:
-                markers.append(marker)
-            result["markers"] = markers
+            markers = [{"name": prop.name, "value": prop.value} for prop in props]
+            result["markers"] = markers if markers else []
 
         metadata = MetadataTestCase.fromelem(case)
         if metadata and metadata.line is not None:
--- a/launchable/test_runners/pytest.py
+++ b/launchable/test_runners/pytest.py
@@ -3,10 +3,10 @@ import json
 import os
 import pathlib
 import subprocess
-from typing import Generator, List
+from typing import Dict, Generator, List
 
 import click
-from junitparser import Properties, TestCase
+from junitparser import Properties, TestCase  # type: ignore
 
 from launchable.commands.record.case_event import CaseEvent, CaseEventType, MetadataTestCase
 from launchable.testpath import TestPath
@@ -136,7 +136,7 @@ def record_tests(client, json_report, source_roots):
         result = {}
         if props is not None:
             markers = []
-            marker = {}
+            marker: Dict[str, str] = {}
             for prop in props:
                 if prop.name in marker:
                     """
--- a/launchable/test_runners/pytest.py
+++ b/launchable/test_runners/pytest.py
@@ -5,15 +5,10 @@ import pathlib
 import subprocess
 from typing import Generator, List
 
-from junitparser import TestCase
-
 import click
+from junitparser import Properties, TestCase
 
-from launchable.commands.record.case_event import (
-    CaseEvent,
-    CaseEventType,
-    MetadataTestCase,
-)
+from launchable.commands.record.case_event import CaseEvent, CaseEventType, MetadataTestCase
 from launchable.testpath import TestPath
 
 from . import launchable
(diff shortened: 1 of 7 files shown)


ono-max:
Oops

ono-max:
@Konboi 

Sorry, the CI failed now, I'm going to mention you after fixing it.

ono-max:
@Konboi 

This PR is ready for review 👍

tests/data/pytest/pytest.ini
```diff
@@ -0,0 +1,8 @@
+[pytest]
+; xunit1 is necessary for adding line attribute.
+junit_family = xunit1
+markers =
```
Konboi, line 4: Why do we need to add the `markers` property??

tests/data/pytest/tests/conftest.py
```diff
@@ -0,0 +1,6 @@
+def pytest_collection_modifyitems(session, config, items):
+    for item in items:
```
Konboi, line 2: What is your intent for adding these properties manually??

launchable/test_runners/pytest.py
```diff
@@ -137,14 +132,41 @@ def _pytest_formatter(test_path):
...
+                                <property name="kwargs" value="{}" />
+                            </properties>
+                        ```
+                    """
+                    markers.append(marker.copy())
+                    marker.clear()
+                marker[prop.name] = prop.value
+            if len(marker) > 0:
+                markers.append(marker)
+            result["markers"] = markers
```
Konboi, line 162: Q. Why did you use `markers` as the field name?
I thought ‘properties’ (or annotations) would match the field name better.

launchable/test_runners/pytest.py
```diff
@@ -137,14 +132,41 @@ def _pytest_formatter(test_path):
...
+                                <property name="name" value="parametrize" />
+                                <property name="args" value="('y', [2, 3])" />
+                                <property name="kwargs" value="{}" />
+                                <property name="name" value="parametrize" />
+                                <property name="args" value="('x', [0, 1])" />
+                                <property name="kwargs" value="{}" />
+                            </properties>
+                        ```
+                    """
+                    markers.append(marker.copy())
```
Konboi, line 157: I'm not sure best marker's data format. But it seems this logic depends on the order of properties. 
If grouping is not important, we can adopt test path like data format.

```
[
  {"type": "name", "value": "parametrize"},
  {"type": "arg", "value:" ('y', [2,3])"},
  {"type": "kwargs", "value", "{}"
   ...
]
```

WDTY??

tests/data/pytest/pytest.ini
ono-max: To register markers.

https://docs.pytest.org/en/stable/example/markers.html#registering-markers

tests/data/pytest/tests/conftest.py
ono-max: To add these information to the XML file.

launchable/test_runners/pytest.py
Konboi: Sorry, please forget this comment, I checked only XML outputs. 
The user needs to add `@pytest.mark` if they want to add. Also I found pytest provides `--markers`  option
The --markers option always gives you a list of available markers:

Thus, the `makrers` of field name isn't bad.

tests/data/pytest/pytest.ini
Konboi: Could leave the ULR as comment?

tests/data/pytest/tests/conftest.py
Konboi: I couldn't understand the motivation that you want to add these information to the XML files. Could you tell me that?

Konboi: Ops, this is not a test case? Are these configurations for all the test cases?

ono-max: > Ops, this is not a test case? Are these configurations for all the test cases?

Yes it is.

Konboi: TIL: https://stackoverflow.com/questions/34466027/what-is-conftest-py-for-in-pytest

launchable/test_runners/pytest.py
ono-max: Okay, let me fix it based on your proposal.

launchable/test_runners/pytest.py
```diff
@@ -137,14 +132,36 @@ def _pytest_formatter(test_path):
...
+                        <property name="name" value="parametrize" />
+                        <property name="args" value="('x', [0, 1])" />
+                        <property name="kwargs" value="{}" />
+                    </properties>
+                ```
+            """
+            markers = []
+            marker: Dict[str, str] = {}
+            for prop in props:
+                markers.append({"name": prop.name, "value": prop.value})
```
Konboi, line 154: I guess you want to `marker.append(...)` not `markers`, right?

Konboi: You can write 

```python
if props is not None:
    markers = [{"name": prop.name, "value": prop.value} for prop in props]
    result["markers"] = markers if markers else []
```

ono-max: Nice suggestion. Let me fix it.

Konboi (approved):

PR #873, opened 2024-06-13
Set renovate package group rule

Konboi:
Want to update these jackson's updates at the same time

https://github.com/launchableinc/cli/pull/848
https://github.com/launchableinc/cli/pull/849
https://github.com/launchableinc/cli/pull/851

So, try to set renovate config ref: https://docs.renovatebot.com/configuration-options/#matchpackageprefixes


Konboi:
Don't need to review deeply

ono-max (approved):

PR #883, opened 2024-06-21
fix windowsh path

Konboi:
I'm not sure why this test was broken by https://github.com/launchableinc/cli/pull/882
But I updated to use `os.path` to assemble file paths


ono-max (approved):
Thanks!

PR #888, opened 2024-06-28
Improve smooth error handling

Konboi:
In the raw profile, we used `dict['key']` method with `or <default value>`. However if it doesn't define a target key, a default value won't used and `KeyError: '<TARGET KEY'` exception will be thrown.
So, I fixed to use `.get(<KEY>, <DEFAULT VALUE>)` instead of `dict['key'] or <DEFAULT VALUE>` style.

And, fixed when the user set duration as string type, we'll covert it to float (number)


ono-max (approved):
thanks!

PR #889, opened 2024-07-09
Add the option to collect commit messages

ono-max:
Currently, Launchable CLI does not collect commit messages because of security concern, but we'll add the option to collect them since we plan to display commit information in the test session detail page. In the default, we don't collect commit messages.

https://app.launchableinc.com/organizations/demo/workspaces/demov2/data/test-sessions/193372?tab=commits is an example, we'll display the commit message instead of the commit hash if we have the commit message data.

### DIFF
--- a/launchable/commands/record/commit.py
+++ b/launchable/commands/record/commit.py
@@ -109,9 +109,7 @@ def exec_jar(source: str, max_days: int, app: Application, commit_message: bool)
         "{}/intake/".format(base_url),
         "-max-days",
         str(max_days),
-        "-scrub-pii",
-        "-commit-message",
-        str(commit_message)
+        "-scrub-pii"
     ])
 
     if Logger().logger.isEnabledFor(LOG_LEVEL_AUDIT):
@@ -120,6 +118,8 @@ def exec_jar(source: str, max_days: int, app: Application, commit_message: bool)
         command.append("-dry-run")
     if app.skip_cert_verification:
         command.append("-skip-cert-verification")
+    if commit_message:
+        command.append("-commit-message")
     command.append(cygpath(source))
 
     subprocess.run(
--- a/launchable/commands/record/commit.py
+++ b/launchable/commands/record/commit.py
@@ -61,11 +61,11 @@ def commit(ctx, source: str, executable: bool, max_days: int, scrub_pii: bool, i
     client = LaunchableClient(tracking_client=tracking_client, app=ctx.obj)
 
     # Commit messages are not collected in the default.
-    commit_message = False
+    is_collect_message = False
     try:
         res = client.request("get", "commits/collect/options")
         res.raise_for_status()
-        commit_message = res.json().get("commitMessage", False)
+        is_collect_message = res.json().get("commitMessage", False)
     except Exception as e:
         tracking_client.send_error_event(
             event_name=Tracking.ErrorEvent.INTERNAL_CLI_ERROR,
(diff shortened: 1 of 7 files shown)


ono-max:
@Konboi 

I'd like you to review the following Java parts since it's my first time to write Java in CLI.

* https://github.com/launchableinc/cli/pull/889/files#diff-2dff6ba117c78022a5556422e0e1943929637b67b9ea65a4b1b232c6646ae449
* https://github.com/launchableinc/cli/pull/889/files#diff-217df1a570eb90c72b1189825a59a9f724f3140b3bcc84a98cc76dedebf051b8
* https://github.com/launchableinc/cli/pull/889/files#diff-b4627b715f46d2fb5d0bebdd74fac96f629eae59e3688c38448720d726666e44

src/main/java/com/launchableinc/ingest/commits/CommitGraphCollector.java
```diff
@@ -268,6 +270,10 @@ private CloseableHttpResponse handleError(URL url, CloseableHttpResponse respons
     return response;
   }
 
+  public void skipCommitMessage(boolean skipCommitMessage) {
```
Konboi, line 273: If my understanding is correct, this behavior seems strange...

When the `commitMessage` flag is true, skipCommitMessage value will be `true`. Isn't this the opposite...??


Are there any reasons that you didn't implement like the below?
```java

public void collectCommitMessage(boolean commitMessage) {
   this.collectCommitMessage
}

...

c.setMessage(collectCommit ? r.getFullMessage() : ""):
```

launchable/commands/record/commit.py
```diff
@@ -90,6 +110,8 @@ def exec_jar(source: str, max_days: int, app: Application):
         "-max-days",
         str(max_days),
         "-scrub-pii",
+        "-commit-message",
```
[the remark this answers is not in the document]
ono-max: That's right. Thank you. I've updated it.

src/main/java/com/launchableinc/ingest/commits/CommitGraphCollector.java
ono-max: I borrowed this name from the old code in mothership repository.

ono-max: If you prefer the above implementation to this current implementation, I'll modify it. Please let me know.

ono-max: I've modified based on your suggestion, and confirmed it works in my local environemnt.

ono-max:
@Konboi 

Can you review this PR again? Thanks!

launchable/commands/record/commit.py
```diff
@@ -54,8 +57,25 @@ def commit(ctx, source: str, executable: bool, max_days: int, scrub_pii: bool, i
...
+        res = client.request("get", "commits/collect/options")
+        res.raise_for_status()
+        is_collect_message = res.json().get("commitMessage", False)
+    except Exception as e:
+        tracking_client.send_error_event(
+            event_name=Tracking.ErrorEvent.INTERNAL_CLI_ERROR,
+            stack_trace=str(e),
+            api="commits/options",
+        )
+        client.print_exception_and_recover(e)
```
Konboi, line 75: Note For Me:
Even if errors happened when request to `commits/collect/options` API, only is_collect_message flag value will be `False`

Konboi (approved):
LGTM

PR #892, opened 2024-07-19
Improve diagnosability

kohsuke:
Allow the server side to map which API requests map to which client invocation.

User-agent header can be seen & processed in Datadog. Often we struggle to understand how customers are integrating us in the CI script, and this will give us a key insight.


Konboi (approved):
This change is LGTM
But please fix type error like https://github.com/launchableinc/cli/pull/893 before merging this PR

Konboi (approved):

PR #955, opened 2024-12-24
Resolving dependency conflicts

kohsuke:
Pipfile specifies a set of constraints on dependencies, and therefore as newer versions of dependencies release, how they resolve will change.

My understanding is that, as a result of this, some time ago our `pipenv lock` stopped functioning. This has been bothering me for a while, but finally I bit the bullet and resolved it.

To prevent this problem further, I'm also committing Pipfile.lock, which stores the results of dependency resolutions. This way, we can guarantee that we always have a set of versions that satisfies constraints.

As noted in the comment, it is possible that this can create different problems around platform dependent stuff. But so far, we don't seem to have that issue.

## Summary by CodeRabbit

- **Documentation**
	- Updated comments in the `Pipfile` to provide guidance on managing dependencies and resolving conflicts.

- **Chores**
	- Removed outdated comments and entries related to `pipenv` and `Pipfile.lock` from the `.gitignore` file.

- **Bug Fixes**
	- Adjusted version constraints for `setuptools`, `importlib-metadata`, and `more_itertools` to prevent conflicts with newer versions.
	- Added `click` package to the development dependencies.

- **New Features**
	- Enhanced dependency management in the workflow by adding a step to generate a `Pipfile.lock` for consistent package versions across environments.
	- Introduced a command to list installed packages, aiding in dependency resolution.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -69,6 +69,8 @@ jobs:
         max_attempts: 3
         timeout_minutes: 5
         retry_on: error
+        # For the discussion on why the added `pipenv lock` is necessary/warranted,
+        # see a lengthy note in Pipfile from Kohsuke dated 2024-12-24.
         command: |
           python -m pip install --upgrade pip
           pip install pipenv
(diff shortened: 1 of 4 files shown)


kohsuke:
I'm requesting multiple reviews to make sure this doesn't have unintended consequences

kohsuke:
Time to cut the loss.

I thought I got this working, then it stopped working. I don't understand how that behavior difference is happening. This is already a way bigger yakshaving. Too bad I couldn't prevail.

PR #973, opened 2025-01-26
subset: relax validation rule

ono-max:
[Ruby team](https://github.com/ruby/ruby) tries to use Zero Input Subsetting in the observation mode for their CI environment, but using both observation mode and Zero Input Subsetting are banned by CLI.

This validation rule has been introduced since https://github.com/launchableinc/cli/pull/435/files#diff-4c62e35fb4e3b2b6bff8593107cdddd1fd47eff90293edc9d838c9b2f3cff94aR138.

After discussion with @Konboi , we reached to the accord that we can relax the validation rule without any problems.

## Summary by CodeRabbit

- **Bug Fixes**
  - Modified flag handling in subset command to remove previous restrictions on conflicting options
  - Simplified option processing logic while maintaining core functionality

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -218,20 +218,6 @@ def subset(
     app = context.obj
     tracking_client = TrackingClient(Tracking.Command.SUBSET, app=app)
 
-    if is_observation and is_get_tests_from_previous_sessions:
-        msg = "Cannot use --observation and --get-tests-from-previous-sessions options at the same time"
-        click.echo(
-            click.style(
-                msg,
-                fg="red"),
-            err=True,
-        )
-        tracking_client.send_error_event(
-            event_name=Tracking.ErrorEvent.INTERNAL_CLI_ERROR,
-            stack_trace=msg,
-        )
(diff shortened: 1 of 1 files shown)


launchable/commands/subset.py
```diff
@@ -218,20 +218,6 @@ def subset(
     app = context.obj
     tracking_client = TrackingClient(Tracking.Command.SUBSET, app=app)
 
-    if is_observation and is_get_tests_from_previous_sessions:
```
Konboi, line 221: I gave it some thought, we should change this validation
from `if is_observation and is_get_tests_from_previous_sessions:` to `if is_observation and is_output_exclusion_rules:` to prevent miss integration.

ono-max: We've already checked those condition at https://github.com/launchableinc/cli/blob/main/launchable/commands/subset.py#L235, don't we?

Konboi: I didn't know that. Thanks

Konboi (approved):

PR #984, opened 2025-02-07
[LCHIB-544] Remove the debug probe

kohsuke:
Now that I figured out the root cause, this probe is no longer needed.

## Summary by CodeRabbit

- **Chores**
	- Removed redundant debug logging to simplify internal commit tracking without affecting any user functionality.


Konboi (approved):
> Now that I figured out the root cause

🎉

PR #991, opened 2025-03-03
[LCHIB-562] Max days override

kohsuke:
Making the future commit re-collection easier by overriding the max days from the server side.


Konboi (approved):
Make sense!

PR #1005, opened 2025-05-22
Support nested classes in Maven plugin

ninjinkun:


takanabe:
LGTM but I will involve another member just in case.

ninjinkun:
I found the test failure to be caused by Click's breaking change since v8.2. I'm going to fix this issue on another PR.

ninjinkun:
To pass the test, I limited the Click version up to 8.1. I noted the issue https://github.com/launchableinc/cli/issues/1006.

launchable/test_runners/maven.py
```diff
@@ -113,4 +115,24 @@ def format_same_bin(s: str) -> List[Dict[str, str]]:
 #
 # So to collectly find tests without duplications, we need to find surefire-reports/TEST-*.xml
 # not surefire-reports/**/TEST-*.xml nor surefire-reports/*.xml
-record_tests = launchable.CommonRecordTestImpls(__name__).report_files(file_mask="TEST-*.xml")
+@click.argument('reports', required=True, nargs=-1)
+@launchable.record.tests
+def record_tests(client, reports):
+    default_path_builder = client.path_builder
+
+    def path_builder(case: TestCase, suite: TestSuite,
```
Konboi, line 123: It seems this logic is same with the gradle one.

So, how about defining the function and use from the maven and the gradle profile like the [nunit](https://github.com/launchableinc/cli/blob/3b003a106f3dca7eae52c966b62a25a6c2916cfd/launchable/test_runners/nunit.py#L91-L124) and [dotnet](https://github.com/launchableinc/cli/blob/3b003a106f3dca7eae52c966b62a25a6c2916cfd/launchable/test_runners/dotnet.py#L8)?

```python
def junit5_parse_func( ...)
   ...
```

Konboi (approved):
LGTM

PR #1015, opened 2025-05-30
Add request configuration with socket timeout to CommitGraphCollector

ono-max:


kohsuke (approved):
You can ship this as-is, if you want to, but the following changes would make this even better:

I don't see the need for a switch to change timeout. Especially if it's only affecting commit collection, but not build/test recording, etc. Python code already uses (5,60) and (5,15) timeout combo depending on the HTTP method (see `http_client.py`), and there's no way to control this. So more natural, consistent thing to do is to use the same values.

(Speaking of `http_client.py`, the way it's dealing with the default value is just a mess.)

PR #1025, opened 2025-06-05
Validate the `--timestamp` option value

Konboi:
`--timestamp` option is for importing historical data. So, added validation not to be able to set feature time


ono-max (approved):
Since I'm busy right now, I'll trust your code and approve your PR if it's urgent.

PR #1027, opened 2025-06-05
Improvement around git-rev-parse

kohsuke:
This is the problem they run into. Just `print(e)` is a bad idea because it doesn't put them on the right track to solve them.

```
$ launchable record commit 
[Errno 2] No such file or directory: 'git'
```

(In this case, the added injury is that we are requiring `git` that shouldn't really be needed, but I didn't tackle that problem here)


ono-max (approved):

ono-max:
Thank you 🙏

PR #1029, opened 2025-06-06
Support XCTest 

Konboi:
# What

Implemented to the XCTest profile. XCTest is the test framework for XCode project ref: https://developer.apple.com/documentation/xctest

This is an example project to try XCTest profile https://github.com/launchableinc/examples/pull/87

# How to use 

## subset

Since there aren't any solutions to list test target without executing test, we'll support subset with the [Zero Input Subsetting](https://www.launchableinc.com/docs/features/predictive-test-selection/requesting-and-running-a-subset-of-tests/subsetting-with-the-launchable-cli/zero-input-subsetting/). And we'll use the subset result with `-skip-testing` option of XCTest

```sh
$ launchable subset --target 80% --get-tests-from-previous-sessions --output-exclusion-rules xctest > subset.txt
$ cat $subst.txt
-skip-testing:XCTestSampleTests/XCTestSampleTests
-skip-testing:XCTestSampleTests/XCTestSampleTests2
$ xcodebuild test -scheme <SCHEMA> -destination <DESTINATION> -parallel-testing-enabled NO $(subset.txt)
```

## record build

XCTest profile requires [xcpretty](https://github.com/xcpretty/xcpretty) to produce test report file.

```sh
$ xcodebuild test -scheme <SCHEMA> -destination <DESTINATION> -parallel-testing-enabled NO | xcpretty -r junit && exit ${PIPESTATUS[0]}
# xcpretty produces JUnit Report under the build/reports/ as default
$ launchable record tests xctest build/reports/*.xml
```

### DIFF
--- a/launchable/test_runners/xctest.py
+++ b/launchable/test_runners/xctest.py
@@ -18,14 +18,13 @@ def record_tests(client, reports):
 
         for testsuite in root.findall('testsuite'):
             for testcase in testsuite.findall('testcase'):
-                print(testcase)
                 failure = testcase.find('failure')
                 if failure is not None:
                     # Note(Konboi): XCTest escape `"` to `&quot;` in the message, so save as unescaped text
                     message = html.unescape(failure.get('message', ''))
                     body = failure.text or ''
                     failure.text = f"{message}\n{body}"
-                    print("failure", failure.text)
+
         return tree
 
     def path_builder(case: TestCase, suite: TestSuite, report_path: str) -> TestPath:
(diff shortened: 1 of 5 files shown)


launchable/test_runners/xctest.py
```diff
@@ -0,0 +1,69 @@
...
+                    body = failure.text or ''
+                    failure.text = f"{message}\n{body}"
+
+        return tree
+
+    def path_builder(case: TestCase, suite: TestSuite, report_path: str) -> TestPath:
+        class_attr = case.classname
+        splits = class_attr.split('.')
+        target = splits[0]
+        class_name = '/'.join(splits[1:])
```
ninjinkun, line 34: [Nit] You can write like this.
```suggestion
        [target, *rest] = class_attr.split('.')
        class_name = '/'.join(rest)
```

PR #1030, opened 2025-06-09
Automate publishing Docker image using GitHub Actions

ono-max:
Currently, we manually publish Docker image, but it would be better to automate publishing Docker image using GH Actions because it makes easier to maintain our development workflow.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -25,7 +25,7 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-  deploy:
+  pypl:
     needs: tagpr
     if: needs.tagpr.outputs.tag != '' || github.event_name == 'workflow_dispatch'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -67,3 +67,44 @@ jobs:
...
+          platforms:
+            - linux/amd64
+            - linux/arm64
+          push: true
+          tags: cloudbees/launchable:${{ needs.tagpr.outputs.tag }}
+
+      - name: Generate artifact attestation
+        uses: actions/attest-build-provenance@db473fddc028af60658334401dc6fa3ffd8669fd # v2.3.0
+        with:
+          subject-name: index.docker.io/my-docker-hub-namespace/my-docker-hub-repository
```
Konboi, line 108: Is it correct?

ono-max: Oh, it's not correct. Thank you for pointing it out.

.github/workflows/python-publish.yml
```diff
@@ -67,3 +70,44 @@ jobs:
...
+          platforms:
+            - linux/amd64
+            - linux/arm64
+          push: true
+          tags: ${{ env.IMAGE_NAME }}\:${{ needs.tagpr.outputs.tag }}
+
+      - name: Generate artifact attestation
+        uses: actions/attest-build-provenance@db473fddc028af60658334401dc6fa3ffd8669fd # v2.3.0
+        with:
+          subject-name: index.docker.io/${{ env.IMAGE_NAME }}
```
Konboi, line 111: > NOTE: When pushing to Docker Hub, please use "index.docker.io" as the registry portion of the image name.

https://github.com/actions/attest-build-provenance

Konboi (approved):

PR #1034, opened 2025-06-09
[AIENG-168] Fix Docker image tag formatting in workflow

ono-max:
https://github.com/launchableinc/cli/actions/runs/15527660601/job/43710216131
```
  Build summary requires a build reference
Error: buildx failed with: ERROR: invalid tag "cloudbees/launchable\\:v1.101.5": invalid reference format
```

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -103,7 +103,7 @@ jobs:
             - linux/amd64
             - linux/arm64
           push: true
-          tags: ${{ env.IMAGE_NAME }}\:${{ needs.tagpr.outputs.tag }}
+          tags: ${{ env.IMAGE_NAME }}:${{ needs.tagpr.outputs.tag }}
 
       - name: Generate artifact attestation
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -103,7 +103,7 @@ jobs:
             - linux/amd64
             - linux/arm64
           push: true
-          tags: ${{ env.IMAGE_NAME }}\:${{ needs.tagpr.outputs.tag }}
+          tags: ${{ env.IMAGE_NAME }}:${{ needs.tagpr.outputs.tag }}
```
ono-max, line 106: Screenshot 2025-06-09 at 14 48 28

Screenshot 2025-06-09 at 14 48 23

takanabe (approved):

.github/workflows/python-publish.yml
ono-max: When I see these color difference, I felt like I need to escape colon. However, it's not necessary, according to the log.

PR #1045, opened 2025-06-16
Support subtest case in go-test profile

Konboi:
# What 

In Go, we can define subtests like the example below.
However, in the test report, they are shown as a single test case using the format: `test function name / subtest name`.
But when using the `go test -list` command, which is used for subsetting, only the test function names are output—subtest names are not included.
So we need to update the Go test profile to support subtests.

e.g) subtest example
```go
t.Run(tt.title, func(t *testing.T) {...}
```

go test -llist result

```
go test -list="Test|Example" ./...
TestExample1
TestExample2
TestExample3
TestExample4
TestExample5
ok      github.com/launchableinc/examples/go    0.287s
TestGreeting
ExampleGreeting
ok      github.com/launchableinc/examples/go/example    0.475s
```

report example
```xml
        <testcase name="TestExample5/Test_case_2_(fail_case)" classname="github.com/launchableinc/examples/go" time="0.000">
            <failure message="Failed"><![CDATA[    example3_test.go:38: expected 8, got 7]]></failure>
        </testcase>
```

Created data used by this PR: https://github.com/launchableinc/examples/pull/89

### DIFF
--- a/launchable/test_runners/go_test.py
+++ b/launchable/test_runners/go_test.py
@@ -61,11 +61,23 @@ def record_tests(client, source_roots):
     def path_builder(case: TestCase, suite: TestSuite, report_file: str) -> TestPath:
         tp = default_path_builder(case, suite, report_file)
         for tpc in tp:
-            if 'type' in tpc and 'name' in tpc and tpc['type'] == 'class':
-                # go-junit-report v2 reports full package name. go-junit-report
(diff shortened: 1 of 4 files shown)


Konboi:
@ninjinkun or @ono-max Could you approve it? You don't have to review carefully

ono-max (approved):

PR #1068, opened 2025-07-05
Add/remove/edit existing commands

ninjinkun:


tests/commands/record/test_build.py
```diff
@@ -29,7 +29,20 @@ def test_submodule(self, mock_check_output):
...
-        result = self.cli("record", "build", "--no-commit-collection", "--build", self.build_name)
+        result = self.cli(
+            "record",
+            "build",
+            "--no-commit-collection",
+            "--build",
+            self.build_name,
+            "--branch",
+            "main",
+            "--repo-branch-map",
```
Konboi, line 40: Q: Will we require branch name when the user sets `--no-commit-collection`, right? (Just in case)

Konboi (approved):
I left a question as comment. But this change seems OK, so I'll approve this PR not to block

tests/commands/record/test_build.py
ninjinkun: Yes, that's right.

PR #1061, opened 2025-06-30
Restructure commands with nested command options

ninjinkun:
#  Summary

  This PR implements the NestedCommand pattern for the Smart Tests CLI, introducing a new command structure where test
  runners come before options (e.g., `smart-tests subset pytest --target 25%` instead of `smart-tests subset --target 25% pytest`).

# For reviewer
This Pull Request is mainly for tracking my progress, so I’m not expecting a detailed review. If you can get a general idea of what I’m doing from the title, feel free to go ahead and approve it.

#  Key Changes

  • New Command Structure: Implemented NestedCommand pattern allowing test runner-specific commands with their own
  options
  • Dynamic Command Generation: Added DynamicCommandBuilder class to dynamically create commands for all registered test
  runners
  • Test Runner Registry: Created a centralized registry system to manage test runner functions and enable dynamic
  command creation
  • Backward Compatibility: Maintained legacy command structure as fallback when NestedCommand fails
  • Plugin Support: Enhanced plugin loading system to work with the new command structure
  • CI Integration: Updated GitHub Actions workflow to use the new command syntax

#  Technical Implementation

###  • New Files:
  - smart_tests/utils/dynamic_commands.py - Core dynamic command building logic
  - smart_tests/utils/test_runner_registry.py - Test runner registration system

###  • Modified Files:
  - smart_tests/__main__.py - Added NestedCommand apps and plugin rebuilding logic
  - smart_tests/commands/subset.py - Added NestedCommand support
  - smart_tests/commands/record/tests.py - Added NestedCommand support
  - smart_tests/test_runners/smart_tests.py - Updated decorators for new registry system
  - Various test runner files - Enhanced with proper type annotations

#  Benefits

  • Improved UX: More intuitive command structure where test runner selection comes first
  • Type Safety: Better type annotations and parameter validation
  • Extensibility: Easier to add new test runners and options
  • Plugin Support: Better integration with external test runner plugins

#  Test Plan

  - Verify new command syntax works: smart-tests subset pytest --target 25%
  - Verify legacy command syntax still works as fallback
  - Test plugin loading with new command structure
  - Verify CI workflow uses new command syntax
  - Test all supported test runners with new structure
  - Verify backward compatibility for existing scripts

  🤖 Generated with https://claude.ai/code

PR #1070, opened 2025-07-10
Fix base path problem

ninjinkun:
https://github.com/launchableinc/cli/pull/1061 broke `client.base_path` because of changing the initialization timing. It cause test failures during [the dogfooding](https://github.com/launchableinc/mothership/pull/12016). This PR fix the problem.


smart_tests/commands/test_path_writer.py
```diff
@@ -20,7 +21,9 @@ def __init__(self, app: Application):
     def default_formatter(cls, x: TestPath):
         """default formatter that's in line with to_test_path(str)"""
         file_name = x[0]['name']
-        if cls.base_path:
+        # Only prepend base_path if it was explicitly set via --base option
+        # Auto-inferred base paths should not affect output formatting
+        if cls.base_path and cls.base_path_explicitly_set:
```
Konboi, line 26: I'm not sure what has changed compared to before.

It seems the caller set base_path
```py
    TestPathWriter.base_path = base_path
    TestPathWriter.base_path_explicitly_set = (base_path is not None)
```

Konboi: Also, I cannot find the place the base_path is set automatically.

Could you tell me the place??

ninjinkun: The place is the changes in `dynamic_commands.py`. But I know it's not intuitive. I'm going to revise this PR after my vacation.

PR #1081, opened 2025-07-23
for PTS v2

Konboi:
To Enable PTS v2 Support

- ~~Add logic to check whether PTS v2 is enabled~~
- ~~When PTS v2 is enabled and no input is provided, enable Zero Input Subset (ZIS) by default~~
- ~~If ZIS has not been computed, automatically list files that look like test files~~

- Add `--get-tests-from-guess` option

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -523,10 +523,10 @@ def subset(
                 git_managed_files = subprocess.run(['git', 'ls-files'], stdout=subprocess.PIPE,
                                                    universal_newlines=True, check=True).stdout.strip().split('\n')
             except subprocess.CalledProcessError as e:
-                warn(f"git ls-files failed (exit code={e.returncode})")
+                warn_and_exit_if_fail_fast_mode(f"git ls-files failed (exit code={e.returncode})")
                 return
             except OSError as e:
-                warn(f"git ls-files failed: {e}")
+                warn_and_exit_if_fail_fast_mode(f"git ls-files failed: {e}")
                 return
 
             found = False
@@ -536,7 +536,7 @@ def subset(
                     found = True
 
             if not found:
-                warn("Nothing that looks like a test file in the current git repository.")
+                warn_and_exit_if_fail_fast_mode("Nothing that looks like a test file in the current git repository.")
 
         def request_subset(self) -> SubsetResult:
             test_runner = context.invoked_subcommand
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -627,7 +627,8 @@ def subset(
             """
             if len(subset_result.subset) == 0 and client.is_enabled_pts_v2():
                 self._collect_potential_test_files()
-                subset_result = self.request_subset()
+                if len(self.test_paths) > 0:
+                    subset_result = self.request_subset()
 
             if len(subset_result.subset) == 0:
                 warn_and_exit_if_fail_fast_mode("Error: no tests found matching the path.")
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -534,7 +534,7 @@ def subset(
 
             return payload
 
-        def _collect_test_like_files(self):
+        def _collect_potential_test_files(self):
             LOOSE_TEST_FILE_PATTERN = r'(\.(test|spec)\.|_test\.|Test\.|Spec\.|test/|tests/|__tests__/|src/test/)'
             EXCLUDE_PATTERN = r'\.(xml|json|txt|yml|yaml|md)$'
 
@@ -626,7 +626,7 @@ def subset(
             CLI will try to collect the test files automatically, and request the subset again.
             """
             if len(subset_result.subset) == 0 and client.is_enabled_pts_v2():
-                self._collect_test_like_files()
+                self._collect_potential_test_files()
                 subset_result = self.request_subset()
 
             if len(subset_result.subset) == 0:
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -260,6 +260,12 @@ def subset(
         tracking_client.send_error_event(event_name=event, stack_trace=msg)
         sys.exit(1)
 
+    if is_get_tests_from_guess and is_get_tests_from_previous_sessions:
+        print_error_and_die(
+            "--get-tests-from-guess (list up tests from git ls-files and subset from there) and --get-tests-from-previous-sessions (list up tests from the recent runs and subset from there) are mutually exclusive. Which one do you want to use?",  # noqa E501
+            Tracking.ErrorEvent.USER_ERROR
+        )
+
     if is_observation and is_output_exclusion_rules:
         msg = (
             "WARNING: --observation and --output-exclusion-rules are set. "
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -538,7 +538,8 @@ def subset(
             LOOSE_TEST_FILE_PATTERN = r'(\.(test|spec)\.|_test\.|Test\.|Spec\.|test/|tests/|__tests__/|src/test/)'
             EXCLUDE_PATTERN = r'\.(xml|json|txt|yml|yaml|md)$'
 
-            git_managed_files = subprocess.run(['git', 'ls-files'], capture_output=True, text=True).stdout.strip().split('\n')
+            git_managed_files = subprocess.run(['git', 'ls-files'], stdout=subprocess.PIPE,
+                                               universal_newlines=True).stdout.strip().split('\n')
 
             git_managed_test_files = []
             for f in git_managed_files:
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -201,6 +201,12 @@ from .test_path_writer import TestPathWriter
     type=str,
     metavar='TEST_SUITE',
 )
+@click.option(
+    "--get-tests-from-guess",
+    "is_get_tests_from_guess",
+    help="get subset list from git managed files",
+    is_flag=True,
+)
 @click.pass_context
 def subset(
     context: click.core.Context,
@@ -228,6 +234,7 @@ def subset(
     prioritize_tests_failed_within_hours: Optional[int] = None,
     prioritized_tests_mapping_file: Optional[TextIO] = None,
     test_suite: Optional[str] = None,
+    is_get_tests_from_guess: bool = False,
 ):
     app = context.obj
     tracking_client = TrackingClient(Command.SUBSET, app=app)
@@ -387,6 +394,7 @@ def subset(
             self.exclusion_output_handler = self._default_exclusion_output_handler
             self.is_get_tests_from_previous_sessions = is_get_tests_from_previous_sessions
             self.is_output_exclusion_rules = is_output_exclusion_rules
+            self.is_get_tests_from_guess = is_get_tests_from_guess
             super(Optimize, self).__init__(app=app)
 
         def _default_output_handler(self, output: List[TestPath], rests: List[TestPath]):
@@ -418,7 +426,7 @@ def subset(
 
         def stdin(self) -> Union[TextIO, List]:
             # To avoid the cli continue to wait from stdin
-            if self.is_get_tests_from_previous_sessions:
+            if self.is_get_tests_from_previous_sessions or self.is_get_tests_from_guess:
                 return []
 
             """
@@ -428,12 +436,6 @@ def subset(
             they didn't feed anything from stdin
             """
             if sys.stdin.isatty():
-                """
-                Note: If pts v2 (state if HANDS_ON_LAB_V2), we allow empty stdin
-                """
-                if client.is_pts_v2_enabled():
-                    return []
-
                 warn_and_exit_if_fail_fast_mode(
                     "Warning: this command reads from stdin but it doesn't appear to be connected to anything. "
                     "Did you forget to pipe from another command?"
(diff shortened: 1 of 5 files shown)


Konboi:
@ono-max You can review this PR in the next week. Thanks

launchable/commands/subset.py
```diff
@@ -526,88 +534,111 @@ def get_payload(
 
             return payload
 
+        def _collect_test_like_files(self):
```
ono-max, line 537: `test_like` is slightly ambiguous for me. How about `_collect_candidate_test_files` or `_collect_potential_test_files`?

launchable/utils/launchable_client.py
```diff
@@ -102,15 +102,29 @@ def print_exception_and_recover(self, e: Exception, warning: Optional[str] = Non
...
+
+    def _get_workspace_state(self) -> dict:
+        """
+        Get the current state of the workspace.
+        """
+        if self._workspace_state_cache is not None:
+            return self._workspace_state_cache
         try:
             res = self.request("get", "state")
             if res.status_code == 200:
```
ono-max, line 120: `raise_for_status()` might be better because we don't have to create the additional indent.

```suggestion
            res.raise_for_status()
```

launchable/utils/launchable_client.py
```diff
@@ -102,15 +102,29 @@ def print_exception_and_recover(self, e: Exception, warning: Optional[str] = Non
             click.echo(click.style(warning, fg=warning_color), err=True)
 
     def is_fail_fast_mode(self) -> bool:
-        if 'fail_fast_mode' in self._workspace_state_cache:
-            return self._workspace_state_cache['fail_fast_mode']
-        # TODO: call api and set the result to cache
+        state = self._get_workspace_state()
+        return state.get('fail_fast_mode', False)
```
ono-max, line 106: Is it enough to check only the existence for the value of 'fail_fast_mode'? Is there a possibility that the value of 'fail_fast_mode' could be false?

launchable/utils/launchable_client.py
```diff
@@ -102,15 +102,29 @@ def print_exception_and_recover(self, e: Exception, warning: Optional[str] = Non
...
+            return self._workspace_state_cache
         try:
             res = self.request("get", "state")
             if res.status_code == 200:
-                self._workspace_state_cache['fail_fast_mode'] = res.json().get('isFailFastMode', False)
-                return self._workspace_state_cache['fail_fast_mode']
+                state = res.json()
+                self._workspace_state_cache = {
+                    'state': state.get('state', ""),
+                    'fail_fast_mode': state.get('isFailFastMode', False)
```
ono-max, line 124: Is it enough to check only the existence for the value of 'isFailFastMode'? Is there a possibility that the value of 'isFailFastMode' could be false?

launchable/commands/subset.py
```diff
@@ -669,11 +703,50 @@ def run(self):
...
                 err=True)
 
     context.obj = Optimize(app=context.obj)
 
 
 def subset_request(client: LaunchableClient, timeout: Tuple[int, int], payload: Dict[str, Any]):
     return client.request("post", "subset", timeout=timeout, payload=payload, compress=True)
+
+
+class SubsetResult:
```
ono-max, line 716: Love this idea ❤️

ono-max:
Let me summarize the workflow of PTS v2 to make sure that we're on the same page. Am I correct?

* If users send us the test list,
  * Request the subset with the list.
    * If the subset response is empty, `git ls-files` will be executed, then retry it again.
* Users send us the subset
  * Request the subset with the Zero Input Subsetting.
    * If the subset response is empty, `git ls-files` will be executed, then retry it again.

launchable/utils/launchable_client.py
```diff
@@ -102,15 +102,29 @@ def print_exception_and_recover(self, e: Exception, warning: Optional[str] = Non
             click.echo(click.style(warning, fg=warning_color), err=True)
 
     def is_fail_fast_mode(self) -> bool:
-        if 'fail_fast_mode' in self._workspace_state_cache:
-            return self._workspace_state_cache['fail_fast_mode']
-        # TODO: call api and set the result to cache
```
ono-max, line 107: Would it be possible to use https://docs.python.org/3/library/functools.html#functools.cached_property?

launchable/utils/launchable_client.py
Konboi: Yes, the `isFailFastMode` field is bool type

Konboi:
> Users send us the subset
Request the subset with the Zero Input Subsetting.
If the subset response is empty, git ls-files will be executed, then retry it again.

This one is yes

> If users send us the test list,
Request the subset with the list.
If the subset response is empty, git ls-files will be executed, then retry it again.

But this one is no, so I need to check the self.test_paths before executing `git ls-files`

launchable/commands/subset.py
```diff
@@ -426,6 +428,12 @@ def stdin(self) -> Union[TextIO, List]:
             they didn't feed anything from stdin
             """
             if sys.stdin.isatty():
+                """
+                Note: If pts v2 (state if HANDS_ON_LAB_V2), we allow empty stdin
+                """
```
kohsuke, line 433: There must be some misunderstandings about Python docstring (the triple double quotes). AFAIK, there's no valid use case in which that appears at the top of the if statement, plus, line 420-422 above was added before the original docstring of the `stdin` method.

launchable/commands/subset.py
```diff
@@ -416,7 +418,7 @@ def rel_base_path(path):
 
         def stdin(self) -> Union[TextIO, List]:
             # To avoid the cli continue to wait from stdin
-            if is_get_tests_from_previous_sessions:
+            if self.is_get_tests_from_previous_sessions:
                 return []
```
kohsuke, line 422: That `stdin()` method is returning `Union[TextIO,List]` feels like a leaky abstraction, though granted, it was not introduced in this PR. I guess the backward compatibility ties our hands, but even then, there are ways to evolve these things. I'll see if I can create a follow-up PR.

launchable/commands/subset.py
```diff
@@ -526,88 +534,112 @@ def get_payload(
 
             return payload
 
+        def _collect_potential_test_files(self):
+            LOOSE_TEST_FILE_PATTERN = r'(\.(test|spec)\.|_test\.|Test\.|Spec\.|test/|tests/|__tests__/|src/test/)'
+            EXCLUDE_PATTERN = r'\.(xml|json|txt|yml|yaml|md)$'
+
+            git_managed_files = []
+            try:
+                git_managed_files = subprocess.run(['git', 'ls-files'], stdout=subprocess.PIPE,
```
kohsuke, line 543: `subprocess.CalledProcessError` doc says the `run` method needs to be invoked with `check=True` for the exception to be raised.

launchable/commands/subset.py
```diff
@@ -526,88 +534,112 @@ def get_payload(
...
-            summary = {}
-            subset_id = ""
-            is_brainless = False
-            is_observation = False
+                if msg:
+                    click.echo(click.style(msg, fg="red"), err=True)
+                    exit(1)
 
+            # When Error occurs, return the test name as it is passed.
+            subset_result = None
```
kohsuke, line 616: Unnecessary initialization. In Python, a block statement like `if` does not create a new variable scope.

launchable/utils/launchable_client.py
```diff
@@ -102,15 +102,30 @@ def print_exception_and_recover(self, e: Exception, warning: Optional[str] = Non
...
+        state = self._get_workspace_state()
+        return state.get('fail_fast_mode', False)
+
+    def is_enabled_pts_v2(self) -> bool:
```
kohsuke, line 108: `is_pts_v2_enabled` would be grammatically correct

ono-max (approved):
Although there are some comments from KK, you changed based on my comments. So, I'll approve this PR not to block you.

launchable/commands/subset.py
```diff
@@ -526,88 +534,112 @@ def get_payload(
...
-                    )
+                subset_result = self.request_subset()
 
-                    client.print_exception_and_recover(
-                        e, "Warning: the service failed to subset. Falling back to running all tests")
+            """
+            If the zero input subset response is empty and the workspace is enabled to PTS v2,
+            CLI will try to collect the test files automatically, and request the subset again.
+            """
+            if client.is_enabled_pts_v2() and self.is_get_tests_from_previous_sessions and len(subset_result.subset) == 0:
```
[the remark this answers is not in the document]
kohsuke: I'm now tentatively thinking defining a dedicated option to collect tests from `git ls-files` is better. It avoids surprises, it makes the code cleaner, and it avoids the abstraction leak of "HANDS_ON_LAB_V2" workspace state.

More concretely, my proposal is to define `--guess-tests` (or maybe `--get-tests-from-guess` to signal the fact that it's mutually exclusive with other `--get-tests-...`) and do local test guess scan if and only if this flag is given.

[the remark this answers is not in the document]
kohsuke: I'm now tentatively thinking defining a dedicated option to collect tests from `git ls-files` is better. It avoids surprises, it makes the code cleaner, and it avoids the abstraction leak of "HANDS_ON_LAB_V2" workspace state.

More concretely, my proposal is to define `--guess-tests` (or maybe `--get-tests-from-guess` to signal the fact that it's mutually exclusive with other `--get-tests-...`) and do local test guess scan if and only if this flag is given.

launchable/commands/subset.py
```diff
@@ -618,6 +650,9 @@ def run(self):
 
             # When Launchable returns an error, the cli skips showing summary
             # report
+            original_subset = subset_result.subset
+            original_rest = subset_result.rest
```
kohsuke, line 654: This code is rather unnerving in so many ways:
* The name 'original_xyz' suggests this variable will be mutated later, and yet somehow we still need to maintain the distinction (which suggests leaky abstraction -- why are we needing to mutate this, and why do we need to tell them apart)?
* Yet the list is not copied, which suggests the mutation code will not just manipulate the list, but rather it clones or creates a whole new list.
* And yet no such mutation seems to be happening later.

I'm currently guessing this is just a refactoring remnant. If so, let's remove this. If not, please share with me the motivation behind this, so that I can work on the follow-up PR to do it differently.

launchable/commands/subset.py
```diff
@@ -635,32 +670,32 @@ def run(self):
...
                 "Launchable created subset {} for build {} (test session {}) in workspace {}/{}".format(
-                    subset_id,
+                    subset_result.subset_id,
                     build_name,
                     test_session_id,
                     org, workspace,
                 ), err=True,
             )
-            if is_observation:
+            if subset_result.is_observation:
```
kohsuke, line 698: Why is `is_observation` a response from the server? How is the different from the local flag `this.is_observation`?

kohsuke: Ah, I have the answer myself. The observation flag is an attribute of a test session, not that of a subset request. Yikes.

kohsuke: Ah, I have the answer myself. The observation flag is an attribute of a test session, not that of a subset request. Yikes.

launchable/utils/launchable_client.py
```diff
@@ -102,15 +102,30 @@ def print_exception_and_recover(self, e: Exception, warning: Optional[str] = Non
...
+        state = self._get_workspace_state()
+        return state.get('state', "") == "HANDS_ON_LAB_V2"
```
kohsuke, line 110: I mentioned "the abstraction leak of HANDS_ON_LAB_V2", let me expand it here a bit.

"PTS v2" is amorphous project. So when you define a flag like this, I worry about how the client of this code can easily drift off from what the server side means with this flag. More so if you think about the fact that the CLI is a packaged software and therefore it's more "stiff" than the server side code, meaning we have less control over how long it lives.

And this starts to feel even more out of place if you think about the fact that the test selection logic is really supposed to be a black box from the CLI perspective anyway. Where and why would the CLI need to know how exactly our service is selecting tests? It asks for scrutiny.

AFAICT, the only reason this is needed is in order to trigger implicit local test collection behavior. For mostly unrelated reasons I'm now thinking that's problematic anyway. So I'm inclined to suggest we shouldn't have the `is_enabled_pts_v2` method.

If we do decide to keep, I'd advise against making this another constant to the broken 'workspace state' enum. Enum means those state constants are mutually exclusive, but this one is not -- an ACTIVE_PAYING workspace might have a V2 PTS, and similarly PUBLIC_OSS_PROJECT might have a V2 PTS.

(For that matter, I think we better clarify the meaning of "workspace state". `ARCHIVED` is probably not an enum state either)

kohsuke (approved):
I left so many comments, but but in hindsight, maybe it's me wondering down a tangential path.

In the interest of not blocking the critical effort, I think the only clear "bug" is the one comment around `subprocess.CalledProcessError`, and I can work on a separate PR to try to improve the relevant code a bit.

launchable/commands/subset.py
Konboi: Yes, right

launchable/commands/subset.py
```diff
@@ -526,88 +534,112 @@ def get_payload(
...
+                    click.style(
+                        "Warning: git ls-files failed.",
+                        fg="yellow"),
+                    err=True)
+
+            for f in git_managed_files:
+                if re.search(LOOSE_TEST_FILE_PATTERN, f) and not re.search(EXCLUDE_PATTERN, f):
+                    self.test_paths.append(self.to_test_path(f))
+
+            self.is_get_tests_from_previous_sessions = False
```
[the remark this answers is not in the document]
kohsuke: I'm still puzzled what this achieves.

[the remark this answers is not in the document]
Konboi: I reviewed it again, and this code is unnecessary.
I'll remove it

launchable/commands/subset.py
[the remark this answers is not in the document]
Konboi: We discussed it directly and decided to introduce a new option, `--get-tests-from-guess` without any fallback behaviro

Konboi:
@kohsuke I fixed some points based on your feedback. Could you review again, just in case?
Especially this part https://github.com/launchableinc/cli/pull/1081#discussion_r2237991336

kohsuke (approved):

PR #1098, opened 2025-08-06
send only file name instead of repo + file name

Konboi:
revert this change https://github.com/launchableinc/cli/pull/1092/files#diff-3d1f54ea9449652274261178ede80ec176beefebdb4d1d83e84c3f20bf6ccfa0R27

because server side doesn't support this format


Konboi:
Don't have to review carefully

ono-max (approved):

PR #1103, opened 2025-08-12
Follow-up for #1102

ono-max:


launchable/utils/click.py
```diff
@@ -29,6 +29,7 @@ class PercentageType(ParamType):
 
     def convert(self, value: str, param: Optional[click.core.Parameter], ctx: Optional[click.core.Context]):
         try:
+            missing_percent = False
```
ono-max, line 32: I need to define this variable to fix the following error:

```
ERROR: test_invalid_float (tests.utils.test_click.PercentageTypeTest.test_invalid_float)
  File "/Users/ono-max/src/github.com/launchableinc/cli/tests/utils/test_click.py", line 47, in test_invalid_float
    pct.convert("abc%", None, None)
  File "/Users/ono-max/src/github.com/launchableinc/cli/launchable/utils/click.py", line 43, in convert
    if missing_percent and sys.platform.startswith("win"):
       ^^^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'missing_percent' where it is not associated with a value
```

PR #1126, opened 2025-09-05
[AIENG-230] defined a switch to control the test selection behavior

kohsuke:
The name is still subject to change


Konboi (approved):
LGTM

But you should change the base branch from main to the v1

PR #1146, opened 2025-09-18
remove observation --option from the subset command

Konboi:
SSIA


Konboi:
Don't need to review carefully

PR #1147, opened 2025-09-18
restore --link option to the build command

Konboi:
SSIA


Konboi:
Don't need to review carefully

ono-max (approved):

PR #1163, opened 2025-10-06
Add new command 'inspect model'

gayanW:
This adds new command `launchable inspect model`.
This should only be merged after https://github.com/launchableinc/mothership/pull/12415

**Usage:**

```
% launchable inspect model
| Metadata                        |   Value |
|---------------------------------|---------|
| Training Cutoff Test Session ID |     256 |
```

```
% launchable inspect model --json
{
  "training_cutoff_test_session_id": 256
}
```

### DIFF
--- a/launchable/commands/inspect/model.py
+++ b/launchable/commands/inspect/model.py
@@ -1,0 +1,50 @@
+import json
+import sys
+from http import HTTPStatus
+
+import click
+from requests import Response
+from tabulate import tabulate
+
+from ...utils.launchable_client import LaunchableClient
+
+
+@click.command()
+@click.option(
+    '--json',
+    'is_json_format',
+    help='display JSON format',
+    is_flag=True
+)
+@click.pass_context
+def model(context: click.core.Context, is_json_format: bool):
+    client = LaunchableClient(app=context.obj)
+    try:
+        res: Response = client.request("get", "model-metadata")
+
+        if res.status_code == HTTPStatus.NOT_FOUND:
+            click.echo(click.style(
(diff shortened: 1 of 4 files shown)


launchable/commands/inspect/model.py
```diff
@@ -0,0 +1,50 @@
...
+
+def display_as_json(res: Response):
+    res_json = res.json()
+    click.echo(json.dumps(res_json, indent=2))
+
+
+def display_as_table(res: Response):
+    headers = ["Metadata", "Value"]
+    res_json = res.json()
+    rows = [["Training Cutoff Test Session ID", res_json['training_cutoff_test_session_id']]]
```
ono-max, line 49: Could `training_cutoff_test_session_id` be null? If not, it's okay; otherwise, this throws KeyError, which is not our expectation.

gayanW: It can't be null because the API returns 404 if the value is null in the DB.
https://github.com/launchableinc/mothership/blob/f75c819c8381ef209e8fbc39084a894b6926043b/src/main/java/com/launchableinc/mercury/intake/controller/ModelController.java#L42

gayanW:
> What will happen if the subset is not enabled in the workspace? In other words, what error message will be displayed if the subset is not allows in the workspace?

I just tested. The `/model-metadata` endpoint only queries the `models` table. So it's not affected by the data in the `workspace_feature_flags` table. If DB entry for that workspace not found, or if the the value of the `training_cutoff_test_session_id` column is null, it returns 404. So in that case CLI would output:

"Model metadata currently not available for this workspace."

ono-max (approved):

PR #1274, opened 2026-04-14
Improve performance for large test sets

Konboi:
## Why

When dealing with large test sets (100k+ test paths), two performance bottlenecks exist:
1. The subset request's connect timeout (5s) is too short for large payloads, causing timeouts during transmission
2. The record tests command uploads chunks sequentially, increasing wall-clock time for large test sets

## What

- **Subset connect timeout**: Dynamically increase connect timeout from 5s to 60s when test paths >= 100,000
- **Parallel chunk uploads**: Use `ThreadPoolExecutor` with up to 3 workers to upload test event chunks concurrently in `record tests`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/launchable/commands/record/tests.py
+++ b/launchable/commands/record/tests.py
@@ -3,6 +3,7 @@ import glob
 import os
 import re
 import xml.etree.ElementTree as ET
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from http import HTTPStatus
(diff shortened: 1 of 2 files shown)


Konboi:
@ono-max @gayanW You don't have to review deeply. 
Could you approve this PR?

gayanW (approved):

PR #1279, opened 2026-04-15
fix: replace clear=True with surgical env cleanup in session tests

ItIsUday:
Six session tests were failing due to missing /state endpoint mocks when using clear=True to remove GitHub Actions environment variables. The clear=True approach wiped all environment variables, including SMART_TESTS_BASE_URL, causing the CLI to use production URLs while mocks were configured for localhost:8080.

This commit replaces clear=True with a _clean_github_env() helper
Benefits:
- Tests now inherit SMART_TESTS_BASE_URL from shell environment
- No redundant HTTP mock setup needed in each test
- Single source of truth for GitHub variable cleanup

### DIFF
--- a/tests/commands/record/test_session.py
+++ b/tests/commands/record/test_session.py
@@ -9,20 +9,42 @@ from smart_tests.utils.link import LinkKind
 from tests.cli_test_case import CliTestCase
 
 
+def _clean_github_env():
+    """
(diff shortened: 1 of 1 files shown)


gokul-cloudbees (approved):
LTGM

SculptingSystems (approved):
LGTM

PR #1280, opened 2026-04-16
fix: Redact Authorization bearer token in debug logs

takanabe:
Added _sanitize_headers() function to redact sensitive Authorization bearer tokens when logging HTTP requests in debug mode. The token value is replaced with [REDACTED] while preserving the "Bearer " prefix.

We can use HTTP library for this but this needs library version upgrade. This is the most simplest way to work with small number of code.


Konboi (approved):

takanabe:
this work is already included in https://github.com/cloudbees-oss/smart-tests-cli/releases/tag/v2.7.3 and https://github.com/cloudbees-oss/smart-tests-cli/pull/1284

PR #1284, opened 2026-04-20
fix: Redact Authorization bearer token in debug logs

takanabe:
dded _sanitize_headers() function to redact sensitive Authorization bearer tokens when logging HTTP requests in debug mode. The token value is replaced with [REDACTED] while preserving the "Bearer " prefix.

We can use HTTP library for this but this needs library version upgrade. This is the most simplest way to work with small number of code.

I will merge v1 change to v2 after merging this PR>

### DIFF
--- a/tests/utils/test_http_client.py
+++ b/tests/utils/test_http_client.py
@@ -2,7 +2,7 @@ import os
 import platform
 from unittest import TestCase, mock
 
-from launchable.utils.http_client import _HttpClient
+from launchable.utils.http_client import _HttpClient, _sanitize_headers
 from launchable.version import __version__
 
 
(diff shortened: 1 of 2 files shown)


takanabe:
@Konboi  there is one failure for Python package build.
```
Warning: Your Pipfile requires "python_version" 3.6, but you are using 3.10.20 
from //home/runner/.local/share/v/s/bin/python.
$ pipenv --rm and rebuilding the virtual environment may resolve the issue.
$ pipenv check will surely fail.
tests/utils/test_http_client.py:5: error: Cannot find implementation or library stub for module named "smart_tests.app"  [import-not-found]
tests/utils/test_http_client.py:6: error: Cannot find implementation or library stub for module named "smart_tests.utils.http_client"  [import-not-found]
tests/utils/test_http_client.py:6: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
tests/utils/test_http_client.py:7: error: Cannot find implementation or library stub for module named "smart_tests.version"  [import-not-found]
Found 3 errors in 1 file (checked 159 source files)
Error: Process completed with exit code 1.
```

Is this OK?

tests/utils/test_http_client.py
```diff
@@ -2,9 +2,9 @@
 import platform
 from unittest import TestCase, mock
 
-from launchable.utils.http_client import _HttpClient
-from launchable.version import __version__
-
+from smart_tests.app import Application
```
Konboi: I think we need to fix the package names

Konboi: v1 is launchable, v2 is smart_tets

Konboi (approved):
LGTM - nice

PR #1292, opened 2026-04-22
Fix misleading error for empty adaptive subsets

ono-max:
Adaptive Dynamic Subset can return an empty subset when the model decides no tests in that suite are relevant to the code change.

Until now, the CLI treated any empty `subset` as:

`Error: no tests found matching the path.`

That's confusing for users even though that empty subset is valid. I got this feedback from users, so I'd like to fix this problem.

### DIFF
--- a/launchable/commands/subset.py
+++ b/launchable/commands/subset.py
@@ -626,8 +626,11 @@ def subset(
                 subset_result = self.request_subset()
 
             if len(subset_result.subset) == 0:
-                warn_and_exit_if_fail_fast_mode("Error: no tests found matching the path.")
-                return
+                if len(subset_result.rest) == 0:
+                    warn_and_exit_if_fail_fast_mode("Error: no tests found matching the path.")
+                    return
+                else:
+                    click.echo(click.style("No tests were selected for this code change.", fg="yellow"), err=True)
 
             if split:
                 click.echo("subset/{}".format(subset_result.subset_id))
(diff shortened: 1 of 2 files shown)


launchable/commands/subset.py
```diff
@@ -626,8 +626,11 @@ def run(self):
                 subset_result = self.request_subset()
 
             if len(subset_result.subset) == 0:
-                warn_and_exit_if_fail_fast_mode("Error: no tests found matching the path.")
-                return
+                if len(subset_result.rest) == 0:
+                    warn_and_exit_if_fail_fast_mode("Error: no tests found matching the path.")
+                    return
+                else:
+                    click.echo(click.style("No tests were selected for this code change.", fg="yellow"), err=True)
```
ono-max, line 633: @Konboi 

What do you think about this change?

Konboi: If this kind of use case exists, it can’t really be helped, can it?
That said, I don’t think this would come up in normal scenarios. If we want to be safer, we could limit it to cases where the adaptive dynamic subset is enabled.
Alternatively, we could add a comment noting that this issue may occur when using the adaptive dynamic subset.

Konboi (approved):
I left a comment however not to block your progress, I'll approve this PR

launchable/commands/subset.py
ono-max: Thanks. Updated based on your suggestion: [e499c91](https://github.com/cloudbees-oss/smart-tests-cli/pull/1292/commits/e499c9155ab80feba3e6afa82207e83b10a477e1)

PR #1293, opened 2026-04-28
Track CLI invocation source and full command

ItIsUday:
## Summary
- Sends a `COMMAND_INVOCATION` tracking event on every CLI invocation with the full command string (`rawCommand`), exit code, caller, and CI provider
- Auto-detects the CLI command from argv (maps to `Command` enum, falls back to `UNKNOWN` for typos)
- `SMART_TESTS_CALLER` env var lets wrappers (e.g. GitHub Action) identify themselves
- CI provider auto-detected from standard env vars (`GITHUB_ACTIONS`, `JENKINS_URL`, `CIRCLECI`, `CODEBUILD_BUILD_ID`)
- `caller`, `ciProvider`, `exitCode` sent inside `metadata` JSONB — no DB migration needed
- Existing error/performance tracking unchanged

## Test plan
- [x] Unit tests for `_detect_command()` across all commands + typos + global options
- [x] Unit tests for `send_command_tracking()` payload structure
- [x] Unit tests for `caller`/`ciProvider` in metadata for all event types
- [x] Full test suite passes (376 tests, 0 new failures)


ItIsUday:
There is two pending items though, would like to have an opinion on how to proceed
- There's no input sanitization yet
- The full raw command is captured, so if a command goes something like this, `smart-tests record tests file --session 10 ./test-results/*.xml`, then the full list of `/test-results/` would be expanded in the raw string, which could be huge. Maybe we can just truncate the raw command?

ItIsUday:
Ok I think I'm gonna truncate the raw command to 2000 characters

smart_tests/utils/tracking.py
```diff
@@ -1,21 +1,67 @@
...
+    ("record", "commit"): Command.COMMIT,
+    ("record", "deployment"): Command.RECORD_DEPLOYMENT,
+    ("subset",): Command.SUBSET,
+    ("detect-flakes",): Command.DETECT_FLAKE,
+    ("gate",): Command.GATE,
+    ("update", "alias"): Command.UPDATE_ALIAS,
+}
+
+
+def _detect_command(argv: list[str]) -> Command:
```
psakthivel04, line 38: Do we need to ignore commands that start with --? . why we need to consider that?

smart_tests/__main__.py
```diff
@@ -59,7 +60,18 @@ def _load_test_runners():
...
-    cli.main()
+    argv = sys.argv[:]
+    exit_code = 0
+    try:
+        cli.main()
+    except SystemExit as e:
+        exit_code = e.code if isinstance(e.code, int) else 1
+    finally:
+        try:
+            send_command_tracking(argv=argv, exit_code=exit_code)
```
psakthivel04, line 75: Can we invoke the CLI tracking asynchronously instead of blocking the execution?

smart_tests/utils/tracking.py
```diff
@@ -1,21 +1,67 @@
...
 from smart_tests.utils.authentication import get_org_workspace
+from smart_tests.utils.env_keys import CALLER_KEY, detect_ci_provider
 from smart_tests.utils.http_client import _HttpClient, _join_paths
 from smart_tests.version import __version__
 
 from .commands import Command
 
+# Map CLI subcommand tokens to Command enum values.
+# Longer matches are tried first so "record build" matches before "record".
+_COMMAND_MAP = {
```
psakthivel04, line 18: could you please double check , did we add all the commands. I feel some commands are missing.

eg: 1. record attachment , 2.compare subsets , 3 inspect subset etc

smart_tests/utils/tracking.py
```diff
@@ -1,21 +1,67 @@
...
+    ("verify",): Command.VERIFY,
+    ("record", "build"): Command.RECORD_BUILD,
+    ("record", "session"): Command.RECORD_SESSION,
```
psakthivel04, line 21: Please add a test case that fails whenever a new command is introduced, so that the developer is required to include the new command in this map.

smart_tests/__main__.py
ItIsUday: It is invoked when all the command is processed already, and the CLI is about to exit. Asynchronous execution here might not be needed.

smart_tests/utils/tracking.py
ItIsUday: The flags will be captured in the full raw command

smart_tests/utils/tracking.py
ItIsUday: I had added only the commands that were already in the proto, will check and add all the commands

ItIsUday: Done

smart_tests/utils/tracking.py
ItIsUday: Added the test cases

smart_tests/__main__.py
psakthivel04: While it's true that tracking happens in the `finally` block after the command completes, 
the HTTP request still **blocks the CLI exit**. This means every user waits an extra 
50-200ms (or more with slow networks) for the tracking call to complete

smart_tests/utils/tracking.py
psakthivel04: I think there might be some confusion here. The question is about 
`_detect_command()` logic, not about `rawCommand` capture.

While `rawCommand` does capture everything (including flags), the `_detect_command()` 
function has a separate responsibility: **identify which Command enum this invocation maps to**.

The current implementation has a subtle bug:

```bash
# This would be detected as Command.VERIFY (wrong!)
smart-tests record --name verify

# Because the sliding window finds "verify" at position 3
just think of to skip global flags and only check for commands from that position

Comment (ItIsUday) [reply smart_tests/utils/tracking.py:38]: Yeah, good catch

Comment (ItIsUday) [reply smart_tests/utils/tracking.py:38]: Updated to consider only the tokens before flags appear. The only trade off is that the command used with dry-run won't be identified.
Eg: `smart-tests --dry-run record session --build cli-tracking --test-suite unit-tests > session.txt`

File: smart_tests/__main__.py (hunk 2 above)
Comment (ItIsUday) [reply smart_tests/__main__.py:75]: But since the CLI is about to exit, we have these options
- Forking: But forking will add extra overhead and might not be worth
- Daemon threads or asyncio process get killed the moment `sys.exit()` is called

But these are not helpful unless we change the shutdown process some how.

I will add a short timeout, because if the server is unreachable, the cli will hang, this is more of an issue.

**Review by psakthivel04 — APPROVED** [review]
LGTM

File: smart_tests/utils/tracking.py [hunk 5]
```diff
@@ -62,34 +109,49 @@ def send_error_event(
         api: str = "",
         metadata: Dict[str, Any] | None = None
     ):
-        org, workspace = get_org_workspace()
         if metadata is None:
             metadata = {}
         metadata["stackTrace"] = stack_trace
-        metadata["organization"] = org or ""
-        metadata["workspace"] = workspace or ""
         metadata["api"] = api
```
ono-max, line 115: Why did you leave stackTrace and api in this method instead of putting them in `construct_payload` method?

ItIsUday: That was intentional because the `stackTrace` and `api` are visible to only `send_error_event()`. And that method can create a partial metadata dict with those values already. It is similar to other method like `send_event()` and  `send_command_tracking()` where a partial metadata is generated with values under that specific method's scope.

ono-max (approved):

PR #1297, opened 2026-05-07
Pre-commit CI workflow for V1

ItIsUday:
## Summary

- Add a pre-commit CI workflow that runs on PRs and pushes to main
- Fix pre-commit hook compatibility: switch from deprecated `mirrors-autopep8` (v1.7.0, uses removed `lib2to3`) to official `hhatto/autopep8` (v2.3.2)
- Use Python 3.12 explicitly in the workflow since pre-commit hooks (flake8 7.0, autopep8 2.x) require >=3.8, while `.python-version` targets 3.6 for test compatibility


NicuPascu (approved):
given this is a pre-commit I think there's no risk in getting it in

PR #1305, opened 2026-05-27
Add --subset-id-file option to capture subset ID without parsing stderr

takanabe:
The subset ID is needed for downstream operations (inspect subset, compare subsets, --input-snapshot-id) but was previously only available by parsing stderr. The new --subset-id-file <path> option writes the subset ID to a file as a side-channel, leaving stdout byte-for-byte unchanged so existing test runner pipelines are unaffected.


Konboi (approved):
LGTM

PR #1306, opened 2026-05-28
fixes intermittent test errors occrued due to lagcy token usage

sandeepbarnwal:
### Summary
  
  Replace LAUNCHABLE_TOKEN with SMART_TESTS_TOKEN in` test_codeceptjs.py` to fix test failures in a scenario  when SMART_TESTS_TOKEN is already set in the environment before executing `uv run poe test-xml.`

### Problem
  
  `LAUNCHABLE_TOKEN` is only a fallback in get_token(). If `SMART_TESTS_TOKEN` was already set in the developer's environment, it took precedence over the injected mock token, causing requests to hit unregistered mock URLs and producing empty output that failed json.loads().

image

### Fix

  Use SMART_TESTS_TOKEN (the primary token) so mock.patch.dict always overwrites any real environment value. It becomes consistent with every other test file in the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -145,3 +145,5 @@ bazel-*
 
 .DS_Store
 .idea/*
+test-results
+.local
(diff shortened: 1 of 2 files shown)


Konboi (approved):
Thanks

PR #1308, opened 2026-05-28
Add fallback-mode option for subset

takanabe:
# Background

When the subset API errors or returns isBrainless=true, users can now opt into 'stop' (exit non-zero to halt CI) or 'random-sample' (split locally using --target) instead of the default run-all behavior. The option is hidden and off by default so existing pipelines are unaffected.

# Changes

- Add `--fallback-mode` option for subset command

### DIFF
--- a/smart_tests/commands/subset.py
+++ b/smart_tests/commands/subset.py
@@ -2,6 +2,7 @@ import glob
 import json
 import os
 import pathlib
+import random
 import re
 import subprocess
 import sys
@@ -44,6 +45,12 @@ class SubsetUseCase(str, Enum):
     RECURRING = "recurring"
 
 
+class FallbackMode(str, Enum):
+    RUN_ALL = "run-all"
+    STOP = "stop"
+    RANDOM_SAMPLE = "random-sample"
+
+
 class SubsetResult:
     def __init__(
             self,
@@ -81,6 +88,14 @@ class SubsetResult:
             is_brainless=False,
             is_observation=False
         )
+
+    @classmethod
+    def from_random_sample(cls, test_paths: List[TestPath], target: float) -> 'SubsetResult':
+        count = max(1, round(len(test_paths) * target))
+        sampled = random.sample(test_paths, min(count, len(test_paths)))
+        sampled_set = {id(t): t for t in sampled}
+        rest = [t for t in test_paths if id(t) not in sampled_set]
+        return cls(subset=sampled, rest=rest, subset_id='', summary={}, is_brainless=False, is_observation=False)
 
 
 # Where we take TestPath, we also accept a path name as a string.
@@ -208,6 +223,14 @@ class Subset(TestPathWriter):
                 "--use-case",
                 hidden=True
             )] = None,
+            fallback_mode: Annotated[FallbackMode, typer.Option(
+                "--fallback-mode",
+                hidden=True,
+                help="Behavior when the subset API is unavailable or the model is untrained. "
+                     "'run-all' (default) runs all tests as usual; 'stop' exits with a non-zero status so CI halts; "
+                     "'random-sample' picks a random subset locally based on the count derived from --target "
+                     "(no duration estimates are available in this path).",
+            )] = FallbackMode.RUN_ALL,
             test_runner: Annotated[str | None, typer.Argument()] = None,
     ):
         super().__init__(app)
@@ -288,6 +311,7 @@ class Subset(TestPathWriter):
         self.same_bin_files = list(same_bin_files)
         self.is_get_tests_from_guess = is_get_tests_from_guess
         self.use_case = use_case
+        self.fallback_mode = fallback_mode
 
         self._validate_print_input_snapshot_option()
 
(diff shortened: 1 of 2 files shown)


smart_tests/commands/subset.py
```diff
@@ -562,6 +586,23 @@ def _collect_potential_test_files(self):
...
+        if self.fallback_mode == FallbackMode.STOP:
+            click.echo(
+                "Warning: Smart Tests could not retrieve a subset. Stopping build (--fallback-mode=stop).",
+                err=True,
+            )
+            sys.exit(1)
+        elif self.fallback_mode == FallbackMode.RANDOM_SAMPLE:
+            target_fraction = float(self.target) if self.target is not None else 1.0
+            click.echo(
+                "Warning: Smart Tests could not retrieve a subset. Falling back to local random sample.",
```
Konboi: Since users may use options other than --target e.g.) --confidence, it might be helpful to include the target percentage in the warning message as well.

Or, if --fallback_mode is specified and the user is using an option other than --target, we could return an error in the validation arera

smart_tests/commands/subset.py
```diff
@@ -82,6 +89,14 @@ def from_test_paths(cls, test_paths: List[TestPath]) -> 'SubsetResult':
             is_observation=False
         )
 
+    @classmethod
+    def from_random_sample(cls, test_paths: List[TestPath], target: float) -> 'SubsetResult':
+        count = max(1, round(len(test_paths) * target))
+        sampled = random.sample(test_paths, min(count, len(test_paths)))
+        sampled_set = {id(t): t for t in sampled}
+        rest = [t for t in test_paths if id(t) not in sampled_set]
+        return cls(subset=sampled, rest=rest, subset_id='', summary={}, is_brainless=False, is_observation=False)
```
Konboi, line 98: Is `is_brainless=False` intentionally?

tests/commands/test_api_error.py
```diff
@@ -420,3 +420,104 @@ def assert_tracking_count(self, tracking, count: int):
...
+        responses.replace(
+            responses.POST,
+            f"{get_base_url()}/intake/organizations/{self.organization}/workspaces/{self.workspace}/subset",
+            status=500)
+
+        with tempfile.NamedTemporaryFile(delete=False) as rest_file:
+            result = self.cli(*self._subset_args(rest_file.name, ("--fallback-mode", "random-sample")), mix_stderr=False)
+            self.assert_success(result)
+            all_tests = result.stdout.strip().split("\n") + Path(rest_file.name).read_text().strip().split("\n")
+            all_tests = [t for t in all_tests if t]
```
Konboi: Q: What is this loop?

smart_tests/commands/subset.py
```diff
@@ -697,6 +738,11 @@ def run(self):
         # TODO(Konboi): split subset isn't provided for smart-tests initial release
         # if split:
         #   click.echo("subset/{}".format(subset_result.subset_id))
+        if subset_result.is_brainless:
```
Konboi, line 740: In Brainless mode, we're already splitting requests between subset and rest on the server side.

I can understand the stop behavior, but for sampling, do you mean re-sampling the tests on the CLI side?

smart_tests/commands/subset.py
takanabe: Yes, `is_brainless=False` is intentional. The random sample result is locally generated, not from the brainless model. I hope the behavior matches with this implementation https://github.com/cloudbees-oss/smart-tests-cli/pull/1308/changes/ea8477af18cc53f219f96cc40e6bfa829ba3d279

smart_tests/commands/subset.py
takanabe: sounds prudent. Let me consider

tests/commands/test_api_error.py
takanabe: This is a filter. "".split("\n") returns [""] rather than []. So if either stdout or the rest file is empty, the concatenation produces empty strings and the count assertion would be wrong.

Probably, there is a better way. I'll consider again.

smart_tests/commands/subset.py
takanabe: yeah, you are right. this is redundant if we split tests on server.

takanabe: handled with https://github.com/cloudbees-oss/smart-tests-cli/pull/1308/changes/ea8477af18cc53f219f96cc40e6bfa829ba3d279

tests/commands/test_api_error.py
takanabe: I treated the empty string as the edge case but let me drop the filter for the test for simplicity https://github.com/cloudbees-oss/smart-tests-cli/pull/1308/changes/0451d3ac77fcf6b8ef9258fdadf51f76605036f6

smart_tests/commands/subset.py
takanabe: handled with https://github.com/cloudbees-oss/smart-tests-cli/pull/1308

smart_tests/commands/subset.py
Konboi: OK 👍 
~Could you add a comment that why we update the brainless mode value from true to false here?~

I misunderstood the data flow, you don't have to add comment, sorry

Konboi (approved):
LGTM 

---

I left a comment but it's nits comment so I'll approve this PR

PR #1318, opened 2026-06-04
fix(AIENG-359): Improve text detection by using ByteBuffer and charse…

psakthivel04:
This pull request updates the logic for detecting whether a file is text or binary in the `GitFile` class and adds comprehensive tests to ensure correctness. The main change replaces the previous character-based reading approach with a more robust UTF-8 decoder, and new unit tests are introduced to cover various scenarios.

**Improvements to text/binary file detection:**
* The `isText()` method in `GitFile` now uses a UTF-8 decoder with strict error reporting to determine if file content is valid UTF-8, making detection of binary files more accurate and efficient.

**Testing enhancements:**
* Added a new test class `GitFileTest` with unit tests for `isText()`, covering text files, binary files, and multi-byte UTF-8 content.
* Registered `GitFileTest` in the `AllTests` suite to ensure it is run with other tests.

**Code cleanup:**
* Removed unused imports and added necessary ones in `GitFile.java` to support the new decoding logic.

### DIFF
--- a/src/main/java/com/launchableinc/ingest/commits/GitFile.java
+++ b/src/main/java/com/launchableinc/ingest/commits/GitFile.java
@@ -5,11 +5,10 @@ import org.eclipse.jgit.lib.ObjectLoader;
 import org.eclipse.jgit.lib.ObjectReader;
 
 import java.io.IOException;
-import java.io.InputStreamReader;
 import java.io.OutputStream;
-import java.io.Reader;
+import java.nio.ByteBuffer;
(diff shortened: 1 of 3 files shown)


Konboi:
This change sounds nice however we need to fix this change not only the smart-tests-cli command (v2) but also the launchable command (v1), right?
So, could you create a PR from v1 branch and merge into v1, fist? Next, please merge v1 branch into the main.

Also when you update the Java code, please run `build-java.sh` to update `exe_deploy.jar` file

PR #1319, opened 2026-06-05
[LCHIB-750] Fix robot subset to use file-level suite name as classname for nested suites

Konboi:
## Why

When Robot Framework tests are organized in subdirectories (e.g. `tests/robot/api/*.robot`), the `parse_func` in `robot.py` was assigning the **directory-level suite name** (e.g. `Api`, `Integration`) as the `classname` instead of the **file-level suite name** (e.g. `Auth`, `Comments`).

This caused the server-side `RobotFrameworkResolver` to fail because it tries to match `Api` → `api.robot`, which doesn't exist. The actual files are `auth.robot`, `comments.robot`, etc.

## What

Added a `find_leaf_suites` helper inside `parse_func` that recursively traverses the suite hierarchy until it reaches leaf suites (suites that contain `<test>` elements directly but no child `<suite>` elements). Only leaf suites — which correspond to individual `.robot` files — are passed to `parse_suite`, ensuring the correct file-level suite name is used as `classname`.

Before:
```xml
<testcase classname="Api" name="Register New User Successfully"/>
```

After:
```xml
<testcase classname="Auth" name="Register New User Successfully"/>
```

Also updated test fixtures (`output.xml`, `record_test_result.json`) to cover both flat (2-level) and nested (3-level) suite structures.

## Please Review Here

- `find_leaf_suites` is recursive and handles any depth (2, 3, 4+ levels), verified by the updated `output.xml` which mixes 2-level and 3-level structures in the same run.
- `test_record_test_legacy` was already failing on `v1` before this change (token/org mismatch in mock setup) — not related to this fix.


Konboi:
Closing to rebase on v1 branch.

PR #1332, opened 2026-06-11
Use cloudbees.io URL when CBP workspace

gayanW:
Use cloudbees.io URL instead of app.launchableinc.com when workspace is a CBP workspace.


Konboi:
Q. Don't we need to support it in v1?

gayanW:
@Konboi I'll create a separate PR for v1.

Konboi:
For changes that need to be applied to both v1 and v2, please create the branch from v1 and merge it into v1 first.
Then, please merge the v1 branch into main.

gayanW:
@Konboi Thanks for the info. I'll close this PR.

PR #1351, opened 2026-07-02
[AIENG-545][Backend] OIDC Jenkins provider support + verify --oidc flow

gokul-cloudbees:
###  Summary

Adds tokenless authentication to the CLI using any OIDC identity provider the CI can act as (Jenkins via the oidc-provider plugin, GitLab, custom IdPs). Instead of a static SMART_TESTS_TOKEN, the pipeline presents a short-lived, signed OIDC id-token that Intake verifies cryptographically. This is the CLI half of AIENG-545 Phase 2 (the Intake backend + WebApp changes ship separately).

**Two new actions on the verify command:**

`smart-tests verify --oidc`

Credential-free bootstrap. Presents the id-token to /intake/oidc/verify, prints eval-able export lines for org/workspace/token so later commands authenticate with the same JWT.

`smart-tests verify --oidc-fetch-issuer`

Run from inside a private network. Reads the issuer from the token, fetches its public JWKS via OIDC discovery, and prints an {issuer, jwks} block for an admin to paste into the WebApp. Authenticates nothing; contacts only the issuer.

### DIFF
--- a/smart_tests/args4p/option.py
+++ b/smart_tests/args4p/option.py
@@ -29,7 +29,9 @@ class Option(Parameter):
             required: bool = False,
             metavar: str | None = None,
             multiple: bool = False,
-            hidden: bool = False):
+            hidden: bool = False,
+            optional_value: bool = False,
+            flag_value: Any = None):
         self.name = name        # type: ignore[assignment]  # once properly constructed, name is never None
         self.option_names = option_names
         self.help = help
@@ -39,15 +41,33 @@ class Option(Parameter):
         self.metavar = metavar
         self.multiple = multiple
         self.hidden = hidden
+        # When True, the value is taken ONLY from an inline `--opt=value` form; a bare `--opt`
+        # (no `=`) binds `flag_value` instead of consuming the following argument. This lets a
+        # single option behave both as a flag (`--opt`) and as a valued option (`--opt=value`).
+        self.optional_value = optional_value
+        self.flag_value = flag_value
 
-    def append(self, existing: Any, option_name: str, args):  # args is ArgList, but typing it creates a circular import
+    # 'args' is ArgList, but typing it creates a circular import
+    def append(self, existing: Any, option_name: str, args, has_inline: bool = False):
         '''
         Given the current value 'existing' that represents the present value to invoke the user function with,
         this method is called when this option was specified as 'option_name' on the command line.
         'args' is pointing at the next argument after 'option_name', which may be the value for this option.
+        'has_inline' is True when the option was given as `--opt=value` (the value has been pushed to
+        the front of 'args' by the parser).
         '''
 
-        if self.type == bool or self.type == Optional[bool]:
+        if self.optional_value:
+            if has_inline:
+                v = args.eat(option_name)
+                try:
+                    v = self.type(v)
+                except ValueError as e:
+                    raise BadCmdLineException(f"Invalid value '{v}' for option '{option_name}': {str(e)}") from e
+            else:
+                # bare `--opt`: bind the configured flag value, do NOT consume the next argument.
+                v = self.flag_value
+        elif self.type == bool or self.type == Optional[bool]:
             v = True
         else:
             v = args.eat(option_name)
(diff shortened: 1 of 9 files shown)


smart_tests/args4p/option.py
```diff
@@ -39,15 +41,33 @@ def __init__(
         self.metavar = metavar
         self.multiple = multiple
         self.hidden = hidden
+        # When True, the value is taken ONLY from an inline `--opt=value` form; a bare `--opt`
```
Konboi: I understand the motivation, but unless I’m missing something, it looks like the newly added verify options aren’t actually being used anywhere as a boolean flag. 

Am I misunderstanding something?

smart_tests/utils/authentication.py
```diff
@@ -39,11 +39,27 @@ def ensure_org_workspace() -> Tuple[str, str]:
...
+
 def authentication_headers():
     token = get_token()
     if token:
         return {'Authorization': f'Bearer {token}'}
 
+    # A pipeline that authenticated via `verify --oidc` carries no SMART_TESTS_TOKEN; it presents
+    # its OIDC id-token directly. Intake routes this by `iss` (RESTAuthVerifier) to the generic OIDC
+    # verifier, so subsequent workspace-scoped calls authenticate with the same JWT.
+    oidc_token = get_oidc_token()
```
Konboi, line 59: This is probably a rare case, but if both SMART_TESTS_TOKEN and the OIDC token are set, SMART_TESTS_TOKEN will take precedence. Is that intentional?

I don’t see a strong reason to make it an error, but just in case

gokul-cloudbees: Yeah, `SMART_TESTS_TOKEN` takes precedence as I thought it is primary auth mechanism. I will check @samyakk-rai for the same

smart_tests/args4p/option.py
gokul-cloudbees: refactored this now.

smart_tests/commands/verify.py
```diff
@@ -161,3 +197,260 @@ def verify(app_instance: Application):
...
+        )
+        raise typer.Exit(1)
+
+    # Emit eval-able export lines so the pipeline can hydrate its environment:
+    #   eval "$(smart-tests verify --oidc)"
+    # Subsequent commands then read org/workspace from these vars and present the same OIDC token
+    # (kept in SMART_TESTS_OIDC_TOKEN) as their bearer.
+    click.echo(f'export {ORGANIZATION_KEY}={_shell_quote(org)}')
+    click.echo(f'export {WORKSPACE_KEY}={_shell_quote(workspace)}')
+    click.echo(f'export {OIDC_TOKEN_KEY}={_shell_quote(token)}')
```
ono-max, line 303: Do we really need to export OIDC_TOKEN_KEY? The token must be present to run the command, so this might create unnecessary credential exposure.

smart_tests/commands/verify.py
```diff
@@ -161,3 +197,260 @@ def verify(app_instance: Application):
...
+            stack_trace=msg,
+        )
+        raise typer.Exit(2)
+
+    # The endpoint is not workspace-scoped, so we bypass SmartTestsClient (which requires an
+    # org/workspace) and call the low-level client directly. The OIDC token is supplied as the
+    # bearer by authentication_headers().
+    http_client = _HttpClient(app=app_instance)
+    try:
+        res = http_client.request("post", OIDC_VERIFY_PATH, payload={})
```
ono-max, line 233: Personal opinion: When I read the code, I lost track of where the token was configured. Passing token in the header directly seems more natural to me because we've already gotten the token in the previous step.

```suggestion
        res = http_client.request("post", OIDC_VERIFY_PATH, payload={}, additional_headers={"Authorization": f"Bearer {token}"})
```

smart_tests/commands/verify.py
gokul-cloudbees: this is to support [jenkins cred binding plugin](https://plugins.jenkins.io/credentials-binding/). As it can pass the jwt to env, we have this env introduced. other OIDC supports either local envs or APIs to get the token and Jenkins is quite different.

ono-max (approved):

PR #1352, opened 2026-07-06
Add a new option --fallback-sampling-target so users can specify different ratio with --target

takanabe:
I noticed that --fallback-mode=random-sample can be used without --target option technically. In that case, we don’t know the how many (ratio) we choose tests from all tests. Relevant code is [subset.py](https://github.com/cloudbees-oss/smart-tests-cli/blob/043e61238dd510d6630cdfa323614cd412d36b12/smart_tests/commands/subset.py#L597) 

To overcome the situation, I considered several options below:

Option1: Add a new option --fallback-sampling-target so users can specify different ratio with --target

Option2: Raise error when --target is not specified with random-sample ->  --fallback-mode=random-sample must use with --target

Option3: Return 50% of tests as subset when --target is not specified

We chose the option 1 to specify the target of random sampling percentage when using fallback mode = random-sample

### DIFF
--- a/smart_tests/commands/subset.py
+++ b/smart_tests/commands/subset.py
@@ -623,6 +623,7 @@ class Subset(TestPathWriter):
             )
             sys.exit(1)
         elif self.fallback_mode == FallbackMode.RANDOM_SAMPLE:
+            assert self.fallback_sampling_target is not None
             target_fraction = float(self.fallback_sampling_target)
             click.echo(
                 f"Warning: Smart Tests could not retrieve a subset. "
--- a/smart_tests/commands/subset.py
+++ b/smart_tests/commands/subset.py
@@ -239,6 +239,7 @@ class Subset(TestPathWriter):
             )] = FallbackMode.RUN_ALL,
             fallback_sampling_target: Annotated[Percentage | None, typer.Option(
                 "--fallback-sampling-target",
+                hidden=True,
                 type=parse_percentage,
                 help="Sampling ratio (0%–100%) used when --fallback-mode=random-sample. "
                      "Required when --fallback-mode=random-sample is specified.",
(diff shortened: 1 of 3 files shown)


smart_tests/utils/exceptions.py
```diff
@@ -1,5 +1,6 @@
 # TODO: add cli-specific custom exceptions
 import sys
+from typing import NoReturn
```
Konboi, line 3: I didn't know 'NoReturn`

smart_tests/commands/subset.py
```diff
@@ -234,9 +234,16 @@ def __init__(
                 hidden=True,
                 help="Behavior when the subset API is unavailable or the model is untrained. "
                      "'run-all' (default) runs all tests as usual; 'stop' exits with a non-zero status so CI halts; "
-                     "'random-sample' picks a random subset locally based on the count derived from --target "
+                     "'random-sample' picks a random subset locally based on the count derived from --fallback-sampling-target "
                      "(no duration estimates are available in this path).",
             )] = FallbackMode.RUN_ALL,
+            fallback_sampling_target: Annotated[Percentage | None, typer.Option(
+                "--fallback-sampling-target",
```
Konboi, line 241: The `--random-sample` option is a hidden option however, this `--fallback-sampling-target` option isn't a hidden option.

Is it intentionally?

Konboi (commented):
I left comments however not to block your task, I'll approve this PR

Konboi (approved):

smart_tests/commands/subset.py
takanabe: oh, good catch! I fixed with https://github.com/cloudbees-oss/smart-tests-cli/pull/1352/changes/6f555ef91d4eac6182bb0bf3dbf3caa40cc1c063

Konboi (approved):

PR #1363, opened 2026-08-24
Add --scan-dryrun-results for Maven subset to honor pom.xml test filtering

ItIsUday:
## Why

The Maven `subset` command scans every `*Test.java` under the source roots and sends them all to the backend. But Surefire filters tests at runtime via `pom.xml` — `<excludedGroups>` / `<includedGroups>` (JUnit 5 `@Tag`), `<excludes>` / `<includes>` (class globs), and profiles. The subset never sees that filtering, so it includes tests Maven will never run. At `record` time those tests have no matching report, which produces "missing test report" warnings and skews subset accuracy.

This is the customer-reported case behind [LCHIB-693](https://cloudbees.atlassian.net/wiki/x/ZgDVdgE): `@Tag("IntegrationTest")` / `@Tag("SmokeTest")` tests excluded via `<excludedGroups>IntegrationTest,SmokeTest</excludedGroups>` still get sent to the subset.

## What

New `--scan-dryrun-results` flag on `subset ... maven`. Instead of scanning source files, it parses `**/target/surefire-reports/TEST-*.xml` and sends the fully-qualified class name from each report's root `name` attribute.

The workflow is to run a JUnit 5 dry run first:

```
mvn test -Djunit.platform.execution.dryRun.enabled=true
smart-tests subset ... maven --scan-dryrun-results
```

## Limitations
- JUnit 5 only. The dry-run parameter is a JUnit Platform feature; JUnit 4 `@Category` and TestNG groups are not covered.

[LCHIB-693]: https://cloudbees.atlassian.net/browse/LCHIB-693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Konboi (commented):
My understanding is that this is basically the same as this [PR](https://github.com/cloudbees-oss/smart-tests-cli/pull/1278), but it emphasizes that the scope is limited to JUnit 5. Is that correct?

Also, since the target tenant is using the Launchable commands (V1), please create the branch from `v1` rather than `main`. After merging the changes into `v1`, please merge `v1` into `main`.

PR #1364, opened 2026-08-25
feature/LCHIB-777: improve the gate command output to display actionable failure details

jothikumar-CB:
## Pull request overview

This PR enhances the `gate` command’s human-readable output so that, when a gate fails, it prints actionable per-test failure details (including test identifier path and stderr), with special formatting for GitHub Actions logs.

**Changes:**
- Extend `gate` table output to include “Actionable Failure Details” derived from `actionableFailedTests`.
- Add GitHub Actions-specific log grouping (`::group::` / `::endgroup::`) for actionable failure details when `GITHUB_ACTIONS` is set.
- Update and expand tests to cover the new output and JSON shape.

### DIFF
--- a/launchable/commands/gate.py
+++ b/launchable/commands/gate.py
@@ -13,6 +13,7 @@ from launchable.utils.click import ignorable_error
 from launchable.utils.env_keys import REPORT_ERROR_KEY
 from launchable.utils.tracking import Tracking, TrackingClient
 
+from ..testpath import unparse_test_path
 from ..utils.commands import Command
 from ..utils.launchable_client import LaunchableClient
 
@@ -83,7 +84,7 @@ def gate(ctx: click.core.Context, session: str, is_json_format: bool):
 
 
 def _escape_github_actions_command_value(value: str) -> str:
-    return value.replace('%', '%25').replace('\r', '%0D').replace('\n', '%0A')
+    return value.replace('\r', '%0D').replace('\n', '%0A')
 
 
 def display_as_json(res: Response):
@@ -110,11 +111,7 @@ def display_as_table(res: Response):
     if failed_tests:
         click.echo("\nActionable Failure Details:\n")
         for i, test in enumerate(failed_tests, 1):
-            test_path = "#".join([
-                p["type"] + "=" + p["name"]
-                for p in test.get("testPath", [])
-                if {"type", "name"} <= p.keys()
-            ])
+            test_path = unparse_test_path(test.get("testPath", []))
             stderr = (test.get("stderr") or "").strip()
             if is_github_actions:
                 safe_test_path = _escape_github_actions_command_value(test_path)
--- a/launchable/commands/gate.py
+++ b/launchable/commands/gate.py
@@ -99,3 +99,27 @@ def display_as_table(res: Response):
     ]]
 
     click.echo(tabulate(rows, headers, tablefmt="github"))
+
+    failed_tests = res_json.get('actionableFailedTests', [])
+    is_github_actions = os.getenv('GITHUB_ACTIONS')
+    if failed_tests:
+        click.echo("\nActionable Failure Details:\n")
+        for i, test in enumerate(failed_tests, 1):
+            test_path = "#".join([
+                p["type"] + "=" + p["name"]
+                for p in test.get("testPath", [])
+                if {"type", "name"} <= p.keys()
+            ])
+            stderr = (test.get("stderr") or "").strip()
+            if is_github_actions:
+                click.echo("::group::{}. {}".format(i, test_path))
+                if stderr:
+                    for line in stderr.splitlines():
+                        click.echo(line)
+                click.echo("::endgroup::")
+            else:
+                click.echo("{}. {}".format(i, test_path))
+                if stderr:
+                    for line in stderr.splitlines():
+                        click.echo("   {}".format(line))
+                click.echo("")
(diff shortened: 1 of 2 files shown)


launchable/commands/gate.py
```diff
@@ -99,3 +99,27 @@ def display_as_table(res: Response):
...
+    if failed_tests:
+        click.echo("\nActionable Failure Details:\n")
+        for i, test in enumerate(failed_tests, 1):
+            test_path = "#".join([
+                p["type"] + "=" + p["name"]
+                for p in test.get("testPath", [])
+                if {"type", "name"} <= p.keys()
+            ])
+            stderr = (test.get("stderr") or "").strip()
+            if is_github_actions:
```
mnamachivayam, line 116:
```py
import uuid

def _escape_github_actions_command_value(value):
    return value.replace("%", "%25").replace("\r", "%0D").replace("\n", "%0A")
```

```suggestion
            if is_github_actions:
                safe_test_path = _escape_github_actions_command_value(test_path)
                token = uuid.uuid4().hex

                click.echo("::group::{}. {}".format(i, safe_test_path))

                click.echo("::stop-commands::{}".format(token))
                if stderr:
                    for line in stderr.splitlines():
                        click.echo(line)
                click.echo("::{}::".format(token))

                click.echo("::endgroup::")
```

mnamachivayam: If possible, please add a test that emits the line to print `::warning::spoofed` or `::add-mask::value` which can be interpreted by the runner as a workflow command in the GitHub actions.

mnamachivayam (approved):
Left a comment for you to address. Otherwise, looks good.

launchable/commands/gate.py
```diff
@@ -99,3 +99,27 @@ def display_as_table(res: Response):
     ]]
 
     click.echo(tabulate(rows, headers, tablefmt="github"))
+
+    failed_tests = res_json.get('actionableFailedTests', [])
+    is_github_actions = os.getenv('GITHUB_ACTIONS')
+    if failed_tests:
+        click.echo("\nActionable Failure Details:\n")
```
mnamachivayam, line 112:
```suggestion
        click.echo("\nActionable Failures:\n")
```

launchable/commands/gate.py
jothikumar-CB: Good catch. Thanks

launchable/commands/gate.py
jothikumar-CB: Our table column header is already `Actionable Failures` so `Actionable Failure Details` seems on point.

launchable/commands/gate.py
```diff
@@ -99,3 +104,31 @@ def display_as_table(res: Response):
...
+        for i, test in enumerate(failed_tests, 1):
+            test_path = "#".join([
+                p["type"] + "=" + p["name"]
+                for p in test.get("testPath", [])
+                if {"type", "name"} <= p.keys()
```
ono-max: Why are you checking if {"type", "name"} is subset of `p.keys()`? Can't we simply verify if {"type", "name"} equals to `p.keys()`?

launchable/commands/gate.py
```diff
@@ -99,3 +104,31 @@ def display_as_table(res: Response):
     ]]
 
     click.echo(tabulate(rows, headers, tablefmt="github"))
+
+    failed_tests = res_json.get('actionableFailedTests', [])
+    is_github_actions = os.getenv('GITHUB_ACTIONS')
+    if failed_tests:
+        click.echo("\nActionable Failure Details:\n")
+        for i, test in enumerate(failed_tests, 1):
+            test_path = "#".join([
```
ono-max: Can we use TestPathWriter class to output test paths?

https://github.com/cloudbees-oss/smart-tests-cli/blob/9fac75e9724206802e58b3caef4ccfcb93ef0dd5/smart_tests/commands/test_path_writer.py#L41

launchable/commands/gate.py
jothikumar-CB: Since we get `params` field I checked for subset but now I verified it's ignored for rest but still I would like to go defensive, JIC if we add new field to the components in future.

jothikumar-CB: Now it's totally removed.

launchable/commands/gate.py
```diff
@@ -99,3 +105,27 @@ def display_as_table(res: Response):
...
+
+    failed_tests = res_json.get('actionableFailedTests', [])
+    is_github_actions = os.getenv('GITHUB_ACTIONS')
+    if failed_tests:
+        click.echo("\nActionable Failure Details:\n")
+        for i, test in enumerate(failed_tests, 1):
+            test_path = unparse_test_path(test.get("testPath", []))
+            stderr = (test.get("stderr") or "").strip()
+            if is_github_actions:
+                safe_test_path = _escape_github_actions_command_value(test_path)
```
ono-max, line 117: This escape is necessary not just gate command, but also all commands which outputs test paths, right? How about moving this method to testpath.py and call it within `unparse_test_path` method?

ono-max (approved):

launchable/commands/gate.py
jothikumar-CB: If we add the escape characters inside `unparse_test_path` it will break other use cases. This escape is required only when used with github commands like group.

